PR #4, opened 2019-09-20
Add incrementer for auto-versioner

iamsubratp:
- Add library for different versions
-- for semantic versioning, the increment will based on the type of release. There will be a main binary that will call the appropriate incrementer based on the type of version being followed for the repository.

### DIFF
--- a/versioner/nonsemanticversion.go
+++ b/versioner/nonsemanticversion.go
@@ -15,18 +15,14 @@ var (
 	// ErrInvalidDate is returned if the date being parsed is in a different
 	// format than expected.
 	ErrInvalidDate = fmt.Errorf("Invalid date format")
-	dateFormat     = "01-02-2006"
-)
-
-const (
-	allowedDigits = "0123456789"
+	dateFormat     = "20060102"
 )
 
 // NonSemanticVer is for packages that follow nonsemantic version release
 type NonSemanticVer struct {
 	// in yyyyMMdd format
 	date     string
-	buildNum uint64
+	buildNum int
 }
 
 // NewNonSemanticVer returns a new non semantic version object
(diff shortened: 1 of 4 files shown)


versioner/nonsemanticversion.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var (
+	// ErrInvalidNonSemanticVer is returned if a version is found to be invalid when
+	// being parsed.
+	ErrInvalidNonSemanticVer = fmt.Errorf("Invalid NonSemantic Version")
```
hopkiw: i think we can call the basic version just 'version' instead of 'nonsemantic version'. references to the absence of semantic versioning in this package are confusing.

versioner/nonsemanticversion.go
```diff
@@ -0,0 +1,108 @@
...
+
+// NonSemanticVer is for packages that follow nonsemantic version release
+type NonSemanticVer struct {
+	// in yyyyMMdd format
+	date     string
+	buildNum uint64
+}
+
+// NewNonSemanticVer returns a new non semantic version object
+func NewNonSemanticVer(v string) (*NonSemanticVer, error) {
```
hopkiw: maybe a nit, but this doesn't produce a new, but rather a conversion

versioner/semanticversion.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// ErrInvalidCharacters is returned when invalid characters are found as
+	// part of a version
+	ErrInvalidCharacters = errors.New("Invalid characters in version")
+
+)
+
+// NewSemanticVer parses a given version and returns an instance of version
+// or an error if unable to parse the version
+func NewSemanticVer(v string) (*SemanticVer, error) {
```
hopkiw: we don't need to support semantic versioning any longer

versioner/nonsemanticversion.go
```diff
@@ -0,0 +1,108 @@
+package versioner
```
hopkiw: who imports package versioner? i'm unsure how this fits into the build system

also, i think one-off go packages probably deserve either their own repos, or to be put in compute-image-tools. this repo isn't set up for go development

iamsubratp: yeah, i totally agree that this does not belong here. I wanted this to be in a new repo but I wanted to check with you before moving ahead with requesting a new repo with the security team.
Let's meet tomorrow. I can explain you my end to end idea with the auto-versioner and also i want some clarification on requirements.

PR #6, opened 2019-10-07
Add git tagger

iamsubratp:


autoversioner/tagger/client.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"io/ioutil"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+
+)
+
+const (
+	botUserName = "guesttestinfra-bot"
```
hopkiw: can we pass these in instead of making them constants? that way updates are just to config rather than to code, you can run testing versions, etc.

autoversioner/cmd/tagger/tag.go
```diff
@@ -0,0 +1,72 @@
...
+	flag.Parse()
+	err := validateFlags()
+	if err != nil {
+		fmt.Printf("Error validating flags: %+v\n", err)
+		os.Exit(1)
+	}
+
+	tagger, err := tagger.NewClient(ctx, *tokenFile)
+	if err != nil {
+		fmt.Printf("Error running tagger: %+v\n", err)
```
hopkiw: s/running/creating/

autoversioner/tagger/client.go
iamsubratp: sure

autoversioner/cmd/tagger/tag.go
iamsubratp: my bad! :( will change

PR #7, opened 2019-10-08
Add version generator

iamsubratp:


autoversioner/cmd/tagger/tag.go
```diff
@@ -42,38 +42,38 @@ func main() {
...
 	ref, err := tagger.CreateRef(ctx, *org, *repo, fmt.Sprintf("refs/tags/%s", *tag.Tag), *tag.SHA)
 	if err != nil {
-		return nil, fmt.Errorf("Error while creating ref: %+v", err)
+		return nil, fmt.Errorf("error while creating ref: %+v", err)
 	}
 
 	return ref, nil
 }
 
 func validateFlags() error {
```
hopkiw: doesn't need to be a function, can be a single if with || conditions

autoversioner/cmd/versiongenerator/versiongenerator.go
```diff
@@ -0,0 +1,105 @@
...
+		os.Exit(1)
+	}
+
+	fmt.Printf("Generated version: %s\n", buildID)
+}
+
+// generateVersion generates the version string.
+// The algo to generate it is as follows:
+// - for the given sha; fetch the date on which it was committed
+// - fetch all the commits for that date,
```
hopkiw: should only be valid for today, no need to support historical version building. this can be much simplified by asking for the latest

autoversioner/cmd/versiongenerator/versiongenerator.go
```diff
@@ -0,0 +1,105 @@
...
+// - in the list of commits of chronological order, find the
+//   which commit of the day was this commit.
```
hopkiw: `find the which commit of the day was this commit` seems unclear grammatically

autoversioner/cmd/versiongenerator/versiongenerator.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if len(commits) == 0 {
+		return "", fmt.Errorf("Invalid commit sha")
+	}
+
+	idx, err := getIndex(sha, commits)
+	return fmt.Sprintf("%s.%02d", until.Format(githubDateFormat), len(commits)-idx-1), nil
+}
+
+func getIndex(sha *string, commits []*github.RepositoryCommit) (int, error) {
```
hopkiw: typically you don't need to take pointer values as function arguments unless they'll be modified by the function

autoversioner/cmd/versiongenerator/versiongenerator.go
```diff
@@ -0,0 +1,105 @@
...
+	for idx, commit := range commits {
+		if strings.Compare(commit.GetSHA(), *sha) == 0 {
+			return idx, nil
+		}
+	}
+	return -1, fmt.Errorf("commit not found")
+}
+
+func getTodaysStartTime(t time.Time) *time.Time {
```
hopkiw: doesn't need to be its own function, it's one line of code

autoversioner/cmd/versiongenerator/versiongenerator.go
```diff
@@ -0,0 +1,105 @@
...
+	since := t.Truncate(24 * time.Hour)
+	return &since
+}
+
+func validateFlags() error {
```
hopkiw: same as above

autoversioner/github/client.go
```diff
@@ -1,16 +1,18 @@
...
 	"fmt"
 	"io/ioutil"
+	"time"
 
 	"github.com/google/go-github/github"
 	"golang.org/x/oauth2"
 )
 
 const (
+	// TagObjectTypeCommit is commit type gitobject
```
hopkiw: this comment and the variable name are unclear

iamsubratp:
> strategy is overly complicated bc it supports arbitrary dates - we will only ever be building for todays date and building the newest possible version, so i think it can be simplified.

it was not so much to support arbitary date. It was one of those corner cases for commits being merged just before midnight and there is delayed time interval between when the commit was merged and when the prowjob was kicked in(may be due to network); and this delay caused
dates to change. In this case; the commit date and it version tag will be inconsistent.

autoversioner/cmd/tagger/tag.go
iamsubratp: initially i had them in || conditions but for readability sake i just moved it to a separate function. also introducing this function would help us to do extra validations; if need be

autoversioner/cmd/versiongenerator/versiongenerator.go
iamsubratp: same as the other comment

autoversioner/cmd/versiongenerator/versiongenerator.go
iamsubratp: my bad... will change

autoversioner/cmd/versiongenerator/versiongenerator.go
iamsubratp: agreed... will change

autoversioner/cmd/tagger/tag.go
hopkiw: > > strategy is overly complicated bc it supports arbitrary dates - we will only ever be building for todays date and building the newest possible version, so i think it can be simplified.
> 
> it was not so much to support arbitary date. It was one of those corner cases for commits being merged just before midnight and there is delayed time interval between when the commit was merged and when the prowjob was kicked in(may be due to network); and this delay caused
> dates to change. In this case; the commit date and it version tag will be inconsistent.

we should only generate version once and hand to the build workflow and use same for post-build-tagging. that will either be before midnight or it will be after - we should not care when it was committed, only when we are running the build.

hopkiw: idk why this got attached to this comment, sorry

autoversioner/cmd/versiongenerator/versiongenerator.go
iamsubratp: since the truncate api was not very intuitive; i thought wrapping it in a function would make the purpose more intuitive

autoversioner/github/client.go
iamsubratp: change the variable name; and added a reference to github doc. 
Let me know if it is still unclear

autoversioner/cmd/tagger/tag.go
iamsubratp: i think there is misunderstanding; we are generating the version only once(and actually, with the current algorithm, even if we generate the version for a particular commit multiple times, it will generate the same version); i introduced this so that the commit date and the version date are consistent. 

without this, how would you find the build number for a particular commit if, for any reason, there is a build kicked off on next day(or any other day for that matter)?

hopkiw: you will always get 'today' component of the version, and only ever need to determine what build number for today we're on - you can look at latest git tag for this, regardless of time or date. if the latest git tag isn't today, then the build number is 00.

iamsubratp: iiuc, dont we end up in the same synchronization problem that we talked yesterday. 

2 parallel builds are running; one for a commit before midnight and one for a commit after midnight. we take the today component of the version and for the build number we check the tag; we do not get any tag for today; so we assign 00 to both builds.

also, i want to understand what part of the algorithm do you think is complicated?

autoversioner/version/nonsemanticversion.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+// NonSemanticVer is for packages that follow nonsemantic version release
+type NonSemanticVer struct {
+	// in yyyyMMdd format
+	date     string
+	buildNum int
+}
+
+type VersionSorter []NonSemanticVer
```
hopkiw: we are comparing a slice, so you can use sort.Slice and pass a closure instead of defining these types and methods (if you want)

autoversioner/version/nonsemanticversion.go
```diff
@@ -0,0 +1,131 @@
...
+	// part of a version
+	ErrInvalidCharacters = errors.New("Invalid characters in version")
+)
+
+const (
+	DateFormat = "20060102"
+)
+
+// NonSemanticVer is for packages that follow nonsemantic version release
+type NonSemanticVer struct {
```
hopkiw: can you update references to 'buildid' and 'nonsemanticver' to 'version' or similar?

autoversioner/cmd/versiongenerator/main.go
```diff
@@ -0,0 +1,97 @@
...
+	tags, err := fetcher.ListTags(ctx, *org, *repo)
+	if err != nil {
+		return "", fmt.Errorf("error fetching tags: %+v", err)
+	}
+
+	versions, err := getVersions(tags)
+	// this repository is using our build and release pipeline
+	// for the first time
+	if len(versions) == 0 {
+		return version.FirstVersionToday().String(), nil
```
hopkiw: since incrementver handles the day change, could we omit the duplicate logic in `firstversiontoday` and just set this like:
```
version := NonSemanticVer{}
if len(versions) != 0 {
  version = versions[len(versions)-1]
}
return version.IncrementVersion().String(), nil
```

or something similar? my overall feeling is we have two paths to creating the first version for the day, this one which is only invoked as first time for repo, and then the date-doesnt-match logic in incrementversion()

iamsubratp: got it. yeah, removed the redundancy.

autoversioner/version/nonsemanticversion.go
iamsubratp: changed nonsemanticver to version

autoversioner/version/nonsemanticversion.go
iamsubratp: i read about sort.Slice. It takes a closure which defines in what order the values will be sorted which is same as "Less" method in this implementation.
Also, though might not be very applicable to our case, sort.Slice is slower than sorting the struct by overriding Sort interface

https://stackoverflow.com/questions/54276285/performance-sorting-slice-vs-sorting-type-of-slice-with-sort-implementation

PR #10, opened 2019-10-16
Add build configuration for autoversioner

iamsubratp:


autoversioner/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM scratch
```
hopkiw: can you put dockerfile in container_images/autoversioner directory?

autoversioner/cloudbuild.yaml
```diff
@@ -0,0 +1,13 @@
+steps:
```
hopkiw: this config should go in the shared cloudbuild.yaml

autoversioner/Dockerfile
iamsubratp: i actually started from there. and i spent an aweful amount of time integrating gobuild with cloudbuild and container image.
the [cloud-builders](https://github.com/GoogleCloudPlatform/cloud-builders) specifications are not very clear, even that did not help. the worse thing was i could not even build the example packages that cloud-builders team have in their repo. :(
I reached out the team through [issues](https://buganizer.corp.google.com/issues/142765191); at least i was able to get some pointers around how to proceed further.
That's when i was exposed to cloudbuild parameter called "dir" that lets you run go commands from a directory of your choice. After all experimentation; i wanted a working build for this package :rofl: 

I think, now i have a better understanding of how cloudbuild organizes the code to be built and what features we have at our disposal to achieve a build.

I should be able to move the cloudbuild configuration and container definition under container_images

autoversioner/cloudbuild.yaml
iamsubratp: done

container_images/autoversioner/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM scratch
+
+WORKDIR /workspace
+ADD gopath/bin /
```
hopkiw: can we copy individually, instead, to be more explicit? I also like to copy the dockerfile into the images we produce so you can see how it was produced

iamsubratp: done.

container_images/cloudbuild.yaml
```diff
@@ -11,7 +11,8 @@ steps:
   args: ['build', '--tag=gcr.io/$PROJECT_ID/selinux-tools:latest', '--tag=gcr.io/$PROJECT_ID/selinux-tools:$COMMIT_SHA', './container_images/selinux-tools']
 - name: 'gcr.io/cloud-builders/docker'
   args: ['build', '--tag=gcr.io/$PROJECT_ID/build-essential:latest', '--tag=gcr.io/$PROJECT_ID/build-essential:$COMMIT_SHA', './container_images/build-essential']
-
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/autoversioner:latest', '--tag=gcr.io/$PROJECT_ID/autoversioner:$COMMIT_SHA', '--file=./container_images/autoversioner/Dockerfile', "/workspace"]
```
hopkiw: i think this line needs updating to match others

PR #12, opened 2019-10-18
Add version injection at build time

iamsubratp:


container_images/daisy-builder/main.sh
```diff
@@ -20,6 +20,20 @@ ZONE="$2"
 WORKFLOW_FILE="$3"      # Workflow to run
 GCS_OUTPUT_BUCKET="$4"  # Destination for artifacts
 
+function GENERATE_NEW_VERSION() {
```
hopkiw: nit: function names should be lower case

container_images/daisy-builder/main.sh
```diff
@@ -20,6 +20,20 @@ ZONE="$2"
...
+  VERSION_GENERATE_CMD="/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME}"
```
hopkiw: why store as a variable just to use in next line?

container_images/daisy-builder/main.sh
```diff
@@ -20,6 +20,20 @@ ZONE="$2"
...
+  VERSION_GENERATE_OUT=$(${VERSION_GENERATE_CMD})
+
+  if [[ $? -ne 0 ]]; then
+    echo -e "could not generate version: $VERSION_GENERATE_OUT"
+  fi
+
+  pattern="Generated version: "
+  local BUILD_ID=$(echo $VERSION_GENERATE_OUT | sed "s/$pattern//")
```
hopkiw: can we remove the $pattern from generator command, or add flag to remove it? we control this stuff so we shouldn't have to be doing string parsing on what our utilities provide us

container_images/daisy-builder/main.sh
```diff
@@ -37,8 +51,16 @@ if [[ "$JOB_TYPE" == "presubmit" ]]; then
   DAISY_VARS+=",pull_ref=pull/${PULL_NUMBER}/head"
 fi
 
+## generate buildID
+if [[ "$JOB_TYPE" == "postsubmit" ]]; then
+  BUILD_ID=$(GENERATE_NEW_VERSION)
+  DAISY_VARS+=",build_id=${BUILD_ID}"
```
hopkiw: can we change all references from 'build_id' to 'version' ? build id is something different

container_images/daisy-builder/main.sh
iamsubratp: ack

container_images/daisy-builder/main.sh
iamsubratp: used the variable to debug env variables. i have a echo in my local repo to verify that the env variables are correctly set.

container_images/daisy-builder/main.sh
iamsubratp: done

container_images/daisy-builder/main.sh
iamsubratp: done

container_images/daisy-builder/main.sh
hopkiw: kk, let's just prune from the PR

iamsubratp: done

container_images/daisy-builder/main.sh
```diff
@@ -20,6 +20,17 @@ ZONE="$2"
 WORKFLOW_FILE="$3"      # Workflow to run
 GCS_OUTPUT_BUCKET="$4"  # Destination for artifacts
 
+function generate_new_version() {
+  local VERSION=`/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME}`
+
+  if [[ $? -ne 0 ]]; then
+    echo -e "could not generate version: $VERSION_GENERATE_OUT"
```
hopkiw: i don't think `$VERSION_GENERATE_OUT` is set

hopkiw:
looks good except for one variable needs pruning or updating. thanks

container_images/daisy-builder/main.sh
iamsubratp: done

container_images/daisy-builder/main.sh
```diff
@@ -21,13 +21,13 @@ WORKFLOW_FILE="$3"      # Workflow to run
 GCS_OUTPUT_BUCKET="$4"  # Destination for artifacts
 
 function generate_new_version() {
-  local VERSION=`/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME}`
+  local VERSION_OUT=`/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME}`
 
   if [[ $? -ne 0 ]]; then
```
hopkiw: $? now will refer to the success of the VERSION_OUT= assignment, and will probably always be true.

container_images/daisy-builder/main.sh
```diff
@@ -21,13 +21,13 @@ WORKFLOW_FILE="$3"      # Workflow to run
 GCS_OUTPUT_BUCKET="$4"  # Destination for artifacts
 
 function generate_new_version() {
-  local VERSION=`/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME}`
+  local VERSION_OUT=`/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME}`
```
hopkiw: does versiongenerator print errors to stdout or stderr?

iamsubratp: yes, it does write errors for stdout

container_images/daisy-builder/main.sh
iamsubratp: changed it to catch the error code correctly

hopkiw: Sorry, I wasn't clear. $? will be set to the value of the assignment, which is typically the value of the right operand. But this statement is actually not an assignment, but a `local`. This form of variable initialization always returns 0.

hopkiw: Consider changing to:
```
local VERSION_OUT
if ! VERSION_OUT=`/versiongenerator --token-file-path=${GITHUB_ACCESS_TOKEN} --org=${REPO_OWNER} --repo=${REPO_NAME} 2>&1`; then
    echo "could not generate version: ${VERSION_OUT}"
    return 1
fi
echo $VERSION_OUT
```

iamsubratp: changed

PR #23, opened 2020-01-17
Store git commit, support checkout of tags

hopkiw:
* Inject source control tags into debian control and RPM spec
* Don't assume pull ref is a branch, instead check out object into working tree using FETCH_HEAD ref (enables checkout of tags)
* Skip unversioned specs if versioned specs exist

### DIFF
--- a/packagebuild/common.sh
+++ b/packagebuild/common.sh
@@ -60,11 +60,11 @@ function git_checkout() {
   git init
 
   # fetch only the branch that we want to build
-  git_command="git fetch https://github.com/${BASE_REPO}/${REPO}.git ${PULL_REF:-"master"}:packaging"
+  git_command="git fetch https://github.com/${BASE_REPO}/${REPO}.git ${PULL_REF:-"master"}"
   echo "Running ${git_command}"
   $git_command
 
-  git checkout packaging
+  git checkout FETCH_HEAD
 }
 
 function try_command() {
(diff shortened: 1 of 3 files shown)


hopkiw:
Deb produced after this change:
```
$ apt show google-osconfig-agent
Package: google-osconfig-agent
Version: 1:20191216.00-g1
Status: install ok installed
Priority: optional
Section: misc
Maintainer: Google Cloud Team <gc-team@google.com>
Installed-Size: 14.4 MB
Depends: init-system-helpers (>= 1.18~)
Git: https://github.com/GoogleCloudPlatform/osconfig/tree/1234567
Download-Size: unknown
APT-Manual-Installed: yes
APT-Sources: /var/lib/dpkg/status
Description: Google Compute Engine OSConfig Agent
 Contains the OSConfig agent service binary as well as systemd
 startup scripts
```

hopkiw:
for RPM, although VCS is a tag defined in the spec (unlike Debian), the show equivalent (`rpm -qi`) only shows a predetermined list of fields. So to see the git info we have to run a formatted query:
```
$ rpm -q --queryformat "Name\t: %{NAME}\nVcs\t: %{VCS}\n" google-osconfig-agent
Name	: google-osconfig-agent
Vcs	: https://github.com/GoogleCloudPlatform/osconfig/commit/191599
```

PR #34, opened 2020-08-12
Add pytest image.

EricEdens:
This adds a Docker image definition that allows running Python tests. Tests are executed by pytest, and virtual environment management is provided by tox.

## Testing:
- Built the image using cloud build
- Ran against a repo that's configured for pytest
- Ran prow:
   - New job: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/467
   - Pull: [1326](https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1326)
   - Command: `bash pj-on-kind.sh disk-inspection-pytest`

Results:

```
➜  cat /mnt/disks/prowjob-out/disk-inspection-pytest/1294041507241660416/finished.json
{"timestamp":1597358551,"passed":true,"result":"SUCCESS","revision":"ab2d51613b2b663b8a2e14630dea4d1e86f08c71"}% 

➜  tree /mnt/disks/prowjob-out/disk-inspection-pytest/1294041507241660416
/mnt/disks/prowjob-out/disk-inspection-pytest/1294041507241660416
├── artifacts
│   ├── junit-py35.xml
│   ├── junit-py36.xml
│   ├── junit-py37.xml
│   ├── junit-py38.xml
│   └── tox
│       ├── py35
│       │   ├── py35-0.log
│       │   ├── py35-1.log
│       │   ├── py35-2.log
│       │   └── py35-3.log
│       ├── py36
│       │   ├── py36-0.log
│       │   ├── py36-1.log
│       │   ├── py36-2.log
│       │   └── py36-3.log
│       ├── py37
│       │   ├── py37-0.log
│       │   ├── py37-1.log
│       │   ├── py37-2.log
│       │   └── py37-3.log
│       ├── py38
│       │   ├── py38-0.log
│       │   ├── py38-1.log
│       │   ├── py38-2.log
│       │   └── py38-3.log
│       ├── pyproject.toml
│       └── tox.ini
├── build-log.txt
├── clone-log.txt
├── clone-records.json
├── finished.json
└── started.json
```


container_images/tox/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+[testenv]
+deps =
+    pytest
+
+commands = pytest --junit-prefix={envname} \
+                  --junit-xml={env:ARTIFACTS}/junit-{envname}.xml
+```
+
+To simulate prow's execution, use the following:
```
hopkiw, line 46: reword this to something like 'to run this manually'. prow does a lot of stuff in setup, so this isn't quite analogous

container_images/tox/README.md
```diff
@@ -0,0 +1,63 @@
...
+Your job:
+1. Create a [tox configuration file](https://tox.readthedocs.io/en/latest/config.html) in your
+repository's test directory.
+   * Optionally, write test artifacts to the `ARTIFACTS` directory.
+1. Configure prow with the test directory's path as the first positional argument.
+
+Prior to invoking, the runner will:
+1. Checkout your repository, mount it as a volume, and set its root as Docker's 
+[workdir](https://docs.docker.com/engine/reference/commandline/run/#set-working-directory--w).
+2. Mount a volume for artifacts, and inject its path with the `ARTIFACTS` env variable.
```
hopkiw, line 15: all volume mounting should be in the prow config. how will prow know to mount this additional volume?

container_images/tox/README.md
```diff
@@ -0,0 +1,63 @@
+A prow worker that executes Python tests using [tox](https://tox.readthedocs.io/).
+Check the Dockerfile for which Python versions are installed.
+
+### Interface
+
+Your job:
+1. Create a [tox configuration file](https://tox.readthedocs.io/en/latest/config.html) in your
+repository's test directory.
+   * Optionally, write test artifacts to the `ARTIFACTS` directory.
```
hopkiw, line 9: does this mean a literal directory named ARTIFACTS ? is it a subdir of test/ ?

container_images/tox/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Assembles multiple versions of Python into a single container.
+
+FROM python:3.5.9-slim-buster
```
hopkiw, line 17: why do we need multiple versions of python?

container_images/tox/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```ini
+[tox]
+envlist = py36, py38
+
+[testenv]
+deps =
+    pytest
+
+commands = pytest --junit-prefix={envname} \
```
hopkiw, line 42: usually we would just make a container image which runs this command, rather than one that runs tox and expects a tox config which may have this command. this puts us back in the position of potentially dangerous PRs - we have to know a tox config exists, and that it was not amended during the PR. 

is the main advantage that it will run for multiple python runtimes?

container_images/tox/README.md
```diff
@@ -0,0 +1,63 @@
...
+        └── tests
+            └── tox.ini
+```
+
+This tox configuration will execute [pytest](https://docs.pytest.org/) for both Python 3.6 and Python 3.8,
+and ensure that results are written to the `ARTIFACTS` directory as junit XML files.
+
+```ini
+[tox]
+envlist = py36, py38
```
hopkiw, line 36: not that familiar with tox, but didn't we load the container with py3.5,6,7? where will py38 come from?

container_images/tox/README.md
hopkiw: if you mean $ARTIFACTS i follow, but you should add the dollar sign to make it clear. maybe 'optionally, configure your tests to write their output to the $ARTIFACTS directory, which will exist during test execution'

but below you've given an example where pytest is writing to $ARTIFACTS. what does this bullet point mean for the user? do they need to do anything to their tests?

container_images/tox/README.md
EricEdens: done!

container_images/tox/README.md
EricEdens: Re-worded to: Optionally, write test artifacts to the directory set in the `ARTIFACTS` environment variable.

container_images/tox/Dockerfile
EricEdens: Regarding Tox, the benefits I've found:
- Test against specific, or multiple, versions of Python. I was previously burned when deploying a change that used Python features available on my workstation (Python 3.8), but were not available in the Debian 9 environment where the change was running (Python 3.5).
- Regarding multiple versions, another team is interested in using [1], but they run a different version of Python, so we'd want to test against both.
- The tests are jailed from the system's dependencies; so, if a dependency isn't explicitly modeled, the test will fail. 
- Support for test-only dependencies. In [1], pyyaml is only required for tests, but not at runtime. Setup tools has a deprecated feature `tests_require`, but are recommending users switch to a dedicated test runner such as 'tox'. [2]

1. https://github.com/EricEdens/compute-image-tools/tree/inspect-unit/daisy_workflows/image_import/inspection
2. https://setuptools.readthedocs.io/en/latest/setuptools.html

container_images/tox/README.md
EricEdens: I'm not familiar with the potentially dangerous PRs situation -- what did you have in mind?

container_images/tox/README.md
EricEdens: py38 comes from the final stage "FROM python:3.8.5-slim-buster"

container_images/tox/README.md
```diff
@@ -0,0 +1,63 @@
...
+Your job:
+1. Create a [tox configuration file](https://tox.readthedocs.io/en/latest/config.html) in your
+repository's test directory.
+   * Optionally, write test artifacts to the directory set in the `ARTIFACTS` environment variable.
+1. Configure prow with the test directory's path as the first positional argument.
+
+Prior to invoking, the runner will:
+1. Checkout your repository, mount it as a volume, and set its root as Docker's 
+[workdir](https://docs.docker.com/engine/reference/commandline/run/#set-working-directory--w).
+2. Mount a volume for artifacts, and set the `ARTIFACTS` environment variable to its path.
```
EricEdens, line 15: This is my understanding of how prow sets up the environment, and I included it to help future readers and maintainers understand the execution environment. (It took me awhile to figure out which configurations were required, especially since we have cruft in [1]).

1. https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/test-infra/prowjobs

container_images/tox/README.md
hopkiw: ah, the way we've built the others is that they don't take any command configuration from the repo to be tested, and they're strictly separated. so i was wondering what the advantage we get from this, so we could understand pros/cons

hopkiw: since you've outlined that in the other comment, maybe we ship the tox config into the container and don't read it from the repo?

EricEdens: Updated  to generate the tox.ini file at test-execution time.

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  source tree is mounted read-only.
+
+  Args:
+    pyproject_toml_file: Absolute path to the project config file. The
+      containing directory will be recursively copied to the execution root.
+
+  Returns:
+    Absolute path to execution root.
+  """
+  assert (os.path.isabs(pyproject_toml_file) and
```
hopkiw, line 67: it's not recommended to use 'assert' outside of tests. i think you should use `if` and `raise` instead, with a useful message

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  return exec_root
+
+
+def to_tox_version(py_version: str):
+  """Convert `{major}.{minor}` to `py{major}{minor}`"""
+  if re.fullmatch(r"\d.\d{1,2}", py_version):
+    major, minor = py_version.split(".")
+    return "py{major}{minor}".format(major=major, minor=minor)
+  else:
+    sys.exit("Invalid version number: " + py_version)
```
hopkiw, line 82: here you have a separate paradigm, and also not recommended for functions or methods to invoke sys.exit. maybe use `raise` here also (since it's not caught, it will be equivalent to sys.exit)

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  file.
+
+  The tox.ini file is generated to ensure that tests are run consistently,
+  and that test reports are written to the correct location.
+  """
+  assert os.path.exists("pyproject.toml")
+
+  with open("pyproject.toml") as f:
+    project_cfg = toml.load(f)
+    test_cfg = project_cfg.get("tool", {}).get("gcp-guest-pytest", {})
```
hopkiw, line 96: why is this line indented? does the file handle need to be open for the `.get()` method to work after load?

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  config["testenv"] = {
+    "envlogdir": "{env:ARTIFACTS}/tox/{envname}",
+    "deps": "\n\t".join(_common_test_deps + test_deps),
+    "commands": " ".join(_per_interpreter_command),
+  }
+
+  if os.path.exists("tox.ini"):
+    print("Ignoring existing tox.ini file.")
```
hopkiw, line 120: s/Ignoring/Removing/, imo

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  #  https://github.com/GoogleCloudPlatform/oss-test-infra/blob/cc1f0cf1ffecc0e3b75664b0d264abc6165276d1/prow/oss/config.yaml#L98
+  "--junit-xml={env:ARTIFACTS}/junit-{envname}.xml",
+]
+
+
+def setup_execution_root(pyproject_toml_file: str) -> str:
+  """Create execution directory and copy project's code to it.
+
+  Tox creates in-directory files, and its execution fails when the
+  source tree is mounted read-only.
```
hopkiw, line 58: does prow make the source tree read only? i didn't know that...

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+def validate_args(args: typing.List[str]):
+  if (len(args) > 1 and args[1].endswith("pyproject.toml") and
+      os.path.exists(args[1])):
+    toml_file = os.path.abspath(args[1])
+  else:
+    sys.exit("First argument must be path to pyproject.toml")
+  if "ARTIFACTS" in os.environ and os.path.exists(os.environ["ARTIFACTS"]):
+    artifact_dir = os.path.abspath(os.environ["ARTIFACTS"])
+    # Protect against a relative artifacts directory.
+    os.environ["ARTIFACTS"] = artifact_dir
```
hopkiw, line 144: this is set by prow, i don't think we should modify it for the environment

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  else:
+    sys.exit("$ARTIFACTS must be set to a directory that exists")
+  return artifact_dir, toml_file
+
+
+def main():
+  print("args: %s" % sys.argv)
+  print("$ARTIFACTS: %s" % os.environ.get("ARTIFACTS"))
+  artifact_dir, toml_file = validate_args(sys.argv)
```
hopkiw, line 153: ah, if we are taking the config in as an argument, that means the prow job will have to be configured with the path. that's fine, but just to be clear that's part of the design here and we aren't assuming any location or doing any searching for it

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+
+  execution_root = setup_execution_root(toml_file)
+  os.chdir(execution_root)
+  print("Executing tests in " + execution_root)
+
+  write_tox_ini()
+  save_configs(os.path.join(artifact_dir, "tox"))
+  result_code = tox.cmdline(["-v"])
```
hopkiw, line 161: this invokes tox, right? and our tox config will mean it invokes pytest. will the output be to stdout or stderr?

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+
+  if os.path.exists("tox.ini"):
+    print("Ignoring existing tox.ini file.")
+    os.remove("tox.ini")
+
+  with open("tox.ini", "w") as f:
+    config.write(f)
+
+
+def save_configs(dst: str):
```
hopkiw, line 127: what's the purpose of the tox subdir and copying the configs to it?

container_images/pytest/main.py
```diff
@@ -0,0 +1,166 @@
...
+  print("Saving config files to " + dst)
+  if not os.path.exists(dst):
+    os.mkdir(dst)
+  for fname in ["pyproject.toml", "tox.ini"]:
+    shutil.copy2(fname, dst)
+
+
+def validate_args(args: typing.List[str]):
+  if (len(args) > 1 and args[1].endswith("pyproject.toml") and
```
hopkiw, line 136: you require that argv[0] is a full path to this file, but then you cd to the directory and expect this file to exist. that means that the filename in argv[0] (as opposed to the dirname) is useless data, as it can't be anything other than `pyproject.toml`

maybe take a directory path instead, as you're already validating the file exists in that path. it also matches our other parameterized test images.

container_images/pytest/main.py
EricEdens: Here's the pattern that I was following:
- `assert` is for self-documenting pre-conditions of functions. If it fails, it means there's a programming bug. In terms of what the user sees, Python will raise an AssertionError that shows a stacktrace, eg:

```
/usr/bin/python3.7 /home/ericedens/projects/guest-test-infra/container_images/pytest/main.py
Traceback (most recent call last):
  File "/home/ericedens/projects/guest-test-infra/container_images/pytest/main.py", line 170, in <module>
    main()
  File "/home/ericedens/projects/guest-test-infra/container_images/pytest/main.py", line 152, in main
    assert os.path.isdir('/baz')
AssertionError
```

- `sys.exit("msg")` is for user-facing messages where a stack trace isn't necessary, and would create poor signal-to-noise. In a larger program, I'd probably raise an error, but catch it later and re-format the output prior to exiting the program. So, it was just a pragmatic choice given how small this is.

Given those explanations, let me know what you'd like in this codebase and I'll follow that :)

container_images/pytest/main.py
EricEdens: Resolving since this is covered above

container_images/pytest/main.py
EricEdens: updated to use a directory instead of a path to the config file.

container_images/pytest/main.py
EricEdens: Hygiene mainly, especially since we're deleting the existing `tox.ini` file. This allows the "run-as-local" instructions to use a read-only mount, too.

container_images/pytest/main.py
EricEdens: Tox's status is written to stdout, and I double checked that it's included in the `build-log.txt` file in the artifacts directory.

container_images/pytest/main.py
EricEdens: Ahh, I see, we're using directory paths in the other prow jobs. Switched to a directory rather than the config file.

container_images/pytest/main.py
EricEdens: The original goal was to ensure tox would have an absolute directory when reading `ARTIFACTS`, but as I reflect, we can simply write the artifacts directory to tox's config, and avoid modifying the environment variable. Updated.

container_images/pytest/main.py
EricEdens: I couldn't find documentation either way regarding Prow, but when running locally, I found it useful to mount read-only to ensure `tox` wasn't leaving files on my local repo.

container_images/pytest/main.py
EricEdens: sgtm -- updated

container_images/pytest/main.py
EricEdens: It's not required -- good point. Removed in current version.

container_images/pytest/main.py
hopkiw: you can't unit test methods with sys.exit in them, is the main thing. will leave to your judgement on that, if you think 'exiting the program' is an OK output of a function
i think we should replace any asserts with:
```python
if not <condition>:
  raise Exception("useful message")
```

container_images/pytest/main.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def validate_args(args: typing.List[str]):
+  if (len(args) > 1
+      and os.path.isdir(args[1])
+      and os.path.isfile(os.path.join(args[1], 'pyproject.toml'))):
+    package_root = os.path.abspath(args[1])
+  else:
+    sys.exit("First argument must be path to python package "
+             "package must contain pyproject.toml.")
```
hopkiw, line 148: i think this string will end up "python package package must contain"

container_images/pytest/main.py
```diff
@@ -0,0 +1,173 @@
...
+
+  if os.path.exists("tox.ini"):
+    print("Removing existing tox.ini file.")
+    os.remove("tox.ini")
+
+  with open("tox.ini", "w") as f:
+    config.write(f)
+
+
+def archive_configs(dst: str):
```
hopkiw, line 133: what happens if dst exists, but is a file?

hopkiw:
one other comment - how will these results look to prow? i think prow expects a single junit output file, can we merge the results files to match that?

EricEdens:
> one other comment - how will these results look to prow? i think prow expects a single junit output file, can we merge the results files to match that?

It parses all of the XML files, and then merges the results. 
 - Here's an example: [testgrid](https://k8s-testgrid.appspot.com/google-gce-compute-image-tools#ci-images-import-export-cli-e2e-tests), [artifacts](https://pantheon.corp.google.com/storage/browser/compute-image-tools-test/logs/ci-images-import-export-cli-e2e-tests/1295284952870424576/artifacts?prefix=)

I've added a classname that indicates the interpreter, and I'm planning on using that to differentiate between the test cases via this config: https://github.com/GoogleCloudPlatform/testgrid/blob/721cb652ef79390863b29f4ca36b5a93043fef6e/pb/config/config.proto#L295

container_images/pytest/main.py
EricEdens: It wouldn't behave as expected! Added an assertion that dst is a directory if it exists.

container_images/pytest/main.py
EricEdens: - Switched away from sys.exit in favor of raising an error
- In Python, `assert` is a syntactic shorthand for the snippet you have above (using `AssertionError`). To help with debugging, I added a string message

container_images/pytest/main.py
EricEdens: good catch. Done!

PR #37, opened 2020-09-29
Add integ test runner

hopkiw:


gaohannk (approved):

container_images/gointegtest/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+COPY --from=1 /go /go
+COPY --from=1 /usr/local/go /usr/local/go
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+
+# Go test junit xml output
+RUN go get -u github.com/jstemmer/go-junit-report
+
+RUN apt-get update && apt-get -y install bash ca-certificates curl libssl-dev wget && \
```
EricEdens, line 30: What's libssl-dev for?

container_images/gointegtest/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+
+COPY --from=1 /go /go
+COPY --from=1 /usr/local/go /usr/local/go
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+
+# Go test junit xml output
+RUN go get -u github.com/jstemmer/go-junit-report
+
```
EricEdens, line 29: What do you think of this? It simplifies your Dockerfile, while clarifying what's being copied to the final image. `/usr/bin` may not be the correct destination for the binaries. That aside, I think this formulation clarifies the role of each stage.

```suggestion
FROM golang
RUN go get -u github.com/jstemmer/go-junit-report

FROM gcr.io/compute-image-tools/daisy:latest
FROM gcr.io/google.com/cloudsdktool/cloud-sdk:latest

COPY --from=0 /go/bin/go-junit-report /usr/bin/go-junit-report
COPY --from=1 /daisy /usr/bin/daisy

```

container_images/gointegtest/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get update && apt-get -y install bash ca-certificates curl libssl-dev wget && \
+    rm -rf /var/cache/apt/archives
+
+COPY . .
+
+ENTRYPOINT ["./main.sh"]
```
EricEdens, line 35: Any reason to not use an absolute path? That protects you from invocations that use a different working directory.

container_images/gointegtest/common.sh
```diff
@@ -0,0 +1,80 @@
...
+  REPO="$2"
+  PULL_REF="$3"
+
+  # pull the repository from github - start
+  mkdir -p $REPO
+  cd $REPO
+  git init
+
+  # fetch only the branch that we want to build
+  git_command="git fetch https://github.com/${BASE_REPO}/${REPO}.git ${PULL_REF:-"master"}"
```
EricEdens, line 63: Since you're using github explicitely, you can download a zip, rather than fuss with installing `git`. 

eg: https://github.com/GoogleCloudPlatform/compute-image-tools/archive/dd8aa81d153bffe7f5bae1213cae89ad5abebc6e.zip

container_images/gointegtest/main.sh
```diff
@@ -0,0 +1,43 @@
...
+  -var:repo_owner="$REPO_OWNER" -var:git_ref="$GIT_REF" \
+  -var:gcs_path="$GCS_PATH" integ-test-all.wf.json
+RET=$?
+
+gcloud auth activate-service-account --key-file="$GOOGLE_APPLICATION_CREDENTIALS"
+gsutil cp "$GCS_PATH"/go-test*txt ./
+
+for f in go-test*.txt; do
+  # $ARTIFACTS is provided by prow decoration containers
+  cat "$f" | go-junit-report > "${ARTIFACTS}/junit_${f%%.txt}.xml"
```
EricEdens, line 39: What's the plan for namespacing so that it's clear which platform failed when reading testgrid and spyglass? 

Does this help?  https://github.com/jstemmer/go-junit-report/blob/master/go-junit-report.go#L14

container_images/gointegtest/main.sh
```diff
@@ -0,0 +1,43 @@
...
+  GIT_REF="$PULL_BASE_REF"
+fi
+
+/daisy -project "$PROJECT" -zone "$ZONE" -var:repo_name="$REPO_NAME" \
+  -var:repo_owner="$REPO_OWNER" -var:git_ref="$GIT_REF" \
+  -var:gcs_path="$GCS_PATH" integ-test-all.wf.json
+RET=$?
+
+gcloud auth activate-service-account --key-file="$GOOGLE_APPLICATION_CREDENTIALS"
+gsutil cp "$GCS_PATH"/go-test*txt ./
```
EricEdens, line 35: Consider archiving go-test-$platform.txt. the junit converter is lossy, eg: https://github.com/jstemmer/go-junit-report/issues/108

An easy solution, I think, is to do something like `-var:gcs_path="$ARTIFACTS/daisy" `, since I think this would keep serial console output as well.

container_images/gointegtest/main.sh
hopkiw: no idea, havent planned that part yet

container_images/gointegtest/common.sh
hopkiw: i dont think that supports pull requests. just reusing code here from some of our other containers

container_images/gointegtest/Dockerfile
hopkiw: yeah all of our container images use this entrypoint bc the working dir might differ depending on what layer each are starting from. so this one will always work

container_images/gointegtest/Dockerfile
hopkiw: first, i had no idea you could make suggestions in PR, cool
second, good suggestion, though we still need to copy ca-certs

container_images/gointegtest/Dockerfile
hopkiw: i dont recall tbh, this is copied from package builder

container_images/gointegtest/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM gcr.io/compute-image-tools/daisy:latest
+FROM golang
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:latest
+
+COPY --from=0 /daisy /daisy
+# Daisy image copies this from gcr.io/distroless/base, we copy here.
+COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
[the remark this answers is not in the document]
hopkiw: maybe. i'm combining stuff from different existing images so i'll check

container_images/gointegtest/Dockerfile
hopkiw: removed, seems OK

container_images/gointegtest/Dockerfile
EricEdens: the entrypoint happens at runtime though, right? But the absolute path to your main.sh file is established at build time.

If you wanted to work from any entrypoint, I'd assume that an absolute path would be the safest option.

container_images/gointegtest/main.sh
EricEdens: serves me for giving a concrete solution to a general suggestion :)

EricEdens (approved):
overall lgtm. I'll be curious to see how this looks on testgrid and spyglass, but it sounds like you're working on that next.

PR #50, opened 2020-12-16
add set-pipeline for new pipeline

gaohannk:


hopkiw:
/hold
these aren't even checked in yet, they can't be automated yet

PR #60, opened 2021-01-21
binding roles to team

gaohannk:


concourse/rbac/dev-team.yaml
```diff
@@ -0,0 +1,10 @@
+roles:
```
hopkiw, line 1: we might not want to name this directory rbac. concourse *also* supports k8s rbac, and if we decide to adopt it, it may seem confusing.

hopkiw (commented):
i want to hold on this until we've met and discussed. is it really worthwhile to create more local users now, or should we just prioritize sso? we have no storage mechanism for the passwords for these users either, are we encouraging devs to use browser password keepers?

concourse/rbac/dev-team.yaml
gaohannk: Agree. There are much thing need to discuss before implement.

PR #63, opened 2021-01-28
Validate integration test

gaohannk:
Update: Add platform suffix in XML and Merge Multiple XML into one.


concourse/tasks/validate-integtest-result.yaml
```diff
@@ -0,0 +1,19 @@
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: gcr.io/gcp-guest/validate-integtest
```
hopkiw, line 6: is there a reason to make this a container image, rather than using a generic image and running the script from a git resource?

container_images/validate-integtest/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM golang
+RUN go get -u github.com/jstemmer/go-junit-report
+
+FROM python:3.7.8-slim-buster
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:latest
+
+RUN pip3 install junit2html
+COPY --from=0 /go/bin/go-junit-report /go-junit-report
```
hopkiw, line 21: is it easier to use python as the last 'from' to minimize these copies?

container_images/validate-integtest/validate-result.sh
```diff
@@ -0,0 +1,38 @@
...
+    RET=$?
+  fi
+  cat "$f"
+  cat "$f" | /go-junit-report > "./junit_${f%%.txt}.xml"
+  junit2html "./junit_${f%%.txt}.xml" "./junit_${f%%.txt}.html"
+done
+
+gsutil cp ./junit_*.* "$GCS_PATH"/
+
+echo Done
```
hopkiw, line 37: maybe echo $GCS_PATH instead? or the full path

concourse/tasks/validate-integtest-result.yaml
gaohannk: We need python3 and a package `junit2html`, so i build a new image with dependency installed.

container_images/validate-integtest/validate-result.sh
gaohannk: I will add the  echo $GCS_PATH

container_images/validate-integtest/Dockerfile
gaohannk: We need several env (node, python, golang, sdk) from multiple image. So i think keep the SDK as the last should be fine.

concourse/tasks/validate-integtest-result.yaml
```diff
@@ -0,0 +1,19 @@
...
+- name: credentials
+- name: guest-test-infra
+
+params:
+  GOOGLE_APPLICATION_CREDENTIALS: "credentials/credentials.json"
+  GCS_PATH: gs://gcp-guest-test-outputs/((timestamp))
+  GCS_URL: https://storage.cloud.google.com/gcp-guest-test-outputs/((timestamp))
+run:
+  path: /validate-result.sh
+  args:
```
hopkiw, line 19: if there are no args, can we leave this line off?

container_images/validate-integtest/validate-result.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "Validate Integration Test Result"
+
+gcloud auth activate-service-account --key-file="$GOOGLE_APPLICATION_CREDENTIALS"
+gsutil cp "$GCS_PATH"/go-test*.txt ./
+
+RET=0
+
+# Convert txt report to xml and html
+for f in go-test*.txt; do
+  if [ "`grep -c "FAIL" "$f"`" -gt "0" ]; then
```
hopkiw, line 27: if you just want to check exit code of grep, try like this:
```
if ! grep -qc "FAIL" "$f"; then
  RET=1
```
and set ret rather than using `$?`, because this will be the result of the [ statement anyway

container_images/validate-integtest/validate-result.sh
```diff
@@ -0,0 +1,51 @@
...
+  fi
+  cat "$f"
+  cat "$f" | /go-junit-report > "./junit_${f%%.txt}.xml"
+  # remove prefix junit_go-test and suffix .txt
+  platform=${f%.txt}
+  platform=${platform#junit_go-test-}
+  echo $platform
+  # add platform suffix for test
+  find "junit_${f%%.txt}.xml" -type f -exec sed -i 's/google_guest_agent/google_guest_agent-'${platform}'/g' {} \;
+  find "junit_${f%%.txt}.xml" -type f -exec sed -i 's/google_metadata_script_runner/google_metadata_script_runner-'${platform}'/g' {} \;
```
hopkiw, line 38: can you explain why there's a hardcoded reference to this binary?

container_images/validate-integtest/validate-result.sh
```diff
@@ -0,0 +1,51 @@
...
+    RET=$?
+  fi
+  cat "$f"
+  cat "$f" | /go-junit-report > "./junit_${f%%.txt}.xml"
+  # remove prefix junit_go-test and suffix .txt
+  platform=${f%.txt}
+  platform=${platform#junit_go-test-}
+  echo $platform
+  # add platform suffix for test
+  find "junit_${f%%.txt}.xml" -type f -exec sed -i 's/google_guest_agent/google_guest_agent-'${platform}'/g' {} \;
```
hopkiw, line 37: this code isn't google-guest-agent specific and shouldn't have hardcoded references to any binary

hopkiw (changes requested):

concourse/tasks/validate-integtest-result.yaml
gaohannk: sure.

container_images/validate-integtest/validate-result.sh
gaohannk: Oh, right. I will parameterize this that could use for other package.

container_images/validate-integtest/validate-result.sh
gaohannk: Oh, right. I will parameterize this that could use for other package.

container_images/validate-integtest/validate-result.sh
gaohannk: I think it should be ```if grep -qc "FAIL" "$f"; then```

gaohannk:
Using the `--package-name` argument to add distro in xml test report

PR #73, opened 2021-03-16
first e2e test test hostname

gaohannk:


e2e-test/hostname_test.go
```diff
@@ -0,0 +1,12 @@
+package linux
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
hopkiw, line 6: golang discourages use of assert, that's why you see you have to import it

e2e-test/hostname_test.go
```diff
@@ -0,0 +1,12 @@
+package linux
```
hopkiw, line 1: you included a binary in this PR

PR #74, opened 2021-03-16
new package build workflow

hopkiw:


hopkiw:
currently live as `liamh-testing-package-build-2`

concourse/tasks/generate-package-version.yaml
```diff
@@ -4,19 +4,29 @@ platform: linux
...
 inputs:
-- name: token
+- name: repo
 
 outputs:
 - name: package-version
 
 run:
-  path: sh
+  path: ash
```
gaohannk, line 16: sh?

hopkiw: no ash is the shell provided in alpine

PR #80, opened 2021-03-19
add test wrapper

gaohannk:


test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	err = formatter.JUnitReportXML(report, false, "", out)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func executeAndSaveOutput(file, name string, arg ...string) error {
+	command := exec.Command(name, arg...)
```
hopkiw, line 101: should cd to directory first

hopkiw (changes requested):
mostly nitpicks, looks like it works
need to copy some logic from metadata script runner though, and create per-run working directories. we should also give a directory to the downloaded object for execution, a scratch space for the downloaded binary, and an 'artifacts' dir for the executed test. then upload whatever we find there to daisy-outs/artifacts/

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,171 @@
...
+
+	out, err := os.Create(output)
+	if err != nil {
+		return fmt.Errorf("Failed to create file %v", err)
+	}
+	report, err := parser.Parse(in, "")
+	if err != nil {
+		return err
+	}
+	err = formatter.JUnitReportXML(report, false, "", out)
```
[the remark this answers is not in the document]
hopkiw: apply this style everywhere (editor should usually make this easy for you)

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		log.Fatalf("Failed to download object: %v\n", err)
+	}
+
+	err = executeAndSaveOutput(testTxtResult, testBinaryLocalPath, testArgument...)
+	if err != nil {
+		log.Fatalf("Failed to execute test binary: %v\n", err)
+	}
+
+	err = covertTxtToJunit(testTxtResult, testJunitResult)
```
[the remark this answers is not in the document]
hopkiw: take a string and return a converted string, i mean

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		log.Fatalf("failed to create cloud storage client: %v", err)
+	}
+
+	testBinaryURL, err := getMetadataAttribute("_test_binary_url")
+	if err != nil {
+		log.Fatalf("failed to get metadata _test_binary_url: %v", err)
+	}
+
+	u, err := url.Parse(testBinaryURL)
```
hopkiw, line 42: no need to parse here, parse in downloadGCSObject

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		log.Fatalf("failed to parse gcs url: %v", err)
+	}
+	bucket, object := u.Host, u.Path
+
+	testRun, err := getMetadataAttribute("_test_run")
+	if err != nil {
+		log.Fatalf("failed to get metadata _test_binary_url: %v", err)
```
hopkiw, line 50: you are going to fatal every test that does not set this value. it should be OK for _test_run to be empty, right?

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	var testArguments = []string{"-test.v"}
+	if testRun != "" {
+		testArguments = append(testArguments, "-test.run", testRun)
+	}
+
+	workDir := "/test-" + randString(5)
```
hopkiw, line 58: workDir prefix maybe a constant? or at least a global

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err = junitFormatter.JUnitReportXML(report, false, "", &b); err != nil {
+		return nil, err
+	}
+	return &b, nil
+}
+
+func executeAndSaveOutput(stdoutFile, cmd string, arg []string) error {
```
hopkiw, line 103: why save output?

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if err = downloadGCSObject(ctx, client, bucket, object, testBinaryLocalPath); err != nil {
+		log.Fatalf("failed to download object: %v", err)
+	}
+
+	if err = executeAndSaveOutput(testTxtResult, testBinaryLocalPath, testArguments); err != nil {
+		log.Fatalf("failed to execute test binary: %v", err)
+	}
+
+	testData, err := convertTxtToJunit(testTxtResult)
```
hopkiw, line 76: this reads a file you JUST wrote. how about passing this function a string or []byte instead

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+		return "", err
+	}
+	req.Header.Add("Metadata-Flavor", "Google")
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	if resp.StatusCode != 200 {
+		return "", fmt.Errorf("http response code is %v", resp.StatusCode)
```
hopkiw, line 130: nitpick but we know statuscode is an int, you can use %d

test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	err = ioutil.WriteFile(stdoutFile, output, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to write stdout: %v", err)
+	}
+	return nil
+}
+
+func getMetadataAttribute(attribute string) (string, error) {
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s%s", metadataURLPrefix, attribute), nil)
```
hopkiw, line 119: sprintf with '%s%s' is the same as 'string1'+'string2', isn't it?

hopkiw (changes requested):

test_wrapper/test_wrapper.go
gaohannk: fix

test_wrapper/test_wrapper.go
gaohannk: Yes, It should be ok, we should allow no value input instead just empty string

test_wrapper/test_wrapper.go
gaohannk: Move to downloadGCSObject

test_manager/test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		log.Fatalf("failed to create cloud storage client: %v", err)
+	}
+
+	testBinaryURL, err := getMetadataAttribute("_test_binary_url")
+	if err != nil {
+		log.Fatalf("failed to get metadata _test_binary_url: %v", err)
+	}
+
+	testRun, err := getMetadataAttribute("_test_run")
```
hopkiw, line 40: if you don't care about err, it is idiomatic not to capture it: 
```
testRun, _ := 
```

test_manager/test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,156 @@
...
+
+	var testArguments = []string{"-test.v"}
+	if testRun != "" {
+		testArguments = append(testArguments, "-test.run", testRun)
+	}
+
+	if err = os.Mkdir(workDir+artifactPath, 0755); err != nil {
```
hopkiw, line 47: don't you need to try creating workDir before creating workDir+artifactPath ?

hopkiw: also lets have artifactPath be a full absolute path, so we don't have to remember how to build it later

test_manager/test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,156 @@
...
+		log.Fatalf("failed to create artifact dir: %v", err)
+	}
+	if err = os.Mkdir(workDir, 0755); err != nil {
+		log.Fatalf("failed to create work dir: %v", err)
+	}
+	if err = os.Chdir(workDir); err != nil {
```
hopkiw, line 53: nitpick but if main thread changes dir mid-work, it is a bit fragile. usually we will download-TO, and then execute-IN, where functions do the chdir if needed. probably not even needed then,though, exec should take a directory argument. these models are just best practice, make it easier to add e.g. goroutine support later, etc. and in case we ever want to write more files, we know we are not in same dir as the child process, nobody can step on any toes

test_manager/test_wrapper/test_wrapper.go
```diff
@@ -0,0 +1,156 @@
...
+	return nil
+}
+
+func uploadGCSObject(ctx context.Context, client *storage.Client, testBinaryURL string, data io.Reader) error {
+	u, err := url.Parse(testBinaryURL)
+	if err != nil {
+		log.Fatalf("failed to parse gcs url: %v", err)
+	}
+	bucket := u.Host
+	des := client.Bucket(bucket).Object(testResultObject).NewWriter(ctx)
```
hopkiw, line 150: np dst

test_manager/test_wrapper/test_wrapper.go
hopkiw: fyi the `exec.Cmd` struct has a `Dir` attribute that you can set for working dir

hopkiw (changes requested):
overall looks good, needs the 'artifacts' feature but we can add on later

test_manager/test_wrapper/test_wrapper.go
gaohannk: Fix

test_manager/test_wrapper/test_wrapper.go
gaohannk: Good catch.

test_manager/test_wrapper/test_wrapper.go
gaohannk: Add a Dir parameter in executeCommand() to set workDir

test_manager/test_wrapper/test_wrapper.go
gaohannk: Fix

test_manager/test_wrapper/test_wrapper.go
hopkiw: unresolving, what we want is a dir param in executeCmd and a dst param in downloadGCSObject, then remove chdir from main()

PR #86, opened 2021-03-26
guest-package-build bugfixes

hopkiw:
* rename inject steps for oslogin build
* correct ash-supported shell arithmetic in generate-package-version
* correct reference to gcloud/alpine image


gaohannk:
lgtm

PR #97, opened 2021-04-10
add test manager

hopkiw:


hopkiw:
lol prow is trying to run these tests.. might have to disable that haha

PR #98, opened 2021-04-12
imagetest package

hopkiw:
filing as smaller PRs, this first PR has only the imagetest package main workflow code


imagetest/junit_test.go
```diff
@@ -0,0 +1,141 @@
...
+		{
+			[]string{testPass, testFail},
+			&testSuite{Tests: 9, Failures: 1},
+		},
+	}
+	for idx, tt := range tests {
+		ts := convertToTestSuite(tt.results)
+		switch {
+		case ts.XMLName != tt.ts.XMLName:
+			fallthrough
```
zoran15, line 63: Shouldn't this fail a test?

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+	baseGCSPath       = "gs://gcp-guest-test-outputs/cloud_image_tests/"
+	createVMsStepName = "create-vms"
+)
+
+// TestWorkflow defines a test workflow which creates at least one test VM.
+type TestWorkflow struct {
+	Image          string
+	Name           string
+	ShortImage     string
+	destination    string
```
zoran15, line 34: What is destination?

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+	testWrapperPath   = testBinariesPath + "/wrapper"
+	baseGCSPath       = "gs://gcp-guest-test-outputs/cloud_image_tests/"
+	createVMsStepName = "create-vms"
+)
+
+// TestWorkflow defines a test workflow which creates at least one test VM.
+type TestWorkflow struct {
+	Image          string
+	Name           string
+	ShortImage     string
```
zoran15, line 33: What is the difference between Image and ShortImage?

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+	destination    string
+	skipped        bool
+	skippedMessage string
+	wf             *daisy.Workflow
+}
+
+func finalizeWorkflows(tests []*TestWorkflow, zone, project string) error {
```
zoran15, line 40: Since the logic is fully within the for loop below, finalizeWorkflow (singular) could naturally be a method on TestWorkflow struct. You can keep finalizeWorkflows if that makes code shorter elsewhere.

Also, add docs what this does in summary since it's not clear without going through the whole method. Just looking at the name, it sounded like this is wrapping up tests (once they finished), but it's actually building/extending test Daisy workflows. It's more of a builder function.

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+
+type testResult struct {
+	testWorkflow    *TestWorkflow
+	skipped         bool
+	workflowSuccess bool
+	err             error
+	results         []string
+}
+
+func getTestResults(ctx context.Context, ts *TestWorkflow) ([]string, error) {
```
zoran15, line 92: this can be a method on *TestWorkflow

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// PrintTests prints all test workflows.
+func PrintTests(ctx context.Context, testWorkflows []*TestWorkflow, project, zone string) {
+	finalizeWorkflows(testWorkflows, zone, project)
+	for _, test := range testWorkflows {
+		if test.wf == nil {
+			continue
+		}
+		test.wf.Print(ctx)
```
zoran15, line 129: minor: can be a func on *TestWorkflow

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+// ValidateTests validates all test workflows.
+func ValidateTests(ctx context.Context, testWorkflows []*TestWorkflow, project, zone string) error {
+	finalizeWorkflows(testWorkflows, zone, project)
+	for _, test := range testWorkflows {
+		if test.wf == nil {
+			continue
+		}
+		if err := test.wf.Validate(ctx); err != nil {
+			return err
+		}
```
zoran15, line 142: minor: can be a func on *TestWorkflow

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	defer outFile.Close()
+
+	outFile.Write(bytes)
+	outFile.Write([]byte{'\n'})
+
+	return
+}
+
+func runTestWorkflow(ctx context.Context, test *TestWorkflow) testResult {
```
zoran15, line 198: can be a method on *TestWorkflow

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+
+	}
+	return nil
+}
+
+type testResult struct {
+	testWorkflow    *TestWorkflow
+	skipped         bool
+	workflowSuccess bool
+	err             error
```
zoran15, line 88: can `err != nil` be used instead of `workflowSuccess == True`?

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+	case res.workflowSuccess:
+		// Workflow completed without error. Only in this case do we try to parse the result.
+		ret = convertToTestSuite(res.results)
+	default:
+		ret.Tests = 1
+		ret.Errors = 1
+		if res.err != nil {
+			ret.SystemErr = res.err.Error()
+		} else {
+			ret.SystemErr = "Unknown status"
```
zoran15, line 239: Can this happen? Looking at `runTestWorkflow` it seems either `err` is being set or `workflowSuccess` is True

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		log.Fatalf("Failed to set up storage client: %v", err)
+	}
+	finalizeWorkflows(testWorkflows, zone, project)
+
+	testResults := make(chan testResult, len(testWorkflows))
+	testchan := make(chan *TestWorkflow, len(testWorkflows))
+
+	var wg sync.WaitGroup
+	for i := 0; i < parallelCount; i++ {
```
zoran15, line 160: if `parallelCount` is greater than `len(testWorkflows)` I think this will dead lock waiting at `range testchan`. Add a check or trim `parallelCount` to ensure this doesn't happen.

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+		wg.Add(1)
+		go func(id int) {
+			defer wg.Done()
+			for test := range testchan {
+				testResults <- runTestWorkflow(ctx, test)
+			}
```
zoran15, line 166: How does this parallelization work? Below, all testWorkflows are dumped into `testchan`. But here, a single gofunc can apparently consume all of them.

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+		}(i)
+	}
+	for _, ts := range testWorkflows {
+		testchan <- ts
+	}
+	close(testchan)
+	wg.Wait()
```
zoran15, line 173: Maybe close testchan after wg.Wait()? Otherwise, channel could be closed while gofunctions are still reading from it.

imagetest/testworkflow.go
```diff
@@ -0,0 +1,245 @@
...
+
+	var suites testSuites
+	for i := 0; i < len(testWorkflows); i++ {
+		suites.TestSuite = append(suites.TestSuite, parseResult(<-testResults))
```
zoran15, line 177: Curious, is `<-testResult` even reached, since wg.Wait() will wait until test runners are done. Test runners output result to `testResults` but there's no one to read from them. Maybe I'm misunderstanding the logic.

imagetest/junit.go
```diff
@@ -0,0 +1,98 @@
...
+	junitFormatter "github.com/jstemmer/go-junit-report/formatter"
+	junitParser "github.com/jstemmer/go-junit-report/parser"
+)
+
+type testSuites struct {
+	XMLName   xml.Name     `xml:"testsuites"`
+	Name      string       `xml:"name,attr"`
+	Errors    int          `xml:"errors,attr"`
+	Failures  int          `xml:"failures,attr"`
+	Tests     int          `xml:"tests,attr"`
```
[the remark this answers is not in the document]
hopkiw: it would be fine to generate these values one-time, but they need to exist in the struct bc the struct represents the final junit output. the junit spec is pre-existing and we are conforming to it

imagetest/junit.go
```diff
@@ -0,0 +1,98 @@
...
+	Time      float64      `xml:"time,attr"`
```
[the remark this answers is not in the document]
hopkiw: junit spec defines this, it is t ime the test took to run

imagetest/junit.go
```diff
@@ -0,0 +1,98 @@
+package imagetest
```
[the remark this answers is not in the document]
hopkiw: i have seen package-level comments before, but not file-level. is there a file with an example of this you're thinking of?

imagetest/junit.go
```diff
@@ -0,0 +1,98 @@
...
+type junitFailure struct {
+	FailMessage string `xml:",chardata"`
+	FailType    string `xml:"type,attr"`
+}
+
+// converts `go test` outputs to a jUnit testSuite
+func convertToTestSuite(results []string) *testSuite {
+	ts := &testSuite{}
+	for _, testResult := range results {
+		tcs, err := convertToTestCase(testResult)
```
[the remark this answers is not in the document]
hopkiw: each string here represents the total, multi-line output of running `go test`. so []results will have the results of multiple `go test` invocations.

imagetest/junit.go
```diff
@@ -0,0 +1,98 @@
...
+			if tc.Failure != nil {
+				ts.Failures++
+			}
+		}
+	}
+	return ts
+}
+
+// converts a single `go test` output to jUnit TestCases
+func convertToTestCase(in string) ([]*testCase, error) {
```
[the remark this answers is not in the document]
hopkiw: this converts a single test output from `go test` to []*testCase

imagetest/junit_test.go
hopkiw: i am handling multiple possible failures with fallthrough to the final case, which does the actual handling. it was an alternative to the comma syntax, but if it's not clear i could switch to the comma syntax instead

imagetest/testworkflow.go
hopkiw: where results from test are uploaded, including daisy outputs

imagetest/testworkflow.go
hopkiw: image: projects/projectname/global/images/family/debian-10
shortimage: debian-10

i will add a comment to clarify this

imagetest/testworkflow.go
hopkiw: yes it 'finalizes' a workflow to make it ready to run. will look at updating as you suggest

imagetest/testworkflow.go
hopkiw: is it beneficial to make anything that can be a method into a method? i tend to make methods when there will be some shared data in the object that is going to be modified, or when i want the method to be public to the user of the package.

imagetest/testworkflow.go
hopkiw: it is ok to read from a closed channel. if there are remaining items, they are returned. otherwise, reading from a closed channel returns the 0 type for the channel type. finally, reading a channel with 'range' will stop when the channel is closed.

imagetest/testworkflow.go
hopkiw: we create N goroutines, give them a test channel and a results channel. each goroutine is using `range` to read from testchan, placing results in results channel, and exit when it is empty
then we load the testchan with all workflows to run, and immediately close the channel
we wait for all goroutines to finish
then we pull all of the results out of the result channel serially and parse them

ime this is a fairly standard form of goroutine based parallelization, sometimes called worker goroutines

imagetest/testworkflow.go
hopkiw: let's say there are 5 workflows to run but we set parallel count to 30
we will create all 30 goroutines, which will wait on test workflows to run
we will load the 5 workflows to the test channel and close it
the first 5 goroutines will each grab one
the remaining 25 goroutines will then see the closed channel and exit
the first 5 will exit once done with their work

imagetest/testworkflow.go
hopkiw: i left a longer explanation of the parallelization in another comment, but if this one goroutine is fast enough to get all the workflows and run them before another goroutine can grab any, that's great. overachiever goroutine :)

imagetest/testworkflow.go
zoran15: The criteria whether something is a method that I use is whether something logically belongs to a class/struct or it operates mostly on class fields. For example, getTestResults retrieves results from a TestWorkflow. It mostly references TestWorkflow fields, so is a good candidate for a method.

imagetest/junit.go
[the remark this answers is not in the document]
zoran15: There's no specific format. Since you prefer to keep multiple classes in a single file, having a top level comment describing what they do together would be helpful

imagetest/testworkflow.go
zoran15: Consider renaming to `outputPath` or similar

zoran15 (approved):

PR #100, opened 2021-04-12
add test fixtures

hopkiw:
depends on #98 
add test fixtures

### DIFF
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -16,7 +16,6 @@ package imagetest
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/GoogleCloudPlatform/compute-image-tools/daisy"
 )
@@ -26,56 +25,6 @@ const (
 	createDisksStepName = "create-disks"
 	successMatch        = "FINISHED-TEST"
 )
-
-// SingleVMTest configures one VM running tests.
-func SingleVMTest(t *TestWorkflow) error {
-	_, err := t.CreateTestVM("vm")
-	return err
-}
-
-// Skip marks a test workflow to be skipped.
-func (t *TestWorkflow) Skip(message string) {
-	t.skipped = true
-	t.skippedMessage = message
-}
-
-// SkippedMessage returns the skip reason message for the workflow.
-func (t *TestWorkflow) SkippedMessage() string {
-	return t.skippedMessage
-}
-
-// CreateTestVM creates the necessary steps to create a VM with the specified name to the workflow.
-func (t *TestWorkflow) CreateTestVM(name string) (*TestVM, error) {
-	// TODO: more robust name validation.
-	name = strings.ReplaceAll(name, "_", "-")
-
-	createDisksStep, err := t.appendCreateDisksStep(name)
-	if err != nil {
-		return nil, err
-	}
-
-	// createDisksStep doesn't depend on any other steps.
-
-	createVMStep, err := t.appendCreateVMStep(name)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := t.wf.AddDependency(createVMStep, createDisksStep); err != nil {
-		return nil, err
-	}
-
-	waitStep, err := t.addWaitStep(name, name, false)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := t.wf.AddDependency(waitStep, createVMStep); err != nil {
-		return nil, err
-	}
-
-	return &TestVM{name: name, testWorkflow: t}, nil
-}
 
 // TestVM is a test VM.
 type TestVM struct {
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -1,0 +1,173 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package imagetest
+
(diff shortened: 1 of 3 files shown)


imagetest/testworkflow.go
```diff
@@ -48,12 +48,119 @@ type TestWorkflow struct {
...
+	instance := &daisy.Instance{}
+	instance.StartupScript = "wrapper"
+	instance.Name = name
+	instance.Scopes = append(instance.Scopes, "https://www.googleapis.com/auth/devstorage.read_write")
+	instance.Disks = append(instance.Disks, attachedDisk)
+
+	instance.Metadata = make(map[string]string)
+	instance.Metadata["_test_vmname"] = name
+	instance.Metadata["_test_package_url"] = "${SOURCESPATH}/testpackage"
+	instance.Metadata["_test_results_url"] = fmt.Sprintf("${OUTSPATH}/%s.txt", name)
```
ericdand, line 69: Please use a composite literal declaration to initialize instance all at once. (see go/go-style/best-practices#vardeclcomposite)

imagetest/testworkflow.go
```diff
@@ -48,12 +48,119 @@ type TestWorkflow struct {
...
+		createVMStep.CreateInstances = createInstances
+	}
+
+	return createVMStep, nil
+}
+
+func (t *TestWorkflow) appendCreateDisksStep(diskname string) (*daisy.Step, error) {
+	bootdisk := &daisy.Disk{}
+	bootdisk.Name = diskname
+	bootdisk.SourceImage = t.Image
```
ericdand, line 93: Use composite literal here too

imagetest/testworkflow.go
```diff
@@ -48,12 +48,119 @@ type TestWorkflow struct {
...
+}
+
+func (t *TestWorkflow) addWaitStep(stepname, vmname string, stopped bool) (*daisy.Step, error) {
+	serialOutput := &daisy.SerialOutput{}
+	serialOutput.Port = 1
+	serialOutput.SuccessMatch = successMatch
+
+	instanceSignal := &daisy.InstanceSignal{}
+	instanceSignal.Name = vmname
+	instanceSignal.Stopped = stopped
```
ericdand, line 120: Use composite literal here too

imagetest/testworkflow_test.go
```diff
@@ -16,8 +16,222 @@ package imagetest
...
+		t.Error("StartInstances step is missing")
+	}
+	if len(step.StartInstances.Instances) != 1 {
+		t.Error("StartInstances step is malformed")
+	}
+	if step.StartInstances.Instances[0] != "vmname" {
+		t.Error("StartInstances step is malformed")
+	}
+	stepFromWF, ok := twf.wf.Steps["start-stepname"]
+	if !ok || step != stepFromWF {
```
ericdand, line 45: You can join these into one line:

`if stepFromWF, ok := twf.wf.Steps["start-stepname"]; !ok || step != stepFromWF {`

ericdand: (Same for the final check of the other newly-added tests too)

imagetest/testworkflow_test.go
```diff
@@ -16,8 +16,222 @@ package imagetest
...
+	}
+	if twf.wf == nil {
+		t.Error("test workflow is malformed")
+	}
+	step, err := twf.appendCreateDisksStep("diskname")
+	if err != nil {
+		t.Errorf("failed to add wait step to test workflow: %v", err)
+	}
+	if step.CreateDisks == nil {
+		t.Error("CreateDisks step is missing")
```
ericdand, line 158: Consider using `t.Fatal` here instead, as your test will otherwise panic on line 160. Not the end of the world if it panics (it's still considered a failure), but it makes the test logs harder to read.

ericdand (commented):
Looking pretty good! Mostly just style things.

imagetest/fixtures.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return &TestVM{name: name, testWorkflow: t}, nil
+}
+
+// TestVM is a test VM.
+type TestVM struct {
+	name         string
+	testWorkflow *TestWorkflow
+	instance     *daisy.Instance
```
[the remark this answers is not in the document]
hopkiw: there is a constructor - testworkflow.CreateTestVM
we will not allow users to specify these private fields, they come from the test workflow.

imagetest/testworkflow.go
hopkiw: not for daisy, daisy has a bunch of weird types that make us do this. it's also a heavily nested structure,  so this style becomes more readable in this instance

imagetest/testworkflow.go
```diff
@@ -48,12 +48,119 @@ type TestWorkflow struct {
...
+
+	startInstancesStep, err := t.wf.NewStep("start-" + stepname)
+	if err != nil {
+		return nil, err
+	}
+	startInstancesStep.StartInstances = startInstances
+
+	return startInstancesStep, nil
+}
+
```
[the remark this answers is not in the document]
hopkiw: no, all the private methods are intended to be private. very little about it is expected to be exposed to the user.

imagetest/testworkflow_test.go
hopkiw: ah yes, this originally did more with stepFromWF

imagetest/testworkflow.go
ericdand:
```suggestion
	instance := &daisy.Instance{
		StartupScript: "wrapper",
		Name:          name,
		Scopes:       []string{"https://www.googleapis.com/auth/devstorage.read_write"},
		Disks:        []disk{attachedDisk},
		Metadata:      map[string]string{
			"_test_vmname": name,
			"_test_package_url": "${SOURCESPATH}/testpackage",
			"_test_results_url": fmt.Sprintf("${OUTSPATH}/%s.txt", name),
		}
	}
```

ericdand:
Liam has expressed a preference for the way types are initialized, despite the style guide, and also pointed out that the imperative style at least does match much of the rest of this project. Since this is GitHub rather than an internal review, we have a little more leeway to ignore the style guide.

/lgtm

PR #107, opened 2021-04-16
add image license test

gaohannk:
image_validation git:(image-test-license) /out/manager -project hanga-testing -zone us-west1-c -images projects/debian-cloud/global/images/debian-10-buster-v20210316
2021/04/23 23:06:11 imagetest: Done with setup
runTestWorkflow: running image_validation on debian-10-buster-v20210316 (ID m0npn)
runTestWorkflow: running image_boot on debian-10-buster-v20210316 (ID rvn01)
```
<testsuites name="" errors="0" failures="0" tests="0" time="0">
	<testsuite name="image_validation-debian-10-buster-v20210316" tests="5" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestHostname" time="0"></testcase>
		<testcase classname="" name="TestFQDN" time="0.01"></testcase>
		<testcase classname="" name="TestHostKeysGeneratedOnce" time="0.03"></testcase>
		<testcase classname="" name="TestArePackagesLegal" time="0.14"></testcase>
		<testcase classname="" name="TestCustomHostname" time="0.01"></testcase>
	</testsuite>
	<testsuite name="image_boot-debian-10-buster-v20210316" tests="2" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestGuestBoot" time="0"></testcase>
		<testcase classname="" name="TestGuestReboot" time="0"></testcase>
	</testsuite>
</testsuites>
```


imagetest/cmd/manager/main.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
-	imagevalidation "github.com/GoogleCloudPlatform/guest-test-infra/imagetest/test_suites/image_validation"
```
hopkiw, line 11: this is not necessary, i was leaving image_validation this way to show how the package name does not have to match the directory name

imagetest/test_suites/imagevalidation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+
+func TestArePackagesLegalToUse_License(t *testing.T) {
+	problemPackages := validatePackageLegal("/usr/share/doc/*/LICENSE")
+	if problemPackages != nil {
+		t.Fatalf("The following packages are not legal to use %s", problemPackages)
+	}
+}
+
+func TestArePackagesLegalToUse_CopyRight(t *testing.T) {
```
hopkiw, line 159: no underscore in function names

imagetest/test_suites/imagevalidation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+	problemPackages := validatePackageLegal("/usr/share/doc/*/copyright")
+	if problemPackages != nil {
+		t.Fatalf("The following packages are not legal to use %s", problemPackages)
+	}
+}
+
+func validatePackageLegal(path string) []string {
```
hopkiw, line 166: this doesn't validate (which would return bool), but returns non-conforming packages. rename to suit what it does

imagetest/test_suites/imagevalidation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"TEXT-SOUNDEX",
+	"TEXT-TABS",
+	"The OpenLDAP Public License",
+	"Unicode",
+	"X11",
+	"X11-2",
+	"ZLIB",
+}
+
+var Licenses = []string{
```
hopkiw, line 65: why is this public?

imagetest/test_suites/imagevalidation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+	return false
+}
+
+func isValidLicense(licenseCheck string) bool {
+	return isValidLicenseName(licenseCheck) || isValidLicenseText(licenseCheck)
+}
+
+
+func TestArePackagesLegalToUse_License(t *testing.T) {
+	problemPackages := validatePackageLegal("/usr/share/doc/*/LICENSE")
```
hopkiw, line 153: this line and the next can be combined as `if problemPackages := ...; problemPackages != nil {`

hopkiw (changes requested):

imagetest/cmd/manager/main.go
gaohannk: Revert the change.

imagetest/test_suites/imagevalidation/license_test.go
gaohannk: return bool instead []string, print []string in this function.

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+}
+
+func TestArePackagesCopyRightLegal(t *testing.T) {
+	if !isPackageLegal("/usr/share/doc/*/copyright") {
+		t.Fatalf("The packages are not legal to use")
+	}
+}
+
+func isPackageLegal(path string) bool {
```
hopkiw, line 163: this is named as if it's a single package but it is expecting a wildcard path
maybe move the for loop outside of this function and really take a single path as it suggests

hopkiw (changes requested):

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+			licenseCheck += scanner.Text()
+		}
+
+		if err := scanner.Err(); err != nil {
+			log.Printf("error read file")
+			continue
+		}
+
+		re := regexp.MustCompile(`(\\*|#)*`)
+		licenseCheck = re.ReplaceAllString(licenseCheck, " ")
```
hopkiw, line 186: licenseCheck is not used?

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+		for scanner.Scan() {
+			licenseCheck += scanner.Text()
+		}
+
+		if err := scanner.Err(); err != nil {
+			log.Printf("error read file")
+			continue
+		}
+
+		re := regexp.MustCompile(`(\\*|#)*`)
```
hopkiw, line 185: i don't think you need escapes in backtick-strings. and here you seem to be saying 0-or-more \ characters or a single \#, it doesn't seem to make sense as a regular expression. i encourage you not to try to create a 1-to-1 translation of the code you're basing this on but to understand what it does and implement it in go.

hopkiw (changes requested):

imagetest/test_suites/image_validation/license_test.go
hopkiw: the python regex you're copying from here is `r'(\*|#)*'`, that is a single asterisk or single \#

hopkiw:
```
Raw string literals are character sequences between back quotes, as in `foo`. Within the quotes,
any character may appear except back quote. The value of a raw string literal is the string
composed of the uninterpreted (implicitly UTF-8-encoded) characters between the quotes; in
particular, backslashes have no special meaning and the string may contain newlines. 
```
https://golang.org/ref/spec

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,200 @@
...
+	if !isPackageLegal("/usr/share/doc/*/copyright") {
+		t.Fatalf("The packages are not legal to use")
+	}
+}
+
+func TestArePackagesCopyRightLegal(t *testing.T) {
+	if !isPackageLegal("/usr/share/doc/*/copyright") {
+		t.Fatalf("The packages are not legal to use")
+	}
+}
```
hopkiw, line 161: these two functions are now identical

gaohannk: copy paste error. `/usr/share/doc/*/License`

imagetest/test_suites/image_validation/license_test.go
gaohannk: Refactor as suggested.

imagetest/test_suites/image_validation/license_test.go
gaohannk: Good catch. I aware that it is trying to remove `*` or `#` with zero or more times. The double \`\ escapes `\` and change the semantics.

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,197 @@
...
+	`Some portions of os-prober`,
+	`Netcat and the associated package is a product of Avian Research, and is freely available in full source form with no restrictions save an obligation to give credit where due.`,
+	`Permission is hereby granted, without written agreement and without licence or royalty fees, to use, copy, modify, and distribute this software`,
+	`Open Market permits you to use, copy, modify, distribute, and license this Software and the Documentation for any purpose, provided that existing copyright notices are retained in all copies and that this notice is included verbatim in any distributions. No written agreement, license, or royalty fee is required for any of the authorized uses.`,
+	`This software is made available under the terms of *either* of the licenses found in LICENSE.APACHE or LICENSE.BSD. Contributions to cryptography are made under the terms of *both* these licenses.`,
+}
+
+func isValidLicenseName(licenseCheck string) bool {
+	for _, name := range licenseNames {
+		var regexString = fmt.Sprintf("(?i)"+"(?:(?:License|Copyright)\\s*:\\s*{0})|(?:(?:covered )*under (?:the )?{0})|(?:under (?:the terms of )*the {%s})", name)
```
hopkiw, line 126: the `{0}` entries are python formatting. replace with `%[1]s` in each instance

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,192 @@
...
+func isPackageLegal(filepath string) bool {
+	bytes, err := ioutil.ReadFile(filepath)
+	if err != nil {
+		log.Printf("error read file")
+	}
+
+	var licenseCheck string = string(bytes)
+
+	// Remove comment
+	re := regexp.MustCompile(`(\*|#)*`)
```
hopkiw, line 180: you probably want + (1-or-more) rather than * (0-or-more) operator
and this change can almost certainly be done with strings.ReplaceAll or regexp.Regexp.ReplaceAllString

imagetest/utils/test_utils.go
```diff
@@ -88,3 +99,23 @@ func DownloadGCSObjectToFile(ctx context.Context, client *storage.Client, gcsPat
 	}
 	return nil
 }
+
+// IsTargetLinux check if the vm is target linux distribution.
+func IsTargetLinux(version string) bool {
+	file, err := os.Open(osVersionFile)
```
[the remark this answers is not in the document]
gaohannk: Good Catch. I want to answer both questions.

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,192 @@
...
+	loc := re.FindStringIndex(licenseCheck)
+	if loc != nil {
+		licenseCheck = licenseCheck[0:loc[0]] + licenseCheck[loc[1]:]
+	}
+	// Replace all whitespace with one space
+	licenseCheck = strings.Join(strings.Fields(licenseCheck), " ")
+	if !isValidLicenseName(licenseCheck) && isValidLicenseText(licenseCheck) {
+		log.Printf("The package %s are not legal to use.", filepath)
```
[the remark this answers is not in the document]
hopkiw: oh and logging should be done with the testing object, not log. don't import log package into tests

imagetest/utils/test_utils.go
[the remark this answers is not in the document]
gaohannk: The second one may need in other tests.

imagetest/test_suites/image_validation/license_test.go
gaohannk: Agree 1-or-more is more reasonable.

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,192 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func isValidLicenseText(licenseCheck string) bool {
+	for _, licenseText := range licenses {
+		// (?i) case insensitive
+		re := regexp.MustCompile(`(?i)(` + licenseText + `)`)
```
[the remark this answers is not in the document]
gaohannk: sorry, i didn't get it. I use `()` to make sure it `?i` to the all licenseText as wholw capture group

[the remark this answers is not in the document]
hopkiw: in other languages the syntax has been `(?i`+licenseText+`)`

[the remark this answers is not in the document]
hopkiw: ok i looked here and you've got it almost right. you just have an unnecessary capture group containing the entire expression. update to 
```
MustCompile('(?i)'+licenseText+')')
```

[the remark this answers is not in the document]
gaohannk: I guess you mean ```MustCompile('(?i)'+licenseText')```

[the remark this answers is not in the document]
hopkiw: yes that's right :)

imagetest/test_suites/image_validation/license_test.go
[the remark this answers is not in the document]
gaohannk: I remove this as we already log in upper function using testing object.

imagetest/test_suites/image_validation/setup.go
```diff
@@ -19,5 +19,10 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 	}
 	vm2.RunTests("TestCustomHostname")
 
+	vm3, err := t.CreateTestVM("vm3.license")
```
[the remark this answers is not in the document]
gaohannk: Got it. move to vm1

imagetest/utils/test_utils.go
```diff
@@ -14,7 +14,16 @@ import (
...
-const metadataURLPrefix = "http://metadata.google.internal/computeMetadata/v1/instance/"
+const (
+	metadataURLPrefix = "http://metadata.google.internal/computeMetadata/v1/instance/"
+)
+
+// OS Version Name
+const (
+	RedHat = "rhel"
+	Debian = "debian"
+	SUSE   = "suse"
```
hopkiw, line 25: i don't understand the model you're going for, here. i think it's up to test author to figure out on what condition to test, right? and there are a lot more and a lot more sub-types than this.

imagetest/utils/test_utils.go
```diff
@@ -88,3 +97,15 @@ func DownloadGCSObjectToFile(ctx context.Context, client *storage.Client, gcsPat
 	}
 	return nil
 }
+
+// IsTargetLinux check if the vm is target linux distribution.
+func IsTargetLinux(version string) bool {
+	imageMetadata, err := GetMetadata("image")
+	if err != nil {
+		log.Fatalf("Failed to get image from metadata server: %v\n", err)
```
hopkiw, line 105: don't use log package in tests, and you should be returning an error so caller can determine how to handle.
and since they're handling that anyway, i think they can just call getmetadata directly, what do you think?

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,190 @@
...
+		if !isPackageLegal(filename) {
+			t.Fatalf("Found illegal package: %v", filename)
+		}
+	}
+}
+
+func isPackageLegal(filepath string) bool {
+	bytes, err := ioutil.ReadFile(filepath)
+	if err != nil {
+		log.Printf("error read file")
```
hopkiw, line 174: don't use log package in tests, use t.Info or similar

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+const (
+	copyrightPathGlob = "/usr/share/doc/*/copyrightPathGlob"
+	licensePathGlob   = "/usr/share/doc/*/LICENSE"
+	licenseNameRegex  = `(?i)((?:(?:License|Copyright)\s*:\s*%[1]s)|(?:(?:covered )*under (?:the )?%[1]s)|(?:under (?:the terms of )*the %[1]s))`
+)
+
+func isValidLicenseName(licenseCheck string) bool {
+	for _, name := range licenseNames {
+		// (?i) case insensitive
```
hopkiw, line 132: this comment might not be relevant here anymore

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,190 @@
...
+		if re.MatchString(licenseCheck) {
+			return true
+		}
+	}
+	return false
+}
+
+func TestArePackagesLegal(t *testing.T) {
+	var filenames []string
+	if utils.IsTargetLinux(utils.RedHat) || utils.IsTargetLinux(utils.SUSE) {
```
hopkiw, line 157: i usually think a sequences of ifs against the same values are best handled with a switch statement. maybe like:
```
image, err := utils.GetMetadata("image")
if err != nil {
  t.Fatalf("couldn't get image from metadata")
}
switch {
case strings.Contains(image, "rhel"), strings.Contains(image, "suse"):
  pathGlob = licensePathGlob
default:
  pathGlob = copyrightPathGlob
}
filenames, err := filepath.Glob(pathGlob)
if err != nil {
  t.Fatalf("couldnt resolve glob %s", pathGlob)
}
for _, filename := range filenames {
```

hopkiw (changes requested):

imagetest/test_suites/image_validation/license_test.go
hopkiw: rather than bool, you could return error here, since you might have errors. then the check can simply be called like:
```
if err := validatePath(filename); err != nil {
  t.Fatalf(err.Error())
}
```

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,190 @@
...
+	var licenseCheck string = string(bytes)
+
+	// Remove comment
+	re := regexp.MustCompile(`(\*|#).`)
+	re.ReplaceAllString(licenseCheck, "")
+
+	// Replace all whitespace with one space
+	whitespaceRegex := regexp.MustCompile(`\s+`)
+	licenseCheck = whitespaceRegex.ReplaceAllString(licenseCheck, " ")
+	if !isValidLicenseName(licenseCheck) && isValidLicenseText(licenseCheck) {
```
hopkiw, line 186: i don't think this logic is correct. one has a ! and one doesn't, and you're using &&
it reads like you would want to say if !.. || !.. meaning if either one fail, short circuiting on the first failure

gaohannk: lost `!` for the second condition. tranlate from `!( ... || ...)` to `!... && !...`

hopkiw: i see, a package has only failed if both have returned false
maybe invert the logic and say if isValidLicenseName || isValidLicenseText { return true }
this way it can stop as soon as it reaches success, and easier to read

imagetest/utils/test_utils.go
gaohannk: I thought these strings are something like fixed, so make it const.

imagetest/test_suites/image_validation/license_test.go
gaohannk: Make the signature as` isPackageLegal(filepath string) (bool, error)`

imagetest/utils/test_utils.go
gaohannk: Agree to remove this util function and directly write it in test.

hopkiw (changes requested):

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,203 @@
...
+func isPackageLegal(filepath string) (bool, error) {
+	bytes, err := ioutil.ReadFile(filepath)
+	if err != nil {
+		return false, fmt.Errorf("error read file")
+	}
+
+	var licenseCheck string = string(bytes)
+
+	// Remove comment
+	re := regexp.MustCompile(`(\*|#).`)
```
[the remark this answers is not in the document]
gaohannk: I was thought the author purpose is to delete the comment start with `*` or `#`. But it seems the author purpose is just remove a line like `#####` or `***`

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,203 @@
...
+	re.ReplaceAllString(licenseCheck, "")
+
+	// Replace all whitespace with one space
+	whitespaceRegex := regexp.MustCompile(`\s+`)
+	licenseCheck = whitespaceRegex.ReplaceAllString(licenseCheck, " ")
+	if !isValidLicenseName(licenseCheck) && !isValidLicenseText(licenseCheck) {
```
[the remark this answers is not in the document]
gaohannk:
```
if isValid.. || isValid ..{
  t.log("pass test")
}
t.fatal("not pass")
```
```
if !isValid.. && !isValid ..{
   t.fatal("not pass")
}
 t.log("pass test")
```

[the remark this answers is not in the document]
gaohannk: i think what you what is to write the first, which is equivalent to second.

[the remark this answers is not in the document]
hopkiw: yes it's a matter of readability since these are all positive assertions. it is a nitpick, though.

imagetest/test_suites/image_validation/license_test.go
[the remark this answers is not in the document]
hopkiw: do you have any example file where you can see the difference as applied? e.g. if you get it wrong, will we be falsely marking files as passed, or falsely marking files as failed?

imagetest/test_suites/image_validation/license_test.go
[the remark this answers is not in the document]
hopkiw: original code:
```
if IsValidLicenseName(license_check) or IsValidLicenseText(license_check):    
    return True  
```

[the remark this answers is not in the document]
hopkiw: or return isValidLicense || isValidLicenseText, nil

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	var licenseCheck string = string(bytes)
+
+	// Remove comment
+	re := regexp.MustCompile(`(\*|#).`)
+	re.ReplaceAllString(licenseCheck, "")
+
+	// Replace all whitespace with one space
+	whitespaceRegex := regexp.MustCompile(`\s+`)
```
hopkiw, line 197: where is this in the original python code? i think that code joins lines, rather than replaces whitespace. regular expressions don't span lines

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -0,0 +1,203 @@
...
+func isPackageLegal(filepath string) (bool, error) {
+	bytes, err := ioutil.ReadFile(filepath)
+	if err != nil {
+		return false, fmt.Errorf("error read file")
+	}
+
+	var licenseCheck string = string(bytes)
+
+	// Remove comment
+	re := regexp.MustCompile(`(\*|#)*`)
```
hopkiw, line 193: don't need parens or asterisk at end, since you're using replaceall
simply
```
`\*|#`
```

imagetest/test_suites/image_validation/license_test.go
hopkiw: new code should be:
```
licenseCheck = strings.ReplaceAll(licenseCheck, "\n", " ")
```
to match original code:
```
license_check = ' '.join(license_check.split())
```

and remove whitespaceregex

hopkiw: what do you suggest for fix? what's missing in the whitespaceRegex is newlines. we must strip newlines or the regex will often fail in this new code where it would have succeeded in the old

PR #108, opened 2021-04-16
Image test boot

gaohannk:
<testsuites name="" errors="0" failures="0" tests="0" time="0">
	<testsuite name="image_validation-debian-10-buster-v20210316" tests="4" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestHostname" time="0"></testcase>
		<testcase classname="" name="TestFQDN" time="0"></testcase>
		<testcase classname="" name="TestHostKeysGeneratedOnce" time="0.02"></testcase>
		<testcase classname="" name="TestCustomHostname" time="0.01"></testcase>
	</testsuite>
	<testsuite name="image_boot-debian-10-buster-v20210316" tests="2" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestGuestBoot" time="0"></testcase>
		<testcase classname="" name="TestGuestReboot" time="0"></testcase>
	</testsuite>
</testsuites>


hopkiw (changes requested):
we should consider the effect of the default workflow timeout. in cases where the VM fails to boot, we are not likely to run the test and see the error message from it, we are more likely to see that as a workflow timeout or a missing results file.

imagetest/test_suites/imageboot/image_boot_test.go
```diff
@@ -0,0 +1,34 @@
...
+	err := syscall.Uname(&syscall.Utsname{})
+
+	if err != nil {
+		t.Fatalf("couldn't get system information, image boot failed")
+	}
+
+}
+
+func TestGuestReboot(t *testing.T) {
+	_, err := os.Stat("boot")
```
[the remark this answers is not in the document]
gaohannk: should be `/tmp/boot`

imagetest/test_suites/image_boot/setup.go
```diff
@@ -0,0 +1,15 @@
+package imageboot
+
+import "github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
+
+// Name is the name of the test package. It must match the directory name.
+var Name = "image-boot"
```
hopkiw, line 6: "It must match the directory name"

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+}
+
+func TestGuestReboot(t *testing.T) {
+	_, err := os.Stat("/tmp/boot")
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create("/tmp/boot"); err != nil {
+			t.Fatal("fail to create file when first boot")
+			return
```
hopkiw, line 40: np unreachable and unnecessary

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+		return
+	}
+
+	// second boot
+	if err = syscall.Sysinfo(&syscall.Sysinfo_t{}); err != nil {
```
hopkiw, line 46: why are you calling these linux-specific syscalls anyway? you don't do anything with it. isn't the fact this code is running proof enough that system calls work? and in this spot you've already confirmed the file exists, so a log message is probably fine

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestGuestReboot(t *testing.T) {
+	_, err := os.Stat("/tmp/boot")
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create("/tmp/boot"); err != nil {
+			t.Fatal("fail to create file when first boot")
+			return
+		}
+		return
```
hopkiw, line 42: i think the file not existing is a failure condition - you've left yourself open to passing the test when the file doesn't exist on second boot.

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Fatalf("couldn't get system information, image boot failed")
+	}
+
+}
+
+func TestGuestReboot(t *testing.T) {
+	_, err := os.Stat("/tmp/boot")
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create("/tmp/boot"); err != nil {
```
hopkiw, line 38: no need to use /tmp, we are the sole owners of the VM. and /tmp is ephemeral on some systems. recommend either / or /root directories for marker file creation

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,50 @@
...
+	flag.Parse()
+	if *runtest {
+		os.Exit(m.Run())
+	} else {
+		os.Exit(0)
+	}
+}
+
+func TestGuestBoot(t *testing.T) {
+	err := syscall.Uname(&syscall.Utsname{})
```
hopkiw, line 26: i think simply reaching this function fulfills the expectation and we can omit the syscall. then it will also be a good test for windows.

hopkiw (changes requested):

imagetest/test_suites/image_boot/image_boot_test.go
gaohannk: change to use `/`

imagetest/test_suites/image_boot/image_boot_test.go
gaohannk: I thought If the file doesn't exist on second boot. That implied the file are not created successfully during the first boot.  It must be failed in this line `t.Fatal("fail to create file when first boot")` during the first boot. So I `return`.

imagetest/test_suites/image_boot/image_boot_test.go
gaohannk: removed

imagetest/test_suites/image_boot/image_boot_test.go
gaohannk: The BCT test run a command `uname` as a signal for success boot. so I find an equivalent syscall like that cmd. But I agree that the test function itself is fulfills the expectation.

imagetest/test_suites/image_boot/image_boot_test.go
gaohannk: Same reason as above comment. I agree we don't need to run some command to signal, file itself is enough.

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: t.Fatal exits the test, so return not reachable. and you return in the outer loop anyway, so not necessary.

it could not exist for many reasons, so you must fail if the file does not exist. so the first boot will be a fail, the second boot will be a success.

gaohannk: Gotcha. Thanks for explanation.

hopkiw (changes requested):

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,40 @@
...
+	_, err := os.Stat("/boot-marker")
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create("/boot"); err != nil {
+			t.Fatalf("failed creating marker file: %v", err)
+		}
+		t.Fatal("marker file does not exist")
+	} else {
+		// second boot
+		t.Log("marker file exist signal the guest reboot successful")
```
hopkiw, line 36: the default outcome of the test is success. you don't need this else block

imagetest/test_suites/image_boot/setup.go
```diff
@@ -0,0 +1,15 @@
+package imageboot
+
+import "github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
+
+// Name is the name of the test package. It must match the directory name.
+var Name = "image_boot"
+
+// TestSetup sets up the test workflow.
+func TestSetup(t *imagetest.TestWorkflow) error {
+	vm, err := t.CreateTestVM("image-boot-test")
```
hopkiw, line 10: please name the vm 'vm' unless you're going to use the value. the workflow will already add image-boot to it.

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestGuestBoot(t *testing.T) {
+	t.Log("Guest booted successfully")
+}
+
+func TestGuestReboot(t *testing.T) {
+	_, err := os.Stat("/boot-marker")
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create("/boot-marker"); err != nil {
```
hopkiw, line 30: consider extracting to a constant

hopkiw (changes requested):

PR #121, opened 2021-04-22
Update readme

gaohannk:


hopkiw:
no i don't think we should encourage people to do this. that's what the dockerfile is for.

PR #139, opened 2021-05-12
image test concourse debian centos rhel

gaohannk:


concourse/pipelines/linux-image-build.yaml
```diff
@@ -741,6 +745,10 @@ jobs:
       publish_version: ((.:publish-version))
       wf: "enterprise_linux/centos_stream_8.publish.json"
       environment: "test"
+  - task: image-test-centos-stream-8
+    file: guest-test-infra/concourse/tasks/image-test.yaml
+    vars:
+      images: projects/bct-prod-images/global/images/family/centos-stream-8
```
hopkiw, line 751: instead of using family, can we specify the image? i think it should just be
```yaml
vars:
    images: projects/bct-prod-images/global/images/centos-stream-8-v((.:publish-version))
```
but you may want to confirm

gaohannk: Make this change apply to all image including the centos-7 image merged in previous PR.

PR #140, opened 2021-05-12
exit with error when test suites has failure or error

gaohannk:


imagetest/cmd/manager/main.go
```diff
@@ -159,8 +159,10 @@ func main() {
 	}
 
 	out, err := imagetest.RunTests(ctx, testWorkflows, *outPath, *project, *zone, *parallelCount)
+	if out != nil {
```
hopkiw, line 162: no we want this how we had it, this failure means a failure to run the tests. there's an idiomatic pattern in golang which is that the non-error output is not to be trusted or used when err is not nil.

imagetest/testworkflow.go
```diff
@@ -390,6 +397,9 @@ func RunTests(ctx context.Context, testWorkflows []*TestWorkflow, outPath, proje
 	outFile.Write(bytes)
 	outFile.Write([]byte{'\n'})
 
+	if suites.Errors != 0 || suites.Failures != 0 {
+		return bytes, fmt.Errorf("test suite has error or failure")
```
hopkiw, line 401: it is not a failure of `RunTests` if one of the tests has a failure status. let's make this function return the junit object rather than []byte, then caller can decide how to handle tests with errors

hopkiw (changes requested):

imagetest/cmd/manager/main.go
```diff
@@ -158,9 +160,25 @@ func main() {
...
 	}
 
-	out, err := imagetest.RunTests(ctx, testWorkflows, *outPath, *project, *zone, *parallelCount)
+	suites, err := imagetest.RunTests(ctx, testWorkflows, *project, *zone, *parallelCount)
 	if err != nil {
 		log.Fatalf("Failed to run tests: %v", err)
 	}
-	fmt.Printf("%s\n", out)
+	if suites.Errors != 0 || suites.Failures != 0 {
+		log.Fatalf("test suite has error or failure")
```
hopkiw, line 168: this will cause it to stop, and not display or write the junit output. i think we should log this but not with fatal, and then we should exit with an appropriate exit code at the end

hopkiw (changes requested):

imagetest/cmd/manager/main.go
gaohannk: Oh. right, I think we could just simple move the `fmt.Printf("%s\n", out)` ahead of this.

PR #143, opened 2021-05-14
merge agent integ test

hopkiw:
merge guest-agent integration tests into the `build-guest-agent` job.

the principle for whether an action should be captured in a distinct `job` is largely whether or not we would want to run that job independently of others. in the case of guest-agent integration tests, we always want to run them as a prerequisite to building packages of this version, and never independently. therefore, merge these steps into the build-package job just as unit tests are already included.


wrigri (approved):
/lgtm

PR #150, opened 2021-05-21
add test for alias ip

gaohannk:
Use default Network(auto mode VPC) for alias ip


imagetest/testworkflow_test.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 func TestAddStartStep(t *testing.T) {
-	twf, err := NewTestWorkflow("name", "image")
+	twf, err := NewTestWorkflow("name", "project", "zone", "image")
```
hopkiw, line 24: project and zone are not the responsibility of the test author; they should be added in finalize function

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,106 @@
...
+func TestAliasAgentRestart(t *testing.T) {
+	var networkInterface string
+
+	image, err := utils.GetMetadata("image")
+	if err != nil {
+		t.Fatalf("couldn't get image from metadata")
+	}
+
+	switch {
+	case strings.Contains(image, "debian-10") || strings.Contains(image, "ubuntu"):
```
hopkiw, line 84: , not ||

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,98 @@
...
+	expected, err := utils.GetMetadata("network-interfaces/0/ip-aliases/0/")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(actual) != string(expected) {
+		t.Fatalf("alias ip is not as expected after reboot, expected %s, actual %s", expected, actual)
+	}
+}
+
+func getAliasIP(networkInterface string) ([]byte, error) {
```
hopkiw, line 54: on reflection this function is more general than alias IPs, can be named 'getGoogleRoutes'

imagetest/fixtures.go
```diff
@@ -119,6 +119,15 @@ func (t *TestVM) Reboot() error {
 		return err
 	}
 
+	stopInstancesAfterRebootStep, err := t.testWorkflow.addStopStep("stoped-"+t.name, t.name)
```
hopkiw, line 122: typo 'stopped'

imagetest/fixtures.go
```diff
@@ -119,6 +119,15 @@ func (t *TestVM) Reboot() error {
...
+	if err != nil {
+		return err
+	}
+
+	if err := t.testWorkflow.wf.AddDependency(stopInstancesAfterRebootStep, waitStartedStep); err != nil {
+		return err
+	}
+
```
hopkiw, line 130: the result of this new block will be that every vm that has a reboot added will be stopped at the end. i don't think this is correct.

imagetest/testworkflow.go
```diff
@@ -99,7 +99,75 @@ func (t *TestWorkflow) CreateTestVM(name string) (*TestVM, error) {
 	return &TestVM{name: vmname, testWorkflow: t}, nil
 }
 
-func (t *TestWorkflow) appendCreateVMStep(name, hostname string) (*daisy.Step, error) {
+// CreateTestVMWithCustomNetwork creates the necessary steps to create a VM with  the specified name and custom network to the workflow.
+func (t *TestWorkflow) CreateTestVMWithCustomNetwork(name, networkName,
```
hopkiw, line 103: the model we have used so far is to call createvm and then call methods on the returned vm object that modify it, for example to start it with secure boot enabled. i don't think we want a pattern of many custom-vm-type creation methods

imagetest/testworkflow.go
```diff
@@ -99,7 +99,75 @@ func (t *TestWorkflow) CreateTestVM(name string) (*TestVM, error) {
...
+	}
+
+	if err := t.wf.AddDependency(waitStep, createVMStep); err != nil {
+		return nil, err
+	}
+
+	return &TestVM{name: vmname, testWorkflow: t}, nil
+}
+
+func (t *TestWorkflow) appendCreateVMStep(name, hostname string, networkInterface []*compute.NetworkInterface) (*daisy.Step, error) {
```
hopkiw, line 170: yeah now you're making it so the network is a required parameter, forcing you to pass nil to get the default network. the paradigm in these fixtures is to fully populate the steps with fixtures over time, then finalize them. you can add these network details to the step *after* it has been created in the workflow.

hopkiw (changes requested):
i think the architectural approach here should be changed. let's review offline

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if actual != expected {
+		t.Fatalf("alias ip is not as expected after reboot, expected %s, actual %s", expected, actual)
+	}
+}
+
+func getGoogleRoutes(networkInterface string) (string, error) {
+	arguments := strings.Split(fmt.Sprintf("route list table local type local scope host dev %s proto 66", networkInterface), " ")
```
hopkiw, line 54: this command can return multi-line output

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,97 @@
...
+	cmd := exec.Command("systemctl", "restart", "google-guest-agent")
+	_, err = cmd.Output()
+	if err != nil {
+		t.Fatal(err)
+	}
+	afterRestart, err := getGoogleRoutes(networkInterface)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if beforeRestart != afterRestart {
```
hopkiw, line 94: comparison is good but we should check the specific route is also present

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		t.Fatalf("couldn't get image from metadata")
+	}
+
+	switch {
+	case strings.Contains(image, "debian-10"), strings.Contains(image, "ubuntu"):
+		networkInterface = defaultPredictableInterface
+	default:
+		networkInterface = defaultInterface
+	}
```
hopkiw, line 29: do this after the marker file check

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,97 @@
...
+	_, err = os.Stat(markerFile)
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create(markerFile); err != nil {
+			t.Fatalf("failed creating marker file: %v", err)
+		}
+		t.Fatalf("failed on first boot")
+	}
+	// second boot
+	actual, err := getGoogleRoutes(networkInterface)
```
hopkiw, line 40: you only check after, you should also check before

imagetest/fixtures.go
```diff
@@ -156,3 +156,51 @@ func (t *TestVM) EnableSecureBoot() {
 		}
 	}
 }
+
+// EnableNetwork ensure the create network step run before create test VMs and
+// current test VMs in workflow use it
+func (t *TestVM) EnableNetwork(networkName, subnetworkName, aliasIPRange, rangeName string) error {
```
hopkiw, line 162: this function does not need to modify the dependencies

imagetest/fixtures.go
```diff
@@ -156,3 +156,51 @@ func (t *TestVM) EnableSecureBoot() {
...
+				},
+				AliasIpRanges: []*compute.AliasIpRange{
+					{
+						IpCidrRange:         aliasIPRange,
+						SubnetworkRangeName: rangeName,
+					},
+				},
+			}
+
+			i.NetworkInterfaces = append(i.NetworkInterfaces, &network)
```
hopkiw, line 201: do we intend to support multiple network interfaces? what happens if someone calls EnableNetwork 2 times?

imagetest/testworkflow.go
```diff
@@ -225,6 +225,70 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+	pFalse := false
+	network := &daisy.Network{
+		Network: compute.Network{
+			Name: networkName,
+		},
+		AutoCreateSubnetworks: &pFalse,
+	}
+	*createNetworks = append(*createNetworks, network)
+
+	createNetworksStep, err := t.wf.NewStep("create-network-" + vmname)
```
hopkiw, line 239: creating a network is not specific to a VM; it should be named after the network name, not the VM name

imagetest/testworkflow.go
```diff
@@ -225,6 +225,70 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+		Subnetwork: compute.Subnetwork{
+			Name:        subnetworkName,
+			IpCidrRange: primaryRange,
+			Network:     networkName,
+			SecondaryIpRanges: []*compute.SubnetworkSecondaryRange{
+				{
+					IpCidrRange: secondaryRange,
+					RangeName:   rangeName,
+				},
+			},
```
hopkiw, line 260: aren't secondary ranges optional?

imagetest/testworkflow.go
```diff
@@ -225,6 +225,70 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+		},
+	}
+	*createSubnetworks = append(*createSubnetworks, subnetwork)
+
+	createSubnetworksStep, err := t.wf.NewStep("create-sub-network-" + vmname)
```
hopkiw, line 265: here also, we don't need vm name

hopkiw: actually, we should use 'shared' `create-networks` and `create-subnets` steps.

hopkiw (changes requested):

imagetest/fixtures.go
```diff
@@ -156,3 +158,39 @@ func (t *TestVM) EnableSecureBoot() {
...
 }
+
+// SetCustomNetwork set current test VMs in workflow using provided network and
+// subnetwork.
+func (t *TestVM) SetCustomNetwork(networkName, subnetworkName string) error {
+	for _, i := range t.testWorkflow.wf.Steps[createVMsStepName].CreateInstances.Instances {
+		if i.Name == t.name {
+			networkInterface := compute.NetworkInterface{
+				Network:    networkName,
+				Subnetwork: subnetworkName,
```
hopkiw, line 169: what happens if this is omitted?

imagetest/testworkflow.go
```diff
@@ -225,6 +254,118 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+	subnetwork := &daisy.Subnetwork{
+		Subnetwork: compute.Subnetwork{
+			Name:        name,
+			IpCidrRange: ipRange,
+			Network:     n.Name,
+		},
+	}
+	*createSubnetworks = append(*createSubnetworks, subnetwork)
+
+	createSubnetworksStep, err := n.testWorkflow.wf.NewStep(createSubNetworkStepName)
```
hopkiw, line 316: i think we always want to use the one create-subnets step  (you created a constant for the name), instead of creating new steps each time

hopkiw: you do need to create the step the first time, if it doesnt exist. otherwise find and append to it, like we do with create-vms

imagetest/testworkflow.go
```diff
@@ -52,6 +52,35 @@ type TestWorkflow struct {
 	counter int
 }
 
+// Network defines the parameter which used by compute network.
+type Network struct {
+	Name         string
+	autoMode     bool
```
hopkiw, line 58: this is not used

imagetest/testworkflow.go
```diff
@@ -52,6 +52,35 @@ type TestWorkflow struct {
...
+	testWorkflow *TestWorkflow
+}
+
+// SubNetwork defines the parameter used by compute subnetwork.
+type SubNetwork struct {
+	Network        *Network
+	Name           string
+	primaryIPRange string
```
hopkiw, line 66: just like the network struct includes a pointer to its test workflow, i think the subnet should or should include a pointer to its network

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+func TestAliasAfterReboot(t *testing.T) {
+	_, err := os.Stat(markerFile)
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create(markerFile); err != nil {
+			t.Fatalf("failed creating marker file: %v", err)
+		}
+		if err := verifyIPAliases(); err != nil {
```
hopkiw, line 24: to test the aliases are right on the first boot, i think we should extract that to a second test function TestAliasOnBoot or something. marker-file reboots must always FAIL their first boot.

gaohannk:
```
 "create-network": {
      "CreateNetworks": [
        {
          "name": "test-net",
          "autoCreateSubnetworks": false
        },
        {
          "name": "test-2",
          "autoCreateSubnetworks": false
        }
      ]
    },
    "create-sub-network": {
      "CreateSubnetworks": [
        {
          "ipCidrRange": "192.168.209.0/24",
          "name": "test-subnet",
          "network": "test-net",
          "secondaryIpRanges": [
            {
              "rangeName": "secondary-range"
            }
          ]
        },
        {
          "ipCidrRange": "192.168.209.0/24",
          "name": "test-sub2",
          "network": "test-2",
          "secondaryIpRanges": [
            {
              "ipCidrRange": "123",
              "rangeName": "secondary-range"
            }
          ]
        }
      ]
    },
```
```
network, err := t.CreateNetwork(networkName, false)
	if err != nil {
		return err
	}
	network2, err := t.CreateNetwork("test-2", false)
	if err != nil {
		return err
	}
	subNetwork, err := network.CreateSubNetwork(subnetworkName, primaryIPRange)
	if err != nil {
		return err
	}
	subNetwork2, err := network2.CreateSubNetwork("test-sub2", primaryIPRange)
	if err != nil {
		return err
	}
	if err := subNetwork2.AddSecondaryRange(rangeName, "123"); err != nil {
		return err
	}

	if err := subNetwork.AddSecondaryRange(rangeName, ""); err != nil {
		return err
	}

	vm2, err := t.CreateTestVM(vm2)
	if err != nil {
		return err
	}
	if err := vm2.SetCustomNetwork(networkName, subnetworkName); err != nil {
		return err
	}
	if err := vm2.AddAliasIPRanges(aliasIPRange, rangeName); err != nil {
		return err
	}
	if err := vm2.Reboot(); err != nil {
		return err
	}
```

gaohannk:
```
          "StartupScript": "wrapper",
            "disks": [
              {
                "source": "vm2"
              }
            ],
            "name": "vm2",
            "networkInterfaces": [
              {
                "accessConfigs": [
                  {
                    "type": "ONE_TO_ONE_NAT"
                  }
                ],
                "aliasIpRanges": [
                  {
                    "ipCidrRange": "10.14.1.0/24",
                    "subnetworkRangeName": "secondary-range"
                  }
                ],
                "network": "test-net"
              }
            ],
```

imagetest/fixtures.go
gaohannk: the subnetwork field will not populated in daisy workflow

imagetest/testworkflow.go
```diff
@@ -225,6 +264,97 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+		if err != nil {
+			return nil, err
+		}
+		// Use default subnetwork, no create custom subnetwork step.
+		if autoCreateSubnetworks {
+			firstStep, ok := t.wf.Steps[createDisksStepName]
+			if !ok {
+				return nil, fmt.Errorf("failed resolve first step")
+			}
+			if err := t.wf.AddDependency(firstStep, createNetworkStep); err != nil {
```
hopkiw, line 347: i think the create-vms step should be the one to depend on create-networks. creating networks and disks can happen in parallel

imagetest/testworkflow.go
```diff
@@ -225,6 +264,97 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+		AutoCreateSubnetworks: &autoCreateSubnetworks,
+	}
+	createNetworkStep, ok := t.wf.Steps[createNetworkStepName]
+	if !ok {
+		createNetworkStep, err := t.addCreateNetworkStep(network)
+		if err != nil {
+			return nil, err
+		}
+		// Use default subnetwork, no create custom subnetwork step.
+		if autoCreateSubnetworks {
```
hopkiw, line 342: there's no reason to wrap this conditional; it'll be fine if the create-vms step depends on create-networks AND create-subnets

imagetest/testworkflow.go
```diff
@@ -225,6 +264,97 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+			Network:     n.name,
+		},
+	}
+	createSubnetworksStep, ok := n.testWorkflow.wf.Steps[createSubNetworkStepName]
+	if !ok {
+		createSubnetworksStep, err := n.testWorkflow.addCreateSubnetworksStep(subnetwork)
+		if err != nil {
+			return nil, err
+		}
+		firstStep, ok := n.testWorkflow.wf.Steps[createDisksStepName]
```
hopkiw, line 305: here also i think create-vms is the right step to add the dependency for

imagetest/testworkflow.go
```diff
@@ -52,6 +52,45 @@ type TestWorkflow struct {
...
+		return err
+	}
+	subnetwork.SecondaryIpRanges = append(subnetwork.SecondaryIpRanges, &compute.SubnetworkSecondaryRange{
+		IpCidrRange: ipRange,
+		RangeName:   rangeName,
+	})
+	return nil
+}
+
+func (s SubNetwork) findSubNetwork(subnetworks daisy.CreateSubnetworks) (*daisy.Subnetwork, error) {
```
hopkiw, line 85: in all other cases we do the matching loop inline: in `AddMetadata`, in `EnableSecureBoot`, in `SetCustomNetwork`, etc.

but if you really want to make this a method, don't pass in subnetworks. look it up here as you do on line 69 in AddSecondaryRange above.

imagetest/fixtures.go
```diff
@@ -22,9 +22,11 @@ import (
 )
 
 const (
-	createVMsStepName   = "create-vms"
-	createDisksStepName = "create-disks"
-	successMatch        = "FINISHED-TEST"
+	createVMsStepName        = "create-vms"
+	createDisksStepName      = "create-disks"
+	createNetworkStepName    = "create-network"
+	createSubNetworkStepName = "create-sub-network"
```
hopkiw, line 28:
```suggestion
	createNetworkStepName    = "create-networks"
	createSubNetworkStepName = "create-subnets"
```

hopkiw: these should be plural, since they're shared steps

hopkiw (changes requested):

imagetest/testworkflow.go
```diff
@@ -225,6 +264,97 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
 	return startInstancesStep, nil
 }
 
+func (t *TestWorkflow) addCreateNetworkStep(network *daisy.Network) (*daisy.Step, error) {
+	createNetworks := &daisy.CreateNetworks{}
+	*createNetworks = append(*createNetworks, network)
+	createNetworkStep, err := t.wf.NewStep(createNetworkStepName)
```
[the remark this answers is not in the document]
gaohannk: I check the safety in the upper level function `CreateNetwork` where call `addCreateNetworkStep`. To follow the paradigm. I move the check in `addCreateNetworkStep` and rename to `appendCreateNetworkStep`

imagetest/testworkflow.go
gaohannk: Good points. creating networks and disks and be parallel.

imagetest/testworkflow.go
gaohannk: It's no harmful to add this dependency as we will always make sure 
`create-vms step` depends on `create-sub-networks` and `create-sub-networks` depends on  `create-networks`.
when create-subnetwork is exist

imagetest/testworkflow.go
gaohannk: Follow the paradigm in `AddMetadata`, in` EnableSecureBoot`, in `SetCustomNetwork`. just make the `AddSecondaryRange` as the these method without return error

gaohannk:
rojects/hanga-testing/global/firewalls/test-net-network-w5stp-khxhi7it6wpyqyni2dgtcxcx'
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="4" time="0">
	<testsuite name="network-debian-10" tests="4" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestDefaultMTU" time="0"></testcase>
		<testcase classname="" name="TestAliasAfterOnBoot" time="0"></testcase>
		<testcase classname="" name="TestAliasAfterReboot" time="0"></testcase>
		<testcase classname="" name="TestAliasAgentRestart" time="0.17"></testcase>
	</testsuite>
</testsuites>

imagetest/testworkflow.go
```diff
@@ -225,6 +249,104 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
...
+// TestVM to config network on vm
+func (t *TestWorkflow) CreateNetwork(networkName string, autoCreateSubnetworks bool) (*Network, error) {
+	network := &daisy.Network{
+		Network: compute.Network{
+			Name: networkName,
+		},
+		AutoCreateSubnetworks: &autoCreateSubnetworks,
+	}
+
+	createNetworkStep, err := t.appendCreateNetworkStep(network)
```
hopkiw, line 334:
```suggestion
	createNetworkStep, err := t.appendCreateNetworkStep(networkName, autoCreateSubnetworks)
```

imagetest/testworkflow.go
```diff
@@ -225,6 +249,104 @@ func (t *TestWorkflow) addStartStep(stepname, vmname string) (*daisy.Step, error
 	return startInstancesStep, nil
 }
 
+func (t *TestWorkflow) appendCreateNetworkStep(network *daisy.Network) (*daisy.Step, error) {
+	createNetworks := &daisy.CreateNetworks{}
+	*createNetworks = append(*createNetworks, network)
```
hopkiw, line 254:
```suggestion
func (t *TestWorkflow) appendCreateNetworkStep(networkName string, autoCreateSubnetworks bool) (*daisy.Step, error) {
	network := &daisy.Network{
		Network: compute.Network{
			Name: networkName,
		},
		AutoCreateSubnetworks: &autoCreateSubnetworks,
	}

	createNetworks := &daisy.CreateNetworks{}
	*createNetworks = append(*createNetworks, network)
```

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+const (
+	markerFile = "/boot-marker"
+)
+
+func TestAliasAfterOnBoot(t *testing.T) {
```
hopkiw, line 17:
```suggestion
func TestAliases(t *testing.T) {
```

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,141 @@
...
+		t.Fatal(err)
+	}
+}
+
+func verifyIPExist(afterRestart []string) error {
+	expected, err := utils.GetMetadata("network-interfaces/0/ip-aliases/0")
+	if err != nil {
+		return fmt.Errorf("couldn't get image from metadata: %v", err)
+	}
+	for _, ip := range afterRestart {
```
hopkiw, line 123:
```suggestion
func verifyIPExist(routes []string) error {
	expected, err := utils.GetMetadata("network-interfaces/0/ip-aliases/0")
	if err != nil {
		return fmt.Errorf("couldn't get first alias IP from metadata: %v", err)
	}
	for _, route := range routes {
```

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,141 @@
...
+	default:
+		networkInterface = defaultInterface
+	}
+
+	beforeRestart, err := getGoogleRoutes(networkInterface)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cmd := exec.Command("systemctl", "restart", "google-guest-agent")
+	_, err = cmd.Output()
```
hopkiw, line 102:
```suggestion
	err = cmd.Run()
```

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func verifyIPExist(afterRestart []string) error {
+	expected, err := utils.GetMetadata("network-interfaces/0/ip-aliases/0")
+	if err != nil {
+		return fmt.Errorf("couldn't get image from metadata: %v", err)
+	}
+	for _, ip := range afterRestart {
+		if ip == expected {
```
hopkiw, line 124: did you test this? i don't think the format from ip-aliases key usually will match the output from `ip` command; usually the metadata key has a mask suffix

hopkiw (changes requested):

imagetest/test_suites/network/alias_ip_test.go
gaohannk: Yes, the metadata key value and `ip` command (`local 10.14.1.0/24`) both have the mask suffix

imagetest/fixtures.go
gaohannk: I test this, it will just not populate subnetwork field. `"subnetwork": "test-subnet"`
> When `subnetwork = ""`, the instance will not populate subnetwork field.
> 
> ```
>           "StartupScript": "wrapper",
>             "disks": [
>               {
>                 "source": "vm2"
>               }
>             ],
>             "name": "vm2",
>             "networkInterfaces": [
>               {
>                 "accessConfigs": [
>                   {
>                     "type": "ONE_TO_ONE_NAT"
>                   }
>                 ],
>                 "aliasIpRanges": [
>                   {
>                     "ipCidrRange": "10.14.1.0/24",
>                     "subnetworkRangeName": "secondary-range"
>                   }
>                 ],
>                 "network": "test-net"
>               }
>             ],
> ```

hopkiw: what will happen in the workflow if it is not populated?

gaohannk: The workflow will just create vm with provided network with default subnetwork which has one subnetwork (subnet) in each region.

hopkiw: in this case we created a network without automatic subnetting; won't omitting this value cause the workflow to fail in that case?
also with discussions, dont resolve til we've both finished with the discussion.

imagetest/fixtures.go
```diff
@@ -156,3 +158,38 @@ func (t *TestVM) EnableSecureBoot() {
...
+			}
+
+			i.NetworkInterfaces = []*compute.NetworkInterface{&networkInterface}
+			break
+		}
+	}
+}
+
+// AddAliasIPRanges add alias ip range to current test VMs.
+func (t *TestVM) AddAliasIPRanges(aliasIPRange, rangeName string) error {
```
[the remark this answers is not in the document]
hopkiw: i gave this feedback expecting you'd accept it, not silently resolve it.

test.workflow
```diff
@@ -0,0 +1,99 @@
+{
```
hopkiw, line 1: dont think we need this file

hopkiw (changes requested):
please make the requested changes.

imagetest/fixtures.go
gaohannk: no, the workflow is still valid in daisy and is still runnable.

hopkiw: what subnet would it pick ? from docs: " If the network is in custom subnet mode, specifying the subnetwork is required."
https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert

gaohannk: oh. I see. you are talking about we set autoCreateSubnetworks to true, but not provide subnetwork. In this case, 
```
	<testsuite name="network-debian-10" tests="1" failures="0" errors="1" disabled="0" skipped="0" time="0">
		<system-err>step &#34;create-vms&#34; run error: googleapi: Error 400: Invalid value for field &#39;resource.networkInterfaces[0]&#39;: &#39;{  &#34;network&#34;: &#34;projects/hanga-testing/global/networks/test-net-network-9z878&#34;,  &#34;accessConfig&#34;: [{  ...&#39;. Subnetwork should be specified for custom subnetmode network, invalid</system-err>
	</testsuite>
```

hopkiw: should we check for this case in this method?

gaohannk: Add a check for this case and return error
```
2021/07/15 23:29:14 network.TestSetup for projects/debian-cloud/global/images/family/debian-10 failed: network test-net with custom mode must provide subnetwork
```

PR #151, opened 2021-05-21
add disk resize test

gaohannk:


imagetest/test_suites/disk/disk_resize_test.go
```diff
@@ -0,0 +1,68 @@
...
+		os.Exit(m.Run())
+	} else {
+		os.Exit(0)
+	}
+}
+
+// TestDiskResize Validate the filesystem is resized on reboot after a disk resize.
+func TestDiskResize(t *testing.T) {
+	_, err := os.Stat(markerFile)
+	if os.IsNotExist(err) {
```
hopkiw, line 36: in this test i think you can just check if the disk is the correct post-resize size in both boots. it'll upload a failure result on the first boot, and (in theory) a success result on the second boot. then you can avoid this first-boot/second-boot switch

imagetest/test_suites/disk/disk_resize_test.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+	// second boot
+	if err := verifyDiskSize(resizeDiskSize); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func verifyDiskSize(expectedGb int) error {
+	cmd := exec.Command("df", "-k", "/")
```
hopkiw, line 52: there are go native ways to find filesystem size using sys.unix.Statfs, no need to shell out

imagetest/fixtures.go
```diff
@@ -133,3 +133,22 @@ func (t *TestVM) EnableSecureBoot() {
...
+	if !ok {
+		return fmt.Errorf("wait-%s step missing", t.name)
+	}
+
+	diskResizeStep, err := t.testWorkflow.addResizeDisk(t.name, t.name, int64(diskSize))
+	if err != nil {
+		return err
+	}
+
+	if err := t.testWorkflow.wf.AddDependency(diskResizeStep, waitStep); err != nil {
```
hopkiw, line 150: you add the new step and with this dependency, you ensure it will run after the VM is done running the test. but in this test you want to boot VM -> run test -> resize disk -> reboot VM -> run test -> wait for test completion. how will you guarantee the reboot happens *after* the resize?

when you're developing new fixtures, use the -print option to see what the resulting workflow looks like.

imagetest/test_suites/disk/disk_resize_test.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return err
+	}
+	diskSizeLine := strings.Split(string(b), "\n")[1]
+	diskSize, err := strconv.Atoi(strings.Fields(diskSizeLine)[1])
+	if err != nil {
+		return err
+	}
+	expectedSize := expectedGb * gb
+	maxDiff := float64(expectedSize * 1 / 100)
```
hopkiw, line 63: is there a difference between expectedSize * 1 / 100 and expectedSize / 100...

hopkiw (changes requested):

imagetest/testworkflow.go
```diff
@@ -434,3 +451,25 @@ func parseResult(res testResult) *testSuite {
...
+	lastStepName := resolveStep(reverseDependencies, createDisksStepName)
+	return t.wf.Steps[lastStepName], nil
+}
+
+func resolveStep(reverseDependencies map[string]string, currentStep string) string {
+	if _, ok := reverseDependencies[currentStep]; !ok {
+		return currentStep
+	}
+	return resolveStep(reverseDependencies, reverseDependencies[currentStep])
+}
```
hopkiw, line 475: there are a few issues with this algorithm.

first, you will never be able to create the reverse-dependency map on any workflow that has more than one VM, because the wait-$VMNAME steps all depend on the create-vms step.

second, you are looking for the last step of the whole workflow, but we need the last step for a given VM. third, using recursion this way has some risky rough edges. you can get infinite recursion in some scenarios and you've got no safeguards here for it. here is a lengthy proposed alternative:

```suggestion
func  (t *TestWorkflow) getLastStepForVM(vmname string) (*daisy.Step, error) {
	step := "wait-" + vmname
	if _, ok := t.wf.Dependencies[step]; !ok {
		return nil, fmt.Errorf("no step " + step)
	}
	rdeps := make(map[string][]string)
	for dependent, dependencies := range t.wf.Dependencies {
		for _, dependency := range dependencies {
			rdeps[dependency] = append(rdeps[dependency], dependent)
		}
	}

	for {
		deps, ok := rdeps[step]
		if !ok {
			// no more steps depend on this one
			break
		}
		if len(deps) > 1 {
			return nil, fmt.Errorf("Workflow has non-linear dependencies!")
		}
		step = deps[0]
	}

	return t.wf.Steps[step], nil

}

```

hopkiw (changes requested):

imagetest/testworkflow.go
gaohannk: Good catch. To work for more than one vm, should start specific VM which could be start at `wait-vm` step

PR #152, opened 2021-06-03
add shutdown script test

gaohannk:
```
image_boot git:(image-test-shutdown-test) ✗ /manager -project hanga-testing -zone us-west1-c -images projects/debian-cloud/global/images/family/debian-10
2021/06/03 20:05:42 imagetest: Done with setup
runTestWorkflow: running image_boot on debian-10 (ID vqbhg)
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="1" time="0">
	<testsuite name="image_boot-debian-10" tests="1" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestGuestShutdownScript" time="0"></testcase>
	</testsuite>
</testsuites>
```

### DIFF
--- a/imagetest/test_suites/image_boot/setup.go
+++ b/imagetest/test_suites/image_boot/setup.go
@@ -16,13 +16,14 @@ done`
 // TestSetup sets up the test workflow.
 func TestSetup(t *imagetest.TestWorkflow) error {
 	vm, err := t.CreateTestVM("vm")
+	vm.SetShutdownScript(script)
 	if err != nil {
 		return err
 	}
 	if err := vm.Reboot(); err != nil {
 		return err
 	}
-	vm.RunTests("TestGuestBoot|TestGuestReboot")
+	vm.RunTests("TestGuestBoot|TestGuestReboot|TestGuestShutdownScript")
 
 	vm2, err := t.CreateTestVM("vm2")
 	if err != nil {
@@ -31,15 +32,5 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 
 	vm2.EnableSecureBoot()
 	vm2.RunTests("TestGuestSecureBoot")
-
-	vm3, err := t.CreateTestVM("vm3")
-	if err != nil {
-		return err
-	}
-	vm3.SetShutdownScript(script)
-	if err := vm3.Reboot(); err != nil {
-		return err
-	}
-	vm3.RunTests("TestGuestShutdownScript")
 	return nil
 }
--- a/imagetest/test_suites/image_boot/setup.go
+++ b/imagetest/test_suites/image_boot/setup.go
@@ -5,6 +5,14 @@ import "github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
 // Name is the name of the test package. It must match the directory name.
 var Name = "image_boot"
 
+const script = `#!/bin/bash
+
+while [[ 1 ]]; do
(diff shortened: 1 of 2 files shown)


imagetest/test_suites/image_boot/shutdown_test.go
```diff
@@ -0,0 +1,31 @@
...
+			t.Fatalf("failed creating marker file: %v", err)
+		}
+		t.Fatal("marker file does not exist")
+	}
+	// second boot
+	bytes, err := ioutil.ReadFile("shutdown.txt")
+	if err != nil {
+		t.Errorf("error reading file: %v", err)
+	}
+	lines := strings.Split(strings.TrimRight(string(bytes), ` \t\r\n\0`), "\n")
```
hopkiw, line 27: strings.Trimspace?

imagetest/test_suites/image_boot/shutdown_test.go
```diff
@@ -0,0 +1,31 @@
...
+	_, err := os.Stat(markerFile)
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create(markerFile); err != nil {
+			t.Fatalf("failed creating marker file: %v", err)
+		}
+		t.Fatal("marker file does not exist")
+	}
+	// second boot
+	bytes, err := ioutil.ReadFile("shutdown.txt")
```
hopkiw, line 23: no path means in current directory. every run is in its own directory. as with our other reboot tests, extract the file path to a constant and put it in / or /root/

imagetest/test_suites/image_boot/setup.go
```diff
@@ -23,5 +31,15 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 
 	vm2.EnableSecureBoot()
 	vm2.RunTests("TestGuestSecureBoot")
+
+	vm3, err := t.CreateTestVM("vm3")
+	if err != nil {
+		return err
+	}
+	vm3.SetShutdownScript(script)
```
hopkiw, line 39: do we need a new VM for it?

hopkiw (changes requested):

imagetest/test_suites/image_boot/setup.go
gaohannk: Using vm1 for this test.

imagetest/test_suites/image_boot/shutdown_test.go
```diff
@@ -0,0 +1,31 @@
...
+// TestGuestShutdownScript test that shutdown scripts can run for around two minutes
+func TestGuestShutdownScript(t *testing.T) {
+	_, err := os.Stat(markerFile)
+	if os.IsNotExist(err) {
+		// first boot
+		if _, err := os.Create(markerFile); err != nil {
+			t.Fatalf("failed creating marker file: %v", err)
+		}
+		t.Fatal("marker file does not exist")
+	}
```
hopkiw, line 21: 2 issues

1. this test AND `TestGuestReboot` are both going to try to create the same marker file on the same VM. this could be race condition where that test creates the file, so this test thinks it is already the second boot.
2. we already have an implicit marker file on this test - the file `/shutdown.txt`. let's just use that since we know it only gets created when the shutdown script runs. and since we already have error handling for it not existing, we can write a simpler test by just omitting this section entirely.

```suggestion

```

PR #154, opened 2021-06-09
Move rhel metal dev build/publish to bare metal pipeline.

zmarano:


zmarano:
Goes with https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1661

PR #162, opened 2021-06-24
Fix several image tests.

zmarano:
/assign @gaohannk

### DIFF
--- a/imagetest/test_suites/disk/disk_resize_test.go
+++ b/imagetest/test_suites/disk/disk_resize_test.go
@@ -9,6 +9,8 @@ import (
 	"math"
 	"os"
 	"testing"
+
+  "github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
 const (
--- a/imagetest/test_suites/disk/disk_resize_test.go
+++ b/imagetest/test_suites/disk/disk_resize_test.go
@@ -18,6 +18,7 @@ const (
 	markerFile      = "/boot-marker"
 	gb              = 1024.0 * 1024.0 * 1024.0
 	defaultDiskSize = 20
+  resizeDiskSize  = 200
 )
 
 var (
@@ -44,7 +45,7 @@ func TestDiskResize(t *testing.T) {
 		t.Skip("disk expansion not supported on RHEL 7.4")
 	}
 
-	_, err := os.Stat(markerFile)
+	_, err = os.Stat(markerFile)
 
 	if os.IsNotExist(err) {
 		// first boot
--- a/imagetest/test_suites/disk/disk_resize_test.go
+++ b/imagetest/test_suites/disk/disk_resize_test.go
@@ -18,7 +18,6 @@ const (
 	markerFile      = "/boot-marker"
 	gb              = 1024.0 * 1024.0 * 1024.0
 	defaultDiskSize = 20
-	resizeDiskSize  = 200
 )
 
 var (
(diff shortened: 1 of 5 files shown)


imagetest/test_suites/disk/disk_resize_test.go
```diff
@@ -32,19 +35,28 @@ func TestMain(m *testing.M) {
...
 func TestDiskResize(t *testing.T) {
-	_, err := os.Stat(markerFile)
+	image, err := utils.GetMetadata("image")
+	if err != nil {
+		t.Fatalf("couldn't get image from metadata")
+	}
+	switch {
+	case strings.Contains(image, "rhel-7-4-sap"):
+		t.Skip("disk expansion not supported on RHEL 7.4")
+	default:
```
hopkiw, line 45: code execution halts at t.Skip, so no reason to capture it in a switch statement like this.

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -43,15 +46,25 @@ func TestGuestReboot(t *testing.T) {
...
+		t.Fatalf("couldn't get image from metadata")
 	}
-	// https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-firmware-efi-vars
-	if data[0] != 1 {
-		t.Fatal("secure boot is not enabled as expected")
+
+	switch {
+	case strings.Contains(image, "debian-9"):
+		t.Skip("secure boot is not supported on Debian 9")
+	default:
```
hopkiw, line 57: same feedback, no reason to use switch statements like this.

hopkiw (changes requested):

imagetest/test_suites/disk/disk_resize_test.go
zmarano: Yeah that makes sense, fixed.

imagetest/test_suites/image_boot/image_boot_test.go
zmarano: Ditto

imagetest/test_suites/security/image_security_test.go
```diff
@@ -272,7 +272,7 @@ func verifyAutomaticUpdate(image string) error {
 		return err
 	}
 	switch {
-	case strings.Contains(image, "debian-9"):
+	case strings.Contains(image, "debian"):
```
gaohannk, line 275: Does this also valid for debian-10?

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -117,6 +118,7 @@ var licenses = []string{
 	`Open Market permits you to use, copy, modify, distribute, and license this Software and the Documentation for any purpose, provided that existing copyright notices are retained in all copies and that this notice is included verbatim in any distributions. No written agreement, license, or royalty fee is required for any of the authorized uses.`,
 	`This software is made available under the terms of *either* of the licenses found in LICENSE.APACHE or LICENSE.BSD. Contributions to cryptography are made under the terms of *both* these licenses.`,
 	`This library.*is public domain software`,
+	`This Font Software is licensed under the SIL Open Font License`,
```
hopkiw, line 121: maybe a nitpick, but we usually cover these files with EITHER a license name OR a license text check; you've added both but i think one will suffice. usually we prefer license name, i think

imagetest/test_suites/security/image_security_test.go
zmarano: Yup- at least in my test cases it was valid for all debian images (9-11)

PR #166, opened 2021-06-25
Show failing licenses

zmarano:


imagetest/test_suites/security/image_security_test.go
```diff
@@ -271,12 +271,12 @@ func verifyAutomaticUpdate(image string) error {
...
 		return err
 	}
-	switch {
-	case strings.Contains(image, "debian-9"):
+	if strings.Contains(image, "debian-9") {
 		if !strings.Contains(automaticUpdateConfig, `APT::Periodic::Enable "1";`) {
 			return fmt.Errorf(`"APT::Periodic::Enable" is not set to 1`)
 		}
-	case strings.Contains(image, "ubuntu"):
+  }
```
gaohannk: indent

imagetest/test_suites/security/image_security_test.go
```diff
@@ -295,9 +295,7 @@ func verifyAutomaticUpdate(image string) error {
 		if interval > maxInterval || interval < minInterval {
 			return fmt.Errorf("autoclean interval is invalid or an unexpected length")
 		}
-	default:
-		return fmt.Errorf("unsupported image %s", image)
-	}
+  }
```
gaohannk: indent

imagetest/test_suites/security/image_security_test.go
zmarano: Fixed (keep forgetting to run gofmt)

imagetest/test_suites/security/image_security_test.go
hopkiw: there's no reason to give style feedback in golang; gofmt will apply a consistent style.

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -174,7 +175,11 @@ func TestArePackagesLegal(t *testing.T) {
 				t.Fatalf(err.Error())
 			}
 			if !isLegal {
-				t.Fatalf("Found illegal package: %v", filename)
+				if filename == "/usr/share/licenses/perl-Module-Build/LICENSE" {
```
gaohannk, line 178: move to line 173 before isPackageLegal or in isPackageLegal()

imagetest/test_suites/image_validation/license_test.go
```diff
@@ -116,9 +116,10 @@ var licenses = []string{
 	`Netcat and the associated package is a product of Avian Research, and is freely available in full source form with no restrictions save an obligation to give credit where due.`,
 	`Permission is hereby granted, without written agreement and without licence or royalty fees, to use, copy, modify, and distribute this software`,
 	`Open Market permits you to use, copy, modify, distribute, and license this Software and the Documentation for any purpose, provided that existing copyright notices are retained in all copies and that this notice is included verbatim in any distributions. No written agreement, license, or royalty fee is required for any of the authorized uses.`,
-	`This software is made available under the terms of *either* of the licenses found in LICENSE.APACHE or LICENSE.BSD. Contributions to cryptography are made under the terms of *both* these licenses.`,
+	`This software is made available under the terms of *either* of the licenses found in LICENSE.APACHE or LICENSE.BSD.`,
 	`This library.*is public domain software`,
-	`This Font Software is licensed under the SIL Open Font License`,
+	`Apache License`,
```
hopkiw, line 121: this does not belong here. we first try for license NAME, then we try license TEXT REGEX. this is the latter, but the string you've added belongs in the former.

imagetest/test_suites/image_validation/license_test.go
hopkiw, line 178: this is not appropriate. if you really want to second guess the entire thing and make this license OK, add 'FULLTEXT' to the license names variable.

imagetest/test_suites/image_validation/license_test.go
zmarano: The issue is with the regex... because its looking for a other context before the license name. For whatever reason, some of the apache licensed packages don't have that... the simply start with Apache License ... so the regex fails. I don't know the history of this regex however.

imagetest/test_suites/security/image_security_test.go
```diff
@@ -271,12 +271,12 @@ func verifyAutomaticUpdate(image string) error {
 	if err != nil {
 		return err
 	}
-	switch {
-	case strings.Contains(image, "debian-9"):
+	if strings.Contains(image, "debian-9") {
```
hopkiw, line 274: the switch statement was fine here, please restore. repeated exclusive cases are where switch statements shine.

imagetest/test_suites/image_validation/license_test.go
hopkiw: same for 'the artistic license 2.0'

imagetest/test_suites/image_validation/license_test.go
zmarano: That seems weird because "FULLTEXT" is not actually a license but sure that will work too (although as I stated before, the license names variable doesn't work).

imagetest/test_suites/image_validation/license_test.go
zmarano: Like I said, this doesn't work. Do you know why we have the wrapper text in the regex?
`licenseNameRegex = `(?i)(((License|Copyright)\s*:\s*%[1]s)|((covered )?under (the )?%[1]s)|(under (the terms of )?the %[1]s))``

The issue with several packages is that they don't conform to this.

hopkiw: my comments were supposed to be back to back :)
thanks for explaining, i think this means we should approach the design again from scratch, but fine to proceed like this for now due to the limitations.

PR #172, opened 2021-06-28
Skip secure boot test on Rocky Linux which doesn't yet support it.

zmarano:
Note that Rocky Linux will support secure boot soon - just not yet.

### DIFF
--- a/imagetest/test_suites/image_boot/image_boot_test.go
+++ b/imagetest/test_suites/image_boot/image_boot_test.go
@@ -56,8 +56,8 @@ func TestGuestSecureBoot(t *testing.T) {
 	}
 
 	if strings.Contains(image, "rocky-linux-8") {
-                t.Skip("secure boot is not yet supported on Rocky Linux")
-        }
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -55,6 +55,10 @@ func TestGuestSecureBoot(t *testing.T) {
 		t.Skip("secure boot is not supported on Debian 9")
 	}
 
+	if strings.Contains(image, "rocky-linux-8") {
```
hopkiw, line 58: i was gonna make a suggested change but github won't let me make suggestions on unchanged lines. i think this is now the time to make a case statement

zmarano: Its somewhat temporary- within a month (I suspect) we'll be back to one case. Still want to do that?

hopkiw: fine either way, then

PR #177, opened 2021-07-07
add host key test

gaohannk:
This is a migration from the BCT about host keys hostkey_publish_test.py;l=40.
The test verify that the ssh host public key that match the value in metadata.
http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/hostkeys
The test will be invoked by test manager in concourse and run on vm for each image as part of release.


imagetest/test_suites/hostkey/hostkey_pubish_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"flag"
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+var (
+	runtest = flag.Bool("runtest", false, "really run the test")
```
EricEdens, line 15: Where does this get used? Also, can you update the description to describe the test.

imagetest/test_suites/hostkey/hostkey_pubish_test.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, hostkey := range hostkeyLines {
+		keyType := strings.Split(hostkey, " ")[0]
+		keyValue := strings.Split(hostkey, " ")[1]
+		hostkeyMap[keyType] = keyValue
+	}
+
+	return hostkeyMap, nil
+}
+
+// TestMatchingKeysInGuestAttributes validate that host keys in guest attributes match those on disk.
```
EricEdens, line 35: Can you include which code is responsible for copying the keys? That way, if there's a failure, then the future reader knows where to look. For example: "validate that the guest agent copies the host keys from metadata to the disk"

imagetest/test_suites/hostkey/hostkey_pubish_test.go
hopkiw: this was from our legacy protection mechanism and should just be removed

imagetest/test_suites/hostkey/hostkey_pubish_test.go
gaohannk: are you saying we add comments for the link of the guest agent code that we test functionality?  I add a comment to before for loop.

EricEdens (approved):

PR #178, opened 2021-07-08
add ssh test

gaohannk:


imagetest/utils/test_utils.go
```diff
@@ -104,3 +105,14 @@ func ExtractBaseImageName(image string) (string, error) {
 	imageName := strings.Join(splits[:len(splits)-1], "-")
 	return imageName, nil
 }
+
+// GenerateSSHKeyPair generate ssh key pair.
+func GenerateSSHKeyPair(user string) error {
+	email := fmt.Sprintf("%s@test.com", user)
```
illfelder, line 111: The canonical fake domain is example.com - we should avoid using other domain names that may exist.

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	client, session, err := connectToHost(user, vmname, keyUrl)
+	if err != nil {
+		t.Fatalf("user %s failed ssh to target host, %s", user, vmname)
+	}
+	if err := session.Run("hostname"); err != nil {
+		t.Fatal("failed to run cmd hostname")
+	}
+	if err := client.Close(); err != nil {
+		t.Fatal("failed to close client")
```
hopkiw, line 53: this can be a warn, not fatal

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,98 @@
...
+		os.Exit(0)
+	}
+}
+
+func TestEmptyTest(t *testing.T) {
+	t.Logf("SSH target boot succesfully")
+	time.Sleep(600 * time.Second)
+}
+
+func TestSSHNonOsLogin(t *testing.T) {
```
hopkiw, line 35:
```suggestion
func TestSSH(t *testing.T) {
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if err := session.Run("hostname"); err != nil {
+		t.Fatal("failed to run cmd hostname")
+	}
+	if err := client.Close(); err != nil {
+		t.Fatal("failed to close client")
+	}
+}
+
+func connectToHost(user, host, keyPath string) (*ssh.Client, *ssh.Session, error) {
```
hopkiw, line 57:
```suggestion
func createSession(user, host, keyPath string) (*ssh.Client, *ssh.Session, error) {
```

imagetest/test_suites/ssh/setup.go
```diff
@@ -0,0 +1,42 @@
...
+	vm2, err := t.CreateTestVM("vm2")
+	if err != nil {
+		return err
+	}
+	// add public key in metadata
+	publicKey, err := ioutil.ReadFile("id_rsa.pub")
+	if err != nil {
+		return err
+	}
+	vm2.AddMetadata("ssh-keys", string(publicKey))
```
hopkiw, line 39: the format for the value of this metadata key is `USER:KEYDATA`

hopkiw:
```suggestion
	vm2.AddMetadata("ssh-keys", fmt.Sprintf("%s:%s", user, string(publicKey)))
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if *runtest {
+		os.Exit(m.Run())
+	} else {
+		os.Exit(0)
+	}
+}
+
+func TestEmptyTest(t *testing.T) {
+	t.Logf("SSH target boot succesfully")
+	time.Sleep(600 * time.Second)
```
hopkiw, line 32: i don't think this is needed - but you can test to be sure. either way this will make the test always take a minimum of 10 minutes

imagetest/test_suites/ssh/setup.go
```diff
@@ -0,0 +1,42 @@
...
+func TestSetup(t *imagetest.TestWorkflow) error {
+
+	vm, err := t.CreateTestVM("vm")
+	if err != nil {
+		return err
+	}
+
+	vm.AddMetadata("block-project-ssh-keys", "true")
+
+	if err := utils.GenerateSSHKeyPair(user); err != nil {
```
hopkiw, line 25: test author must call this, then read the pubkey file, then set up metadata... it would be better if we have one helper like vm.AddTestUser(username) that returns the data needed for the test author.

imagetest/test_suites/ssh/setup.go
```diff
@@ -0,0 +1,42 @@
...
+		return err
+	}
+	vm.RunTests("TestSSHNonOsLogin")
+
+	vm2, err := t.CreateTestVM("vm2")
+	if err != nil {
+		return err
+	}
+	// add public key in metadata
+	publicKey, err := ioutil.ReadFile("id_rsa.pub")
```
hopkiw, line 35: test author must know this filename?

gaohannk: return pub key when generate key pair, no need to know filename for test author.

imagetest/test_suites/ssh/setup.go
gaohannk: Move the function to fixture and make it as a method for vm.

gaohannk:
```

2021/07/14 00:01:51 imagetest: Done with setup
runTestWorkflow: running ssh on debian-10 (ID 6qpbl)
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="2" time="0">
	<testsuite name="ssh-debian-10" tests="2" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestSSH" time="0.44"></testcase>
	</testsuite>
</testsuites>
```

imagetest/fixtures.go
```diff
@@ -156,3 +159,30 @@ func (t *TestVM) EnableSecureBoot() {
 		}
 	}
 }
+
+// AddTestUser generate ssh key pair and return public key.
+func (t *TestVM) AddTestUser(user string) ([]byte, error) {
+	email := fmt.Sprintf("%s@example.com", user)
```
hopkiw, line 165: i dont really understand this. you're only using it in the optional 'comment' field of the key. i don't think you need this to be in email format?

imagetest/fixtures.go
```diff
@@ -156,3 +159,30 @@ func (t *TestVM) EnableSecureBoot() {
...
+	}
+	return publicKey, nil
+}
+
+func randNum() int {
+	min := 0
+	max := 100
+	delta := max - min
+
+	return rand.Intn(delta) + min
```
hopkiw, line 187: 2 issues. first is that this vanity function is essentially equivalent to `rand.Intn(100)` and so you should just use that where you are calling this. second is that random number out of 100 is pretty likely to result in collisions and you've got no collision detection. why not use uuid generator?

imagetest/fixtures.go
gaohannk: Yes, I did notice that this is a comment field. It seems that's not necessary to have.

imagetest/testworkflow.go
```diff
@@ -395,6 +397,12 @@ func finalizeWorkflows(tests []*TestWorkflow, zone, project, bucket string) erro
 	return nil
 }
 
+func (t *TestWorkflow) setKeyFileName() string {
+	keyFileName := "id_rsa_" + uuid.New().String()
```
hopkiw, line 401:
```suggestion
	keyFileName := "/id_rsa_" + uuid.New().String()
```

imagetest/fixtures.go
```diff
@@ -199,3 +201,21 @@ func (t *TestVM) AddAliasIPRanges(aliasIPRange, rangeName string) {
 		}
 	}
 }
+
+// AddTestUser generate ssh key pair and return public key.
+func (t *TestVM) AddTestUser() ([]byte, error) {
+	keyFileName := t.testWorkflow.setKeyFileName()
+	if keyFileName == "" {
+		return nil, fmt.Errorf("no key file found for user")
+	}
```
hopkiw, line 210: doesn't seem to be possible for this function to return ""
```suggestion
```

imagetest/fixtures.go
```diff
@@ -199,3 +201,21 @@ func (t *TestVM) AddAliasIPRanges(aliasIPRange, rangeName string) {
 		}
 	}
 }
+
+// AddTestUser generate ssh key pair and return public key.
+func (t *TestVM) AddTestUser() ([]byte, error) {
```
hopkiw, line 206: should be named like 'addSSHKey' or something, the user isn't involved here at all

hopkiw: also, does this function belong on TestVM? it doesn't do anything to the VM struct or the associated create-instances step

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestSSH(t *testing.T) {
+	vmname, err := utils.GetRealVMName("vm2")
+	if err != nil {
+		t.Fatal("failed to get real vm name")
+	}
+	keyURL, err := utils.GetMetadataAttribute("_ssh_key_url")
+	if err != nil {
+		t.Fatalf("Couldn't get key path from metadata")
+	}
+	client, session, err := createSession(user, fmt.Sprintf("%s:22", vmname), keyURL)
```
hopkiw, line 46:
```suggestion
	}
	pembytes, err := downloadPrivateKey(keyURL)
	if err != nil {
		t.Fatalf("failed to download private key: %v", err)
	}
	client, session, err := createSession(user, fmt.Sprintf("%s:22", vmname), keydata)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if err := client.Close(); err != nil {
+		t.Log("failed to close client")
+	}
+}
+
+func createSession(user, host, keyPath string) (*ssh.Client, *ssh.Session, error) {
+	pembytes, err := downloadPrivateKey(keyPath)
+	if err != nil {
+		panic(err)
+	}
```
hopkiw, line 62:
```suggestion
func createSession(user, host string, pembytes []byte) (*ssh.Client, *ssh.Session, error) {
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+	keyURL, err := utils.GetMetadataAttribute("_ssh_key_url")
+	if err != nil {
+		t.Fatalf("Couldn't get key path from metadata")
+	}
+	client, session, err := createSession(user, fmt.Sprintf("%s:22", vmname), keyURL)
+	if err != nil {
+		t.Fatalf("user %s failed ssh to target host, %s, err %v", user, vmname, err)
+	}
+	if err := session.Run("hostname"); err != nil {
+		t.Fatal("failed to run cmd hostname")
```
hopkiw, line 51:
```suggestion
		t.Fatalf("failed to run cmd hostname: %v", err)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if err := client.Close(); err != nil {
+		t.Log("failed to close client")
```
hopkiw, line 54:
```suggestion
		t.Logf("failed to close client: %v", err)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+	_, err := utils.GetMetadataAttribute("ssh-keys")
+	if err != nil {
+		t.Fatalf("couldn't get ssh public key from metadata")
+	}
+}
+
+func TestSSH(t *testing.T) {
+	vmname, err := utils.GetRealVMName("vm2")
+	if err != nil {
+		t.Fatal("failed to get real vm name")
```
hopkiw, line 40:
```suggestion
		t.Fatalf("failed to get real vm name: %v", err)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	keyURL, err := utils.GetMetadataAttribute("_ssh_key_url")
+	if err != nil {
+		t.Fatalf("Couldn't get key path from metadata")
```
hopkiw, line 44:
```suggestion
		t.Fatalf("Couldn't get key path from metadata: %v", err)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil, nil, err
+	}
+	return client, session, nil
+}
+
+func downloadPrivateKey(gcsPath string) ([]byte, error) {
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("failed to create cloud storage client: %v", err)
```
hopkiw, line 92:
```suggestion
		return nil, err
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -0,0 +1,99 @@
+package ssh
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
```
hopkiw, line 7:
```suggestion
```

hopkiw (changes requested):

imagetest/testworkflow.go
```diff
@@ -140,6 +141,7 @@ func (t *TestWorkflow) appendCreateVMStep(name, hostname string) (*daisy.Step, e
 	instance.Metadata = make(map[string]string)
 	instance.Metadata["_test_vmname"] = name
 	instance.Metadata["_test_package_url"] = "${SOURCESPATH}/testpackage"
+	instance.Metadata["_ssh_key_url"] = "${SOURCESPATH}/ssh-key"
```
hopkiw, line 144: this does suggest only one ssh key can exist in a single test workflow. might consider naming the key after the user in some way?

gaohannk: On way I can think to make per user ssh-key is to move this into setup where we generate keys. During setup we could decide how many keys needs.

imagetest/fixtures.go
gaohannk: move this to testworkflow as it's set key file source path on workflow

PR #187, opened 2021-07-16
CIT fixups

hopkiw:
* move all new network fixtures to fixtures.go
* maintain a convenience pointer in fixture structs
* use convenience pointer instead of name matching loops
* require fixture structs in SetCustomNetwork, reducing necessary safety checks
* add tests for new network fixtures


gaohannk:
/lgtm
One possible work is to continue break down fixture.go to multiple component such as network.go subnetwork.go vm.go

EricEdens (approved):

PR #190, opened 2021-07-26
add local user check

gaohannk:
Some extract check on ssh remote vm.

```
2021/07/26 19:01:44 imagetest: Done with setup
runTestWorkflow: running ssh on debian-9 (ID 8435m)
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="2" time="0">
	<testsuite name="ssh-debian-9" tests="2" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestSSH" time="0.29"></testcase>
	</testsuite>
</testsuites>
```

### DIFF
--- a/imagetest/test_suites/ssh/image_ssh_test.go
+++ b/imagetest/test_suites/ssh/image_ssh_test.go
@@ -40,23 +40,27 @@ func TestSSH(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to download private key: %v", err)
 	}
-	client, session, err := createSession(user, fmt.Sprintf("%s:22", vmname), pembytes)
+	client, err := createClient(user, fmt.Sprintf("%s:22", vmname), pembytes)
 	if err != nil {
 		t.Fatalf("user %s failed ssh to target host, %s, err %v", user, vmname, err)
 	}
-	if err := session.Run("hostname"); err != nil {
-		t.Fatalf("failed to run cmd hostname: %v", err)
+	if err := checkLocalUser(client, user); err != nil {
+		t.Fatalf("failed to check local user: %v", err)
+	}
+
+	if err := checkSudoGroup(client, user); err != nil {
+		t.Fatalf("failed to check sudo group: %v", err)
 	}
 	if err := client.Close(); err != nil {
 		t.Logf("failed to close client: %v", err)
 	}
 }
 
-func createSession(user, host string, pembytes []byte) (*ssh.Client, *ssh.Session, error) {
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -67,13 +71,37 @@ func createSession(user, host string, pembytes []byte) (*ssh.Client, *ssh.Sessio
...
+		return nil, err
 	}
+	return client, nil
+}
 
+// checkLocalUser test that the user account exists in /etc/passwd
+func checkLocalUser(client *ssh.Client, user string) error {
 	session, err := client.NewSession()
 	if err != nil {
 		client.Close()
```
bkatyl, line 83: If you close the client, all tests after it would likely fail as the client wouldn't usable. It'll also likely hit "failed to close client" error as it'll already be closed.

I'd suggest removing client.Close() from checkLocalUser and checkSudoGroup, so instead it only returns that this tests failed and returned an error.

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -67,13 +71,37 @@ func createSession(user, host string, pembytes []byte) (*ssh.Client, *ssh.Sessio
...
-		return nil, nil, err
+		return err
+	}
+	defer session.Close()
+	grepPasswdCmd := fmt.Sprintf("grep %s: /etc/passwd", user)
+	if err := session.Run(grepPasswdCmd); err != nil {
+		return err
+	}
+	return nil
```
bkatyl, line 91: This looks like it assumed success if there is no error. 

Is it possible to ensure the command succeeded in such a way that the output is also verifed. 
Ex: How would it know the difference between running and successfully returning with no results vs getting the expected result?

gaohannk: Check the manual of grep command:
```
Normally, the exit status is 0 if selected lines are found and 1 otherwise. But the exit status is 2 if an error occurred, unless the -q or --quiet or --silent option is used and a selected line is found. 
```
So, i thought it would be fine to just use error code 0 to indicate that found the line matched.

imagetest/test_suites/ssh/image_ssh_test.go
gaohannk: Remove this line.

imagetest/test_suites/ssh/image_ssh_test.go
gaohannk: https://linux.die.net/man/1/grep

bkatyl (approved):

PR #193, opened 2021-07-26
add ar-plugin test

gaohannk:
TODO: add proper authentication to bct-prod-images


imagetest/test_suites/artifact_registry/artifact_registry_test.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+const (
+	aptCreatePrivateRepo = `cat | sudo tee /etc/apt/sources.list.d/artifact-registry-private-repo.list <<EOF
+deb [ trusted=yes ] ar+https://us-central1-apt.pkg.dev/projects/bct-prod-images apt main
+EOF
+sudo DEBIAN_FRONTEND=noninteractive apt update
+`
+
+	yumCreatePrivateRepo = `cat | sudo tee /etc/yum.repos.d/artifact-registry-private-repo.repo <<EOF
```
hopkiw, line 20: as we discussed, there's no reason to encapsulate this logic with shell commands. we can just write the appropriate file  contents directly.

hopkiw:
/hold

this test will only succeed if the test artifacts (package repos with correct permissions) exist in the project. we should either make this test capable of running from any project, or add a CIT feature that a test is only valid in a certain project. that's design work for CIT, so let's hold this one for now.

PR #196, opened 2021-07-27
add metadata test

gaohannk:


hopkiw (changes requested):
one small change but looks fine otherwise

PR #200, opened 2021-07-29
build imagetest container on commit

hopkiw:
add a new git resource specific to the imagetest directory and which only produces new versions for changed go files in that directory. configure the image build job to trigger on all new changes.

### DIFF
--- a/concourse/pipelines/linux-image-build.yaml
+++ b/concourse/pipelines/linux-image-build.yaml
@@ -17,6 +17,12 @@ resources:
   source:
     uri: https://github.com/GoogleCloudPlatform/guest-test-infra.git
     branch: master
+- name: imagetest
+  type: git
+  source:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/linux-image-build.yaml
```diff
@@ -17,6 +17,12 @@ resources:
   source:
     uri: https://github.com/GoogleCloudPlatform/guest-test-infra.git
     branch: master
+- name: imagetest
+  type: git
+  source:
+    uri: https://github.com/GoogleCloudPlatform/guest-test-infra.git
+    branch: master
+    paths: ['imagetest/.*\.go']
```
gaohannk, line 25: So this is not a new repo, it's the subfolder and concourse will only download the subfolder of the repo.

PR #204, opened 2021-07-30
add test multinic

gaohannk:
Using ssh session to get private internal ip from remote vm metadata server
```
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="5" time="0">
	<testsuite name="network-debian-10" tests="5" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="" name="TestDefaultMTU" time="0"></testcase>
		<testcase classname="" name="TestAliasAfterReboot" time="0"></testcase>
		<testcase classname="" name="TestAliasAgentRestart" time="0.17"></testcase>
		<testcase classname="" name="TestVmToVm" time="8.59"></testcase>
	</testsuite>
</testsuites>
```


imagetest/fixtures.go
```diff
@@ -253,6 +266,16 @@ func (t *TestVM) AddAliasIPRanges(aliasIPRange, rangeName string) error {
 	return nil
 }
 
+// SetPrivateIP set IPv4 internal IP address to assign to the current test VMs.
+func (t *TestVM) SetPrivateIP(networkIP string) error {
+	if t.instance.NetworkInterfaces == nil {
+		return fmt.Errorf("Must call SetCustomNetwork prior to AddAliasIPRanges")
```
hopkiw, line 272: well this function is not AddAliasIPRanges

imagetest/fixtures.go
```diff
@@ -228,13 +235,19 @@ func (t *TestVM) SetCustomNetwork(network *Network, subnetwork *Subnetwork) erro
 	networkInterface := compute.NetworkInterface{
 		Network:    network.name,
 		Subnetwork: subnetworkName,
+		NetworkIP:  networkIP,
```
hopkiw, line 238: now networkIP is required? what happens if someone doesn't want to specify it? i think we can pull that bit out to its own function, or have some different signatures for this one

imagetest/fixtures_test.go
```diff
@@ -131,7 +132,7 @@ func TestAddAliasIPRanges(t *testing.T) {
 	if err != nil {
 		t.Errorf("failed to create network: %v", err)
 	}
-	if err := tvm.SetCustomNetwork(network, nil); err != nil {
+	if err := tvm.SetCustomNetwork(network, nil, ""); err != nil {
```
hopkiw, line 135: i think having to update all invocations of this function to add an empty method that only one new invocation will use is a strong indicator we've got the signature wrong

hopkiw (changes requested):

imagetest/fixtures.go
gaohannk: oops. typo.

imagetest/fixtures.go
gaohannk: Just like subnetwork. "" empty means not setting. but we could provide a SetCustomNetworkWithNetworkIP dedicate for this and make SetCustomNetwork call  SetCustomNetworkWithNetworkIP with empty string without need to change other setup.

imagetest/fixtures_test.go
```diff
@@ -236,6 +237,7 @@ func TestCreateNetworkDependencies(t *testing.T) {
 	}
 	var found bool
 	for _, dep := range deps {
+		fmt.Print(dep)
```
EricEdens, line 240: I'm assuming this was for debugging and not meant to be committed?

imagetest/test_suites/metadata/metadata_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build cit
 // +build cit
```
EricEdens, line 2: You removed `+build cit` in another file but left it here; also in another file, you only have  `+build cit` . What's the difference between `//go:build cit` and `// +build cit`>

imagetest/fixtures.go
```diff
@@ -220,6 +221,12 @@ func (t *TestVM) EnableSecureBoot() {
 // subnetwork. If subnetwork is empty, not using subnetwork, in this case
 // network has to be in auto mode VPC.
 func (t *TestVM) SetCustomNetwork(network *Network, subnetwork *Subnetwork) error {
+	return t.SetCustomNetworkWithNetworkIP(network, subnetwork, "")
+}
+
+// SetCustomNetworkWithNetworkIP set current test VMs in workflow using provided
+// network, subnetwork and a static IP selected by user.
+func (t *TestVM) SetCustomNetworkWithNetworkIP(network *Network, subnetwork *Subnetwork, networkIP string) error {
```
EricEdens, line 229: How about accepting `compute.NetworkInterface` instead of the individual fields? That allows you to avoid the somewhat-confusing (to me anyway :)  )  usage of empty strings and nils, such as `SetCustomNetworkWithNetworkIP(network1, nil, "")`

imagetest/fixtures.go
```diff
@@ -234,13 +241,19 @@ func (t *TestVM) SetCustomNetwork(network *Network, subnetwork *Subnetwork) erro
...
 				Type: "ONE_TO_ONE_NAT",
 			},
 		},
 	}
-	t.instance.NetworkInterfaces = []*compute.NetworkInterface{&networkInterface}
+	if t.instance.NetworkInterfaces == nil {
+		t.instance.NetworkInterfaces = []*compute.NetworkInterface{&networkInterface}
+	} else {
+		t.instance.NetworkInterfaces = append(t.instance.NetworkInterfaces, &networkInterface)
+	}
```
EricEdens, line 257: Based on the function's name and documentation, I would have expected any current network interfaces to be replaced. (The function talks about "set" rather than "append")

imagetest/fixtures.go
```diff
@@ -259,6 +272,16 @@ func (t *TestVM) AddAliasIPRanges(aliasIPRange, rangeName string) error {
 	return nil
 }
 
+// AddPrivateIP set IPv4 internal IP address to assign to the current test VMs.
+func (t *TestVM) AddPrivateIP(networkIP string) error {
+	if t.instance.NetworkInterfaces == nil {
+		return fmt.Errorf("Must call SetCustomNetwork prior to AddPrivateIP")
```
EricEdens, line 278: I'd suggest against this, unless you have a strong usecase: 
 1. It implies a before/after relationship of calling methods
 2. You update the first network interface, even though "SetCustomNetworkWithNetworkIP" is currently appending to the list of network interfaces.

imagetest/testworkflow.go
```diff
@@ -565,3 +565,20 @@ func (t *TestWorkflow) AddSSHKey(user string) (string, error) {
 
 	return string(publicKey), nil
 }
+
+// CreateFirewallRule create firewall rule.
+func (t *TestWorkflow) CreateFirewallRule(firewallName, networkName, protocal string, ports []string) error {
```
EricEdens, line 570: The naming and docs are unclear about what happens with "protocol" and "ports": Are they blocked? Allowed?

To make this clearer, I'd suggest changing the argument to a single "compute.Firewall" instance, so that it's clear what rule is being created.

imagetest/fixtures_test.go
gaohannk: oops. revert

imagetest/fixtures.go
gaohannk: We don't want to explore the whole GCE struct to end user to setup for test. However we could add a wrapper Network Interface struct as we did for Network and Subnetwork.

imagetest/fixtures.go
gaohannk: This method actually is not called in current revision. I make a refactor for this to inidcate which network interface as first parameter, then not only for 1st nic.

imagetest/fixtures.go
gaohannk: This wrapper method has be removed as we set network Ip latter instead set here. This method is just for create nic (keep original functional)

imagetest/fixtures.go
gaohannk: The function handle two cases. create 1st interface if not exist or append one more.

imagetest/test_suites/metadata/metadata_test.go
gaohannk: I don't quite understand the diff, seems introduced by go check. I remove this as no reason to make change.

imagetest/fixtures.go
EricEdens: The word `set` typically implies overwrite. So if you want the described behavior, you'd want to name the function `AddCustomNetwork` or `AppendCustomNetwork`

imagetest/fixtures.go
EricEdens: lgtm (I don't have permission to resolve the comment thread)

imagetest/fixtures.go
EricEdens: lgtm (I don't have permission to resolve the comment thread)

imagetest/test_suites/metadata/metadata_test.go
EricEdens: lgtm

imagetest/fixtures.go
gaohannk: Agree

imagetest/testworkflow.go
hopkiw: we intentionally do not take compute or daisy objects in this framework, that's not set in stone but if we want to change it i think it should be done conscientiously and not just for this test

gaohannk: We could consider provide a Firewall struct defined by us just like Network

EricEdens (approved):

PR #205, opened 2021-08-03
add sleep for ssh and log

gaohannk:


imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -27,6 +28,8 @@ func TestSSH(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to download private key: %v", err)
 	}
+	time.Sleep(time.Second)
```
hopkiw, line 31:
```suggestion
	time.Sleep(60 * time.Second)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -100,6 +103,8 @@ func TestHostKeysAreUnique(t *testing.T) {
...
```
hopkiw, line 106:
```suggestion
	time.Sleep(60 * time.Second)
```

imagetest/test_suites/ssh/image_ssh_test.go
```diff
@@ -27,6 +28,8 @@ func TestSSH(t *testing.T) {
...
+	t.Logf("connect to remote host at %d", time.Now().UnixNano())
```
hopkiw, line 32: probably we don't need nanosecond precision

hopkiw (approved):

PR #210, opened 2021-08-08
fail on test compilation failure

hopkiw:
test suites can fail to compile but not fail the docker build because this run command will always exit with the status from `cd ..`. this change simply makes a compilation failure into a Docker build failure.

we also need an improvement to presubmits in this repo for test suites, to catch compilation errors.


bkatyl (commented):
lgtm

PR #213, opened 2021-08-11
add shutdown script test

gaohannk:
1. Move TestGuestShutdownScript from image_boot test_suite to metdata test_suite
2. Add three new tests TestShutdownScript for normal case, TestRandomShutdownScriptNotCrashVM for bad case, TestShutdownUrlScript for script url
3. Add SetShutdownScriptURL method for setup.go.


imagetest/test_suites/metadata/setup.go
```diff
@@ -1,17 +1,87 @@
...
+  date +%s >> /shutdown.txt
+  sync
+  sleep 1
+done`
+	shutdownScriptTemplate = `#!/bin/bash
+echo "%s" > %s`
+	shutdownContent    = "The shutdown script worked."
+	shutdownOutputPath = "/shutdown_out.txt"
+	shutdownMaxLength  = 32768 // max shutdown metadata value
+	shutdownScriptURL  = "gs://gcp-guest-cloud-test-artifacts/shutdown-script.txt"
```
hopkiw, line 26: this makes the test only work for us. why do we need a test artifact to pre-exist? we can accomplish the same thing by writing the content to a file that will get uploaded by the daisy workflow.

imagetest/test_suites/metadata/setup.go
```diff
@@ -1,17 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+	vm.RunTests("TestTokenFetch|TestMetaDataResponseHeaders|TestGetMetaDataUsingIP")
+
+	vm2, err := t.CreateTestVM("vm2")
+	if err != nil {
+		return err
+	}
+	vm2.SetShutdownScript(fmt.Sprintf(shutdownScriptTemplate, shutdownContent, shutdownOutputPath))
```
hopkiw, line 41: is there some reason we need to store a template as a constant and then modify it before use, here? it's only used in this one place. the python test had templates for code re-use between startup and shutdown scripts.

imagetest/test_suites/metadata/shutdown_script_test.go
```diff
@@ -0,0 +1,60 @@
...
+	output := strings.TrimSpace(string(bytes))
+	if output != shutdownContent {
+		t.Fatalf(`shutdown script output expect "%s", but actually "%s"`, shutdownContent, output)
+	}
+}
+
+// TestRandomShutdownScriptNotCrashVM test that a script with random content
+// doesn't crash the vm.
+func TestRandomShutdownScriptNotCrashVM(t *testing.T) {
+	// second boot
```
hopkiw, line 43: confusing comment. this code will be run on both boots

imagetest/test_suites/metadata/shutdown_script_test.go
```diff
@@ -0,0 +1,60 @@
...
+	if _, err := utils.GetMetadataAttribute("shutdown-script"); err != nil {
```
hopkiw, line 44: so in this test we are testing whether setting the metadata at instance creation will result in successful get of that same metadata here in the test. i don't know how it relates to the title of this test. what would a failing test look like?

imagetest/test_suites/metadata/setup.go
```diff
@@ -1,17 +1,87 @@
...
 	}
+	vm5.SetShutdownScript(shutdownScript)
+	if err := vm5.Reboot(); err != nil {
+		return err
+	}
+	vm5.RunTests("TestGuestShutdownScript")
 	return nil
 }
+
+func randStringRunes(n int) string {
```
hopkiw, line 79: is there some reason it specifically has to be *random* data? are we hoping the test will succeed sometimes and fail other times, depending on the contents?

imagetest/test_suites/metadata/shutdown_script_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+const shutdownTime = 110 // about 2 minutes
+
+// TestGuestShutdownScript test that shutdown scripts can run for around two minutes
+func TestGuestShutdownScript(t *testing.T) {
```
hopkiw, line 15: rename to TestShutdownScriptTime, since that's what's different from TestShutdownScript
```suggestion
func TestShutdownScriptTime(t *testing.T) {
```

hopkiw (changes requested):

gaohannk:
```
2021/08/11 19:15:33 Running in project gcp-guest zone us-west1-c
2021/08/11 19:15:33 using -filter metadata
2021/08/11 19:15:33 Add test workflow for test metadata on image projects/debian-cloud/global/images/family/debian-10
2021/08/11 19:15:33 imagetest: Done with setup
2021/08/11 19:15:33 Storing artifacts and logs in gs://gcp-guest-cloud-test-outputs/2021-08-11T19:15:33Z
2021/08/11 19:15:33 runTestWorkflow: running metadata on debian-10 (ID xlzr6)
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="7" time="0">
	<testsuite name="metadata-debian-10" tests="7" failures="0" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="metadata-debian-10" name="TestTokenFetch" time="0"></testcase>
		<testcase classname="metadata-debian-10" name="TestMetaDataResponseHeaders" time="0"></testcase>
		<testcase classname="metadata-debian-10" name="TestGetMetaDataUsingIP" time="0"></testcase>
		<testcase classname="metadata-debian-10" name="TestShutdownScript" time="0"></testcase>
		<testcase classname="metadata-debian-10" name="TestRandomShutdownScriptNotCrashVM" time="0"></testcase>
		<testcase classname="metadata-debian-10" name="TestShutdownUrlScript" time="0"></testcase>
		<testcase classname="metadata-debian-10" name="TestGuestShutdownScript" time="0.02"></testcase>
	</testsuite>
</testsuites>

```

imagetest/test_suites/metadata/setup.go
gaohannk: Good catch, wee could config this in setup.

imagetest/test_suites/metadata/shutdown_script_test.go
gaohannk: In my understanding, the random script is more like a bad script, i.e. a bad script failed and exit doesn't crash vm in terms of make disk bad that can't reboot.

gaohannk: curl metadata is just a way to make sure the instance is normal as it's first boot without running shutdown script

imagetest/test_suites/metadata/setup.go
gaohannk: Right, it's not pretty useful here to use template. Remove

imagetest/test_suites/metadata/shutdown_script_test.go
hopkiw: how can it 'crash' the vm? so this test could do anything, it could just log 'success' and it would be a good test?

imagetest/test_suites/metadata/shutdown_script_test.go
hopkiw: the comment is just incorrect. this is not always the second boot. the comment also doesn't convey any meaningful information

imagetest/test_suites/metadata/setup.go
gaohannk: I don't expect the random data will succeed. The test is more like to test failure case.

imagetest/test_suites/metadata/shutdown_script_test.go
gaohannk: So I rename test to TestShutdownScriptFailedNotCrashVM

gaohannk:
Update 1. refactor to use daisy upload source file for shutdown script and the URL is daisy source path
2 Rename random test to TestShutdownScriptFailedNotCrashVM

imagetest/test_suites/metadata/shutdown_script_test.go
gaohannk: I remove the comment

imagetest/fixtures.go
```diff
@@ -28,6 +29,7 @@ const (
 	createNetworkStepName    = "create-networks"
 	createSubnetworkStepName = "create-sub-networks"
 	successMatch             = "FINISHED-TEST"
+	localShutdownScriptPath  = "/shutdown_script.txt"
```
hopkiw, line 32: cant use a constant here, it will get clobbered. needs to be a uuid or the 3-tuple of testname, vmname, testimage

PR #215, opened 2021-08-13
add startup script test

gaohannk:
This is similar test like shutdown script. we only test startup script but not startup-script-url as it's already tested by wrapper.


imagetest/test_suites/metadata/setup.go
```diff
@@ -1,17 +1,52 @@
...
 
 // Name is the name of the test package. It must match the directory name.
 var Name = "metadata"
 
+const (
+	startupScriptTemplate = `#!/bin/bash
+echo "%s" > %s`
+	startupOutputPath = "/startup_out.txt"
+	startupContent    = "The startup script worked."
+	startupMaxLength  = 32768 // max shutdown metadata value
```
EricEdens, line 18: Can you add a link to where this number is defined? Is it a constant imposed by GCE?

imagetest/test_suites/metadata/startup_script_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+// TestStartupScript test the standard metadata script.
+func TestStartupScript(t *testing.T) {
+	bytes, err := ioutil.ReadFile(startupOutputPath)
+	if err != nil {
+		t.Fatalf("failed to read shutdown output %v", err)
```
EricEdens, line 17: do you mean startup output?

imagetest/test_suites/metadata/setup.go
```diff
@@ -1,17 +1,52 @@
...
+	if err := vm2.Reboot(); err != nil {
+		return err
+	}
+	vm2.RunTests("TestStartupScript$")
+
+	vm3, err := t.CreateTestVM("vm3")
+	if err != nil {
+		return err
+	}
+	vm3.SetShutdownScript(strings.Repeat("a", startupMaxLength))
```
EricEdens, line 45: I see that the test references "StartupScript", but this sets the shutdown script.

imagetest/test_suites/metadata/startup_script_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+// TestStartupScript test the standard metadata script.
```
EricEdens, line 13: Can you add some more context to this?

imagetest/test_suites/metadata/startup_script_test.go
```diff
@@ -0,0 +1,30 @@
...
+func TestStartupScript(t *testing.T) {
+	bytes, err := ioutil.ReadFile(startupOutputPath)
+	if err != nil {
+		t.Fatalf("failed to read shutdown output %v", err)
+	}
+	output := strings.TrimSpace(string(bytes))
+	if output != startupContent {
+		t.Fatalf(`shutdown script output expect "%s", but actually "%s"`, startupOutputPath, output)
```
EricEdens, line 21: shutdown again

EricEdens (commented):
Overall looks good! A few comments about adding comments for future readers, plus some notes regarding mixed usage of "startup" vs "shutdown".

imagetest/test_suites/metadata/setup.go
gaohannk: Yes. https://cloud.google.com/compute/docs/metadata/setting-custom-metadata#limitations

imagetest/test_suites/metadata/startup_script_test.go
gaohannk: Typo.

imagetest/test_suites/metadata/startup_script_test.go
gaohannk: Typo.

imagetest/test_suites/metadata/startup_script_test.go
gaohannk: TestStartupScript verify that the standard metadata script could run successfully by checking the output content

imagetest/test_suites/metadata/setup.go
gaohannk: Typo.

imagetest/test_suites/metadata/setup.go
```diff
@@ -22,10 +26,12 @@ done`
 echo "%s" > %s`
 	shutdownOutputPath = "/shutdown_out.txt"
 	shutdownContent    = "The shutdown script worked."
-	shutdownMaxLength  = 32768 // max shutdown metadata value
+	// max metadata value https://cloud.google.com/compute/docs/metadata/setting-custom-metadata#limitations
+	metadataMaxLength = 32768
```
EricEdens, line 30: Thanks for the link; I'm seeing "Each metadata value has a maximum limit of 256 KB" on that page -- is 32768 something else? Also, in the comment, can you specify the unit? something like: Maximum size of metadata value payload, in bytes.

gaohannk: Interesting, i didn't pay much attention on this value. I found this link, it seems these is inconsistent which I think one is legacy value.
https://cloud.google.com/sdk/gcloud/reference/container/clusters/create

gaohannk: ```Values must be less than or equal to 32,768 bytes in length. The total size of all keys and values must be less than 512 KB.```

EricEdens (approved):

PR #216, opened 2021-08-16
Don't block on tests until tests are stable.

zmarano:


hopkiw:
no, this definitely will not work. now you have a parallelization in the pipeline. let's discuss offline.

PR #219, opened 2021-08-18
Image test daemon scirpt

gaohannk:


imagetest/test_suites/metadata/startup_script_test.go
```diff
@@ -29,3 +30,16 @@ func TestStartupScriptFailed(t *testing.T) {
 		t.Fatalf("couldn't get startup-script from metadata, %v", err)
 	}
 }
+
+// TestDaemonScript test a backgrounded daemon is still running in the VM.
```
EricEdens, line 34: Can you clarify in the comment: Is this testing that the startup script itself was written to daemonize? Or that the guest agent started the process in a particular way?

PR #225, opened 2021-09-02
Add cli tool for publishing a cloud monitoring time series data point

Linskeyd:
This PR contains a new CLI tool for publishing a single time series data point to cloud monitoring. Made it flexible enough to use for other Concourse use-cases (i.e. metric name is configurable, could use the monitoring pkg directly from other go code in other tasks).

This isn't used in any tasks or pipelines yet. Planning to submit separate PRs for a Dockerfile with this tool baked in, updating cloud build configuration to build the new image on submit, and utilize the new cli tool in a Concourse task.

Unit tests:
```
$ go test
PASS
ok      github.com/GoogleCloudPlatform/guest-test-infra/container_images/concourse-metrics/pkg/monitoring       0.052s
```

Manual test execution (links to dashboards of emitted data available via chat):
```
$ go run container_images/concourse-metrics/cmd/publish-job-result/main.go --project-id="linskeyd-gce" --zone="us-west1-a" --pipeline="linskeyd-gce" --job="test-job-three" --task="publish-job-metrics" --result-state="success" --start-timestamp="1630534391108" --end-timestamp="1630534408108" --metric-path="testing/duration-generic"
```


container_images/concourse-metrics/pkg/monitoring/requests_test.go
```diff
@@ -0,0 +1,96 @@
...
+		{JobResultArgs{EndTimestamp: &validEndTimestamp, MetricPath: metricPath, StartTimestamp: validEndTimestamp + 1000, Job: job, Pipeline: pipeline, ProjectId: projectId, ResultState: resultState, Task: task, Zone: zone}, "end-timestamp cannot occur before start-timestamp"},
+	}
+	for _, test := range tests {
+		if _, err := BuildJobResultRequest(test.args); err == nil || !strings.Contains(err.Error(), test.message) {
+			t.Errorf("BuildJobResultRequest(%+v) = (_, %v); want error to contain %s", test.args, err, test.message)
+		}
+	}
+}
+
+func assertEqual(t *testing.T, x interface{}, y interface{}) {
```
zoran15, line 90: There's `github.com/stretchr/testify/assert` for asserts.

Linskeyd: Ah I didn't know this existed; thanks for the pointer. I'll swap assert usage out for testify in a separate PR.

container_images/concourse-metrics/pkg/monitoring/requests.go
```diff
@@ -0,0 +1,125 @@
...
+		endTimestamp = *input.EndTimestamp
+	}
+
+	e := validateJobResultRequestInput(&input)
+	if e != nil {
+		return nil, e
+	}
+
+	// Calculate the duration to publish.
+	duration := endTimestamp - input.StartTimestamp
```
hopkiw, line 90: is this assuming both are in ms? and do we need to store this in a variable if it's only used once?

container_images/concourse-metrics/pkg/monitoring/requests.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/golang/protobuf/ptypes/timestamp"
+	metricpb "google.golang.org/genproto/googleapis/api/metric"
+	"google.golang.org/genproto/googleapis/api/monitoredres"
+	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+)
+
+// JobResultArgs defines the required and optional arguments for building a new
+// job result request (in the form of a cloud monitoring time series request).
+type JobResultArgs struct {
```
hopkiw, line 30: comment suggests some of these are optional, how should we know which ones?

i would also say it's somewhat uncommon to use struct-of-parameters in method signatures, though probably fine in this usage

container_images/concourse-metrics/pkg/monitoring/requests.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// Calculate the duration to publish.
+	duration := endTimestamp - input.StartTimestamp
+
+	return &monitoringpb.CreateTimeSeriesRequest{
+		Name: "projects/" + input.ProjectID,
+		TimeSeries: []*monitoringpb.TimeSeries{{
+			Metric: &metricpb.Metric{
+				Type: "custom.googleapis.com/" + input.MetricPath,
```
hopkiw, line 96: any rules or restrictions on what can be used here? docs say 'an existing metric type', must the metricpath already exist, or will it be created on-demand?

container_images/concourse-metrics/cmd/publish-job-result/main.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		fmt.Printf("Error creating a new Cloud Monitoring metric client: %+v.\n", err)
+		os.Exit(1)
+	}
+	defer c.Close()
+
+	req, err := requests.BuildJobResultRequest(requests.JobResultArgs{
+		EndTimestamp:   endTimestamp,
+		Job:            *job,
+		MetricPath:     *metricPath,
```
hopkiw, line 55: this is the actual name of the metric - shouldn't it simply be a derived value from inputs of pipeline/job/task ? and it's not clear to me what you intend to provide for this value when calling it, could you give an example?

container_images/concourse-metrics/pkg/monitoring/requests.go
```diff
@@ -0,0 +1,125 @@
...
+	// Provide a default for the endTimestamp if one was not provided.
+	var endTimestamp int64
+	if input.EndTimestamp == nil {
+		endTimestamp = time.Now().UnixNano() / 1000000
+	} else {
+		endTimestamp = *input.EndTimestamp
+	}
+
+	e := validateJobResultRequestInput(&input)
+	if e != nil {
```
hopkiw, line 85: what happens if we don't validate in advance? are we just trying to save the round trip to server?

also suggested change for go, since we don't need this e variable:

```suggestion
	if e := validateJobResultRequestInput(&input); e != nil {
```

container_images/concourse-metrics/cmd/publish-job-result/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"time"
+
+	monitoring "cloud.google.com/go/monitoring/apiv3"
+	requests "github.com/GoogleCloudPlatform/guest-test-infra/container_images/concourse-metrics/pkg/monitoring"
+)
+
+var (
+	endTimestamp   = flag.Int64("end-timestamp", time.Now().UnixNano()/1000000, "End timestamp of the job run. Defaults to now.")
+	job            = flag.String("job", "", "Concourse job name.")
```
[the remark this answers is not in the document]
hopkiw: alternatively if this is really going to be single-purpose, we might make it more concourse-integrated by reading values from files, saving us from adding load_var steps to the job. can always make this a future improvement, too

container_images/concourse-metrics/cmd/publish-job-result/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	monitoring "cloud.google.com/go/monitoring/apiv3"
+	requests "github.com/GoogleCloudPlatform/guest-test-infra/container_images/concourse-metrics/pkg/monitoring"
```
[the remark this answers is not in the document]
Linskeyd: Yeah at first I didn't realize it was a convention in go to use the last path segment as the name. 'monitoring' is the recommended name in the examples for the go/monitoring/apiv3 import. Will update 'monitoring' to 'monitoringapi'.

container_images/concourse-metrics/cmd/publish-job-result/main.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+var (
+	endTimestamp   = flag.Int64("end-timestamp", time.Now().UnixNano()/1000000, "End timestamp of the job run. Defaults to now.")
```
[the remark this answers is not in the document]
Linskeyd: Primarily for testing, but another use-case here could be for manually specifying an end time to avoid manual approval steps.

container_images/concourse-metrics/cmd/publish-job-result/main.go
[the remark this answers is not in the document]
Linskeyd: Yeah I kept it somewhat still geared towards Concourse use-cases. We can re-evaluate this later if we find that the load_var step is too time-consuming / cumbersome.

container_images/concourse-metrics/cmd/publish-job-result/main.go
```diff
@@ -0,0 +1,73 @@
...
+	job            = flag.String("job", "", "Concourse job name.")
+	metricPath     = flag.String("metric-path", "", "Path of the custom metric name to use (custom.googleapis.com/[metric-path]).")
+	projectID      = flag.String("project-id", "", "GCP project Id.")
+	pipeline       = flag.String("pipeline", "", "Concourse pipeline name.")
+	task           = flag.String("task", "", "Concourse task name publishing this metric.")
+	resultState    = flag.String("result-state", "", "Concourse job result state ('success' or 'failure')")
+	startTimestamp = flag.Int64("start-timestamp", -1, "Start timestamp of the job run.")
```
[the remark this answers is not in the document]
Linskeyd: Good point that this should totally call out the units of the flag, I'll add that. Is there a better way to default the input value to something that will fail validation for an int? Otherwise I can add a comment to show why -1 is set (to purposely fail validation when not provided).

container_images/concourse-metrics/cmd/publish-job-result/main.go
Linskeyd: The full metric name is the custom.googleapis.com/metric/path bit. For the path, I've made it configurable input so that if we want to use it for other areas (such as duration within a task, of a particular action, then we can do so).

Example within a task executable: pulling in this go pkg code to measure time series duration points around doing a build or some other work.

Example as a Concourse task definition: we'd just specify the input as --metric-path=((pipeline))/((job))/duration

container_images/concourse-metrics/pkg/monitoring/requests.go
Linskeyd: I followed the convention that Google's go client seems to be following. Optional params having the ability to be 'nil' by being a pointer rather than an explicit input value. Also, go readability suggests that we use a struct like this when there starts to be a lot of params: go/go-option-structure

container_images/concourse-metrics/pkg/monitoring/requests.go
Linskeyd: For most of the parameters, it is just saving the round trip. For result-state specifically, it is doing a correctness check on what should be allowed in our request (which would otherwise just be blindly accepted by the cloud monitoring backend).

For the suggested style change, I love it. Will update.

container_images/concourse-metrics/pkg/monitoring/requests.go
Linskeyd: Yes the assumption is ms and we don't have to store. I've stored explicitly to give the value a name though, but could also just add a comment in the struct construction that calls out this is intended to be a duration.

container_images/concourse-metrics/pkg/monitoring/requests.go
Linskeyd: If it doesn't exist, it'll be created on-demand. Check out https://cloud.google.com/monitoring/custom-metrics/creating-metrics#auto-creation

container_images/concourse-metrics/pkg/monitoring/requests.go
Linskeyd: Also, I wish there was better support for explicitly marking optional/required args. If you know of a better way, please let me know.

container_images/concourse-metrics/cmd/publish-job-result/main.go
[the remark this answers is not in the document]
hopkiw: or you can change the actual folder name and package name to 'requests' so they match. either way

container_images/concourse-metrics/cmd/publish-job-result/main.go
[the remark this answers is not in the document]
hopkiw: i don't know of a way, no. makes sense now that i get why you set it

container_images/concourse-metrics/pkg/monitoring/requests.go
hopkiw: so the go clients are autogenerated and therefore not good examples of idiomatic go, sadly. optional params will only be nil if they are pointers, something you can also do with explicit function parameters. 

generally since go doesn't have polymorphism we just ape polymorphism with multiple signatures, return a modifiable struct, or if it's unwieldy we use a struct as you're doing now. it's just not very common to need a lot of things in one method like this :)

hopkiw: oh and the bit about using a struct was really an aside. the feedback here is to document the optional params

container_images/concourse-metrics/pkg/requests/requests_test.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+}
+
+func assertEqual(t *testing.T, x interface{}, y interface{}) {
+	if x == y {
+		return
+	}
+
+	t.Errorf("Received %v (type %v), expected %v (type %v)", x, reflect.TypeOf(x), y, reflect.TypeOf(y))
```
hopkiw, line 129: just for future ref, i think you can tell you've gone down a wrong path when you have to do things like this. you've really written this test in a style from some other language, idk any golang testing that looks like this.

Linskeyd: I think this may be more common in go than you think (a popular framework makes use of it - the one suggested by Zoran): https://github.com/stretchr/testify/blob/master/assert/assertions.go

hopkiw: yeah zoran and i disagree here haha, i never use stretchr/testify

PR #232, opened 2021-09-13
build linux images daily

hopkiw:
* add daily-time resource to linux-image-build pipeline
* add triggering get on daily-time resource to all build jobs in linux-image-build pipeline


LawrenceHwang (approved):
LGTM. Thank you.

PR #233, opened 2021-09-13
fix package generation task

hopkiw:
existing logic is incorrectly producing e.g. `20210913.1` instead of `20210913.01`. fix with printf instead of echo

### DIFF
--- a/concourse/tasks/generate-package-version.yaml
+++ b/concourse/tasks/generate-package-version.yaml
@@ -17,16 +17,12 @@ run:
   args:
   - -exc
   - |
-    cd repo
-    latest=$(git tag -l '20*'|tail -1)  # TODO: use shell logic only
+    latest=$(cd repo;git tag -l '20*'|tail -1)  # TODO: use shell logic only
     latest_date=${latest/.*}
     todays_date=$(date '+%Y%m%d')
+    latest_build=0
     if [[ $latest_date == $todays_date ]]; then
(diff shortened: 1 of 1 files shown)


concourse/tasks/generate-package-version.yaml
```diff
@@ -17,16 +17,12 @@ run:
   args:
   - -exc
   - |
-    cd repo
-    latest=$(git tag -l '20*'|tail -1)  # TODO: use shell logic only
+    latest=$(cd repo;git tag -l '20*'|tail -1)  # TODO: use shell logic only
     latest_date=${latest/.*}
     todays_date=$(date '+%Y%m%d')
+    latest_build=0
```
LawrenceHwang, line 23: Seems to me the previous default was `00` (line 29). Just curious if using `0`  now is intended? Thanks.

hopkiw: on line 28 we take $latest_build and apply it with 'printf %02d', which will 0-pad it to a minimum of two digits as needed. prior to this PR we used a string in the default case (.00 on line 29 as you indicate), but a bash integer variable in other cases, which were printed as .1 through .9

LawrenceHwang: Got it! Thanks!

LawrenceHwang (approved):
LGTM. Thank you.

PR #238, opened 2021-09-28
Instrument the dev-image-build pipeline with success/failure/duration metrics

Linskeyd:
```
fly validate-pipeline --config concourse/pipelines/dev-image-build.yaml 
looks good
```

Tested that the same success/failure handlers work in my dev pipeline in Concourse.

The config lines are a bit repetitive; looked to see if there is a way to reduce clutter by referencing vars available in the pipeline overall (rather than directly re-specifying the same ones each time). Turned up "Parameterizing a task file with vars" [here](https://concourse-ci.org/jobs.html#schema.step.task-step.vars) which will check in with a credential manager to try and resolve implicit vars. The only thing I found to make this less explicit was a 'dummy' var source (which lets you specify static values at the top of a pipeline). That would at least let us reduce the duplicated entries which are the same between different jobs.

### DIFF
--- a/concourse/pipelines/dev-image-build.yaml
+++ b/concourse/pipelines/dev-image-build.yaml
@@ -72,6 +72,10 @@ jobs:
   plan:
   - get: compute-image-tools
   - get: guest-test-infra
+  - task: generate-timestamp
+    file: guest-test-infra/concourse/tasks/generate-timestamp.yaml
+  - load_var: start-timestamp-ms
+    file: timestamp/timestamp-ms
   - task: generate-build-id
     file: guest-test-infra/concourse/tasks/generate-build-id.yaml
     vars:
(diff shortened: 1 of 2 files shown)


concourse/pipelines/dev-image-build.yaml
```diff
@@ -96,10 +100,34 @@ jobs:
...
+      result_state: "success"
+      start_timestamp: ((.:start-timestamp-ms))
+      metric_path: "concourse/job/duration"
+  on_failure:
+    task: failure
+    file: guest-test-infra/concourse/tasks/publish-job-result.yaml
+    vars:
+      pipeline: "dev-image-build"
+      job: "build-debian-10-worker"
+      task: "publish-job-result"
```
hopkiw, line 119: this is not a variable in the task, i think it will give an error in use

concourse/pipelines/dev-image-build.yaml
```diff
@@ -96,10 +100,34 @@ jobs:
...
+      result_state: "failure"
+      start_timestamp: ((.:start-timestamp-ms))
+      metric_path: "concourse/job/duration"
```
hopkiw, line 122: if metric-path would always be concourse/job/duration, we might just bake it into the task

hopkiw (approved):

concourse/pipelines/dev-image-build.yaml
Linskeyd: This didn't fail in dev pipeline testing, but the arg is no longer needed since it is directly specified anyways (moved into the task as per your suggestion).

concourse/pipelines/dev-image-build.yaml
Linskeyd: Agreed, moved this to the task definition.

PR #239, opened 2021-10-06
add image testing

hopkiw:
add image tests at the end of publish-to-testing jobs. failure of tests will fail the publish job, so also add retries.

### DIFF
--- a/concourse/pipelines/linux-image-build.yaml
+++ b/concourse/pipelines/linux-image-build.yaml
@@ -708,6 +708,12 @@ jobs:
       publish_version: ((.:publish-version))
       wf: "enterprise_linux/almalinux_8.publish.json"
       environment: "test"
+  - task: image-test-almalinux-8
+    file: guest-test-infra/concourse/tasks/image-test-args.yaml
+    attempts: 3
+    vars:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/linux-image-build.yaml
```diff
@@ -1062,6 +1122,11 @@ jobs:
       publish_version: ((.:publish-version))
       wf: "debian/debian_9.publish.json"
       environment: "test"
+  - task: image-test-debian-9
+    file: guest-test-infra/concourse/tasks/image-test-args.yaml
+    vars:
+      images: projects/bct-prod-images/global/images/debian-9-v((.:publish-version))
```
zmarano, line 1128: The image names for Debian include the release, so debian-9-stretch-vYYYYMMDD

zmarano (changes requested):

zmarano (approved):

LawrenceHwang (commented):
/LGTM

Additional thoughts for future:
To reveal only the minimum necessary configs, it can be beneficial to save these extra-args and the image projects in a different store (maybe concourse var sources or secrets manager..etc).

PR #292, opened 2021-11-02
Add OS Login tests for getent passwd.

wrigri:


imagetest/test_suites/oslogin/oslogin_test.go
```diff
@@ -96,3 +103,69 @@ func TestOsLoginDisabled(t *testing.T) {
...
+		t.Errorf("getent passwd output does not contain %s", TEST_USER_ENTRY)
+	}
+	cmd = exec.Command("/usr/bin/getent", "passwd", TEST_USERNAME)
+	out, err = cmd.Output()
+	if err != nil {
+		t.Fatalf("getent command failed %v", err)
+	}
+	if !strings.Contains(string(out), TEST_USER_ENTRY) {
+		t.Errorf("getent passwd output does not contain %s", TEST_USER_ENTRY)
+	}
```
zoran15, line 151: How is this different to `TestGetentPasswdOsloginUser` and why is asserting against `TEST_USER_ENTRY` when `TEST_USERNAME` is provided?

hopkiw: the output of the command should be what's in `TEST_USER_ENTRY`, exactly. compared to the less specific `TEST_USERNAME` which will only be part of the output

imagetest/test_suites/oslogin/oslogin_test.go
```diff
@@ -96,3 +103,69 @@ func TestOsLoginDisabled(t *testing.T) {
...
+	if !strings.Contains(string(out), "nobody:x:") {
+		t.Errorf("getent passwd output does not contain user nobody")
+	}
+}
+
+func TestGetentPasswdInvalidUser(t *testing.T) {
+	cmd := exec.Command("/usr/bin/getent", "passwd", "__invalid_user__")
+	err := cmd.Run()
+	if err.Error() != "exit status 2" {
+		t.Errorf("getent passwd did not give error on invaid user")
```
hopkiw, line 169:
```suggestion
		t.Errorf("getent passwd did not give error on invalid user")
```

imagetest/test_suites/oslogin/setup.go
```diff
@@ -14,7 +14,7 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		return err
 	}
 	vm.AddMetadata("enable-oslogin", "true")
-	vm.RunTests("TestOsLoginEnabled")
+	vm.RunTests("TestOsLoginEnabled|TestGetentPasswdAllUsers|TestGetPasswdOsloginUser|TestGetPasswdOsloginUID|TestGetentPasswdLocalUser|TestGetentPasswdInvalidUser")
```
hopkiw, line 17: totally optional feedback but these are prefixes, so you could just say `TestGetent` here if you don't need to split them across hosts.

hopkiw (changes requested):
overall looks great, mostly minor feedback. also as zoran mentioned i think you accidentally duplicated the username test

imagetest/test_suites/oslogin/oslogin_test.go
wrigri: So, in discussions with Liam before the PR, it was noted that we should test both that asking for username returns the correct UID and asking for UID returns the correct username. TEST_USER_ENTRY contains both the username and the UID. However, when I added the test to make sure that username returns UID, I didn't realize that TestGetentPassweOsloginUser already tests this. I'll remove this one and just let the previous test cover it.

zoran15 (approved):

PR #298, opened 2021-11-06
fix network test

hopkiw:
* merge tests to 2 VMs
* test combination of alias routes + multinic together
* refactor network/setup.go for readability


hopkiw:
this is tested working on latest image and tested failing on the image with the multinic bug.

zoran15 (approved):

PR #318, opened 2021-12-04
Add SQL server builds/ releases

LawrenceHwang:
Add the rest of SQL server builds and change release to test flag to true in the base windows image.

/cc bkatyl


bkatyl (approved):
For PR's that are large or greater, please only do a single change. The changes to the Windows server release parts of the pipeline would have been better as its own PR.

PR #320, opened 2021-12-09
Update triggered resources

LawrenceHwang:
- Update triggered resources so the jobs are linked to resources more meaningful to the jobs
- Add skip_download: "true" to publish steps to avoid unnecessary download

/cc @bkatyl
/cc @hopkiw

/hold

### DIFF
--- a/concourse/pipelines/windows-image-build.yaml
+++ b/concourse/pipelines/windows-image-build.yaml
@@ -1242,9 +1242,12 @@ jobs:
 - name: build-sql-2012-enterprise-windows-2012-r2-dc
   plan:
   - get: compute-image-tools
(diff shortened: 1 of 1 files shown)


hopkiw:
i don't believe you can test a PR in draft mode using prow. You can take it out of draft but leave it in `do-not-merge/hold` status if you want to ensure it doesn't get prematurely merged.

For my part, this looks fine. Only one note to call out, you have created the correct linkages now based on your needs I think, but the derivative jobs still aren't actually using those resources. I would set it so the image version from the get gcs step is passed into the daisy workflow, rather than using the family. This can be a future improvement.

/lgtm

PR #328, opened 2021-12-15
Add rollout rate support for the publish to testing

LawrenceHwang:
We need to be able to publish to all zones without rollout policy. This PR enables the support to specify rollout rate in the workflows and task. The publish to testing workflows have the rollout rate set to `0`.


concourse/tasks/daisy-publish-images.yaml
```diff
@@ -12,13 +12,15 @@ inputs:
...
   GOOGLE_APPLICATION_CREDENTIALS: "credentials/credentials.json"
+  ROLLOUT_RATE: 60
 
 run:
   path: /gce_image_publish
   args:
   - -source_gcs_path=((source_gcs_path))
   - -source_version=((source_version))
   - -publish_version=((publish_version))
+  - -rollout_rate=$ROLLOUT_RATE
```
hopkiw, line 23: this form of referencing environment variables is a shell feature. this container doesn't run a shell, it runs the binary `/gce_image_publish`. so this string will be read literally.

what is your testing mechanism for this change?

concourse/pipelines/windows-image-build.yaml
```diff
@@ -3829,6 +3907,8 @@ jobs:
     file: publish-version/version # produced from generate-version task
   - task: publish-2019-for-containers
     file: guest-test-infra/concourse/tasks/daisy-publish-images.yaml
+    params:
+      ROLLOUT_RATE: 0
```
hopkiw, line 3911: i see you're setting this to 0 in many places, should the default be zero?

hopkiw (changes requested):
my understanding is rollout policies can be stored in the image publish workflows themselves, have we considered that option?

concourse/pipelines/windows-image-build.yaml
LawrenceHwang: 0 is desired value for those “publish to testing” workflows. The other publish jobs (to staging and prod) should still uses 60.

concourse/tasks/daisy-publish-images.yaml
LawrenceHwang: Thank you for the feedback and pointers. I've updated the PR with new task in this case.

LawrenceHwang:
> my understanding is rollout policies can be stored in the image publish workflows themselves, have we considered that option?

Thank for sharing the historical context. As chatted, the current implementation doesn't support describing rollout policies in the workflow. Will create a tracking bug.

hopkiw (approved):

PR #329, opened 2021-12-16
fix the wf reference

LawrenceHwang:


LawrenceHwang:
fix workflow reference

bkatyl (approved):

PR #330, opened 2021-12-22
add line coverage for guest agent

hopkiw:
* add new `publish-coverage` binary to the `concourse-metrics` container
* add gotest Concourse Task which invokes the gotest container originally written for prow
* add steps to `build-guest-agent` Concourse Job to produce and later publish coverage metrics

### DIFF
--- a/container_images/gotest/main.sh
+++ b/container_images/gotest/main.sh
@@ -36,7 +36,7 @@ fi
 cp /go-test.txt ${ARTIFACTS}/
 
 # $ARTIFACTS is provided by prow decoration containers
-go tool cover -func=/coverage.out | grep ^total | awk '{print $NF}' > ${ARTIFACTS}/coverage.txt
+go tool cover -func=/coverage.out | grep ^total | awk '{print $NF}' | tr -d '%' > ${ARTIFACTS}/coverage.txt
 cat /go-test.txt | go-junit-report > ${ARTIFACTS}/junit.xml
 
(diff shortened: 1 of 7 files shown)


container_images/gotest/main.sh
```diff
@@ -36,7 +36,7 @@ fi
 cp /go-test.txt ${ARTIFACTS}/
 
 # $ARTIFACTS is provided by prow decoration containers
-go tool cover -html=/coverage.out -o ${ARTIFACTS}/coverage.html
+go tool cover -func=/coverage.out | grep ^total | awk '{print $NF}' | tr -d '%' > ${ARTIFACTS}/coverage.txt
```
Linskeyd, line 39: Suggestion: save the full output in a place that will let use analyze where coverage is low/high (unless the junit report already does this for us).

hopkiw: no, the jUnit here only has test results. and it isn't saved in this invocation either.

you mean you want the 'coverage.out' file to be kept? I think we only ever need the latest, and if we come up with a use for it, we would wire it here. Otherwise I don't think it's necessary to create an archive of coverage details at different historical times.

Linskeyd: Ah is there a place where we keep the latest info already? I was mostly interested in just understanding what the current coverage is for the latest release (so we can understand if we have a directory that we think should be covered, but isn't).

PR #331, opened 2021-12-23
Publish stable tags for guest-agent

hopkiw:
* Publish `stable` and `stable-YYYYMMDD` tags on each stable promotion
* Publish a time-since-last-release metric

### DIFF
--- a/concourse/pipelines/guest-package-build.yaml
+++ b/concourse/pipelines/guest-package-build.yaml
@@ -303,6 +303,11 @@ jobs:
       name: last-stable-tag/stable  # file contains only the word 'stable'
       tag: last-stable-tag/stable
       commitish: guest-agent-tag/commit_sha
+  - put: guest-agent-tag
+    params:
+      name: last-stable-tag/stable-today  # file contains 'stable-YYYYMMDD'
+      tag: last-stable-tag/stable-today
+      commitish: guest-agent-tag/commit_sha
   on_success:
     task: success
     file: guest-test-infra/concourse/tasks/publish-job-result.yaml
--- a/concourse/pipelines/guest-package-build.yaml
+++ b/concourse/pipelines/guest-package-build.yaml
@@ -247,6 +247,8 @@ jobs:
   plan:
   - get: guest-agent-tag
     passed: [inject-guest-agent-staging]
+  - get: guest-agent
+    params: {fetch_tags: true}
   - get: guest-test-infra
   - task: generate-timestamp
     file: guest-test-infra/concourse/tasks/generate-timestamp.yaml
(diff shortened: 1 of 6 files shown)


concourse/tasks/publish-time-since.yaml
```diff
@@ -0,0 +1,22 @@
...
+- name: credentials
+
+params:
+  GOOGLE_APPLICATION_CREDENTIALS: "credentials/credentials.json"
+
+run:
+  path: /publish-time-since
+  args:
+  - -project-id=gcp-guest
+  - -metric-path=concourse/job/duration
```
Linskeyd, line 20: This may be a copy pasta error -> this metric should be different than the job duration metric. Maybe 'concourse/package/stable_stale_time' to follow the [naming conventions](https://cloud.google.com/monitoring/api/v3/naming-conventions)

concourse/pipelines/guest-package-build.yaml
```diff
@@ -285,6 +293,21 @@ jobs:
...
+  - put: guest-agent-tag
+    params:
+      name: last-stable-tag/stable  # file contains only the word 'stable'
+      tag: last-stable-tag/stable
+      commitish: guest-agent-tag/commit_sha
+  - put: guest-agent-tag
+    params:
+      name: last-stable-tag/stable-today  # file contains 'stable-YYYYMMDD'
+      tag: last-stable-tag/stable-today
+      commitish: guest-agent-tag/commit_sha
```
Linskeyd, line 310: Does this produce a 'stable-today' tag on the GH 'releases' page? Or does it produce 'stable-YYYYMMDD'?

concourse/pipelines/guest-package-build.yaml
```diff
@@ -285,6 +293,21 @@ jobs:
           universe: cloud-yum
           repo: google-guest-agent-el8
           environment: stable
+  - task: publish-time-since-stable
+    file: guest-test-infra/concourse/tasks/publish-time-since.yaml
+    vars:
+      last_publish_date: ((.:last-stable-date))
+      package_name: "google-guest-agent"
```
Linskeyd, line 300: This task is being added to the promote job for the package. This seems to imply that we would then only get stable staleness metric data points whenever a promotion is actually performed. Wouldn't this mean that we wouldn't be able to see something as being really stale until it is actually no longer stale (since we would have just run a promotion?)

An alternative approach to this would be a separate job which runs at a daily interval and does this calculation/metric publish directly for the package (rather than being coupled directly into the promotion to stable job).

Linskeyd (changes requested):

concourse/pipelines/guest-package-build.yaml
hopkiw: the latter; 'stable-today' is the name of the file within the Concourse job

concourse/tasks/publish-time-since.yaml
hopkiw: thanks for catching that, updated

concourse/pipelines/guest-package-build.yaml
hopkiw: this metric only produces staleness data for each release, associated with each release

Linskeyd: Talked with @hopkiw directly -> this metric will be used to measure the specific releases staleness; we can set up another job to run daily to measure both image and package release staleness (for more adhoc understandings of staleness of the latest release).

PR #333, opened 2022-01-06
Add JSONnet templating to linux-image-build pipeline

hopkiw:
The file 'linux-image-build.json' is generated from the file 'linux-image-build.jsonnet'. We will not check-in this generated file, this is a stopgap for reviewing until the presubmit is updated. The intention was also not to make changes to the final rendered pipeline - the rendered pipeline was compared to the existing, and only ordering changes were made, and one bug was fixed (debian-9-gcs was not set to skip_download: true).


concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -0,0 +1,460 @@
...
+      'rhel-8-byos',
+      'rhel-7-6-sap',
+      'rhel-7-7-sap',
+      'rhel-7-9-sap',
+      'rhel-8-1-sap',
+      'rhel-8-2-sap',
+      'rhel-8-4-sap',
+    ]),
+  ],
+}
```
hopkiw, line 460: 460 lines compared to 7,514 in the final rendering

PR #367, opened 2022-02-16
Add pipeline for building Daisy

EricEdens:
This adds a pipeline for building [compute-daisy](https://github.com/GoogleCloudPlatform/compute-daisy). To allow for validation, it currently pushes artifacts to compute-image-tools-test rather than compute-image-tools.

This PR adds a Dockerfile to compute-daisy: https://github.com/GoogleCloudPlatform/compute-daisy/pull/8

### DIFF
--- a/concourse/pipelines/daisy-build.yaml
+++ b/concourse/pipelines/daisy-build.yaml
@@ -10,8 +10,8 @@ resources:
 - name: compute-daisy
   type: git
   source:
-    uri: https://github.com/GoogleCloudPlatform/compute-daisy.git
-    branch: master
+    uri: https://github.com/hopkiw/compute-daisy.git
+    branch: dockerfile
 - name: compute-image-tools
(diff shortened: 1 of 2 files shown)


hopkiw (changes requested):

concourse/pipelines/daisy-build.yaml
```diff
@@ -0,0 +1,190 @@
...
+    bucket: compute-image-tools-test
+    json_key: |
+      ((concourse-sa.credential))
+    versioned_file: release/windows/daisy
+- name: darwin-executable
+  type: gcs
+  source:
+    bucket: compute-image-tools-test
+    json_key: |
+      ((concourse-sa.credential))
```
hopkiw, line 43: i don't think this is the right secret name? it looks in the template to be `((gcs-key.credential))`

PR #369, opened 2022-02-17
Update container build workflow

hopkiw:
* convert to jsonnet
* add missing container builds
* remove credential steps
* removed pinned kaniko version


concourse/pipelines/container-build.jsonnet
```diff
@@ -0,0 +1,149 @@
...
+};
+
+// Start of output.
+{
+  resources: [
+    common.GitResource('guest-test-infra'),
+    common.GitResource('compute-image-tools') {
+      source+: { paths: ['daisy_workflows/**'] },
+    },
+  ],
```
Linskeyd, line 90: Does this need the git resource for compute-daisy?

From the fly validate pipelines step on prow:

```
Error loading existing config:
  - invalid jobs:
	jobs.build-daisy.plan.do[0].get(compute-daisy): unknown resource 'compute-daisy'
```

Linskeyd (commented):
Looks like there is a failure still on tests; otherwise the code overall lgtm. Thanks for doing this!

PR #370, opened 2022-02-17
Sap task

patelne:


concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,78 @@
...
+      run:
+        path: sh
+        args:
+        - -exc
+        - |
+          if gsutil cat gs://test-bucket-for-terraform/log_output | grep "ERROR - Deployment Exited" | grep -v grep; then
+            return 1
+          else
+            return 0
+          fi
```
hopkiw, line 70: still think this should be output from the job, but a few fixes.
1. `grep -q` instead of `| grep -v grep`
2. `return` is only valid in a function context
3. we set exit on errors with the `-exc` args; so we only need to run a command that fails
4. swap the logic - if we don't see success, it's a failure

```suggestion
          gsutil cat gs://test-bucket-for-terraform/log_output | grep -q "SUCCESS"
```

hopkiw (changes requested):

PR #390, opened 2022-02-28
ARM image tests

hopkiw:
* Add `-machine_type` support to CIT
* Add image tests to `debian-11-arm64` image build
* Use metadata lookup for interface lookups in tests rather than predictive naming


imagetest/cmd/manager/main.go
```diff
@@ -35,6 +35,7 @@ var (
 	parallelCount = flag.Int("parallel_count", 5, "TestParallelCount")
 	filter        = flag.String("filter", "", "only run tests matching filter")
 	exclude       = flag.String("exclude", "", "skip tests matching filter")
+	machineType   = flag.String("machine_type", "", "machine type")
```
zoran15, line 38: I guess machine type for VMs running the test? Extra description would be useful.

imagetest/testworkflow.go
```diff
@@ -272,8 +272,19 @@ func finalizeWorkflows(ctx context.Context, tests []*TestWorkflow, zone, gcsPref
...
-		twf.wf.Sources["testpackage"] = fmt.Sprintf("/%s.test", twf.Name)
+		arch := "amd64"
+		if machineType != "" {
+			createVMsStep := twf.wf.Steps[createVMsStepName]
+			for _, vm := range createVMsStep.CreateInstances.Instances {
+				vm.MachineType = machineType
+			}
+			if strings.HasPrefix(machineType, "t2a") {
+				arch = "arm64"
+			}
```
zoran15, line 283: Is this logic going to hold forever? I'm not very familiar with how machine types map to the architecture.

zoran15 (approved):
Only reviewed the Go part of the code. A couple of minor comments.

imagetest/testworkflow.go
hopkiw: t2a will always be ARM, but there may be new ARM types in the future. the Right Thing To Do™ here is to use compute API and read arch from machine type, but that's not yet available in prod API, and would also be the first direct use of compute API in this code.

zoran15: SGTM. For tracking, I would personally create a bug to update this code once the API is available.

zoran15 (approved):

PR #394, opened 2022-02-28
Add SAP workload test job

patelne:
Add sap pipeline for testing


concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+        - -exc
+        - |
+          cat <<EOT >> sap_hana.tf
+          terraform {
+            backend "gcs" {
+              bucket = "test-bucket-for-terraform"
+              prefix = "((.:id))/terraform"
+            }
+          }
+          module "sap_hana" {
```
patelne, line 68: @hopkiw You mentioned if we could move this chunk of code into its own file.  I'm keeping it here because we'll need to set some of the variables during script execution (like the instance_name, post_deployment_script and image information).

Do you think it's still appropriate to have a separate file?

hopkiw: we could have flat, checked-in files. and parameterize them with `-var:` arguments at runtime. i think that is a better model than generating the file dynamically, wdyt?

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+        source:
+          repository: google/cloud-sdk
+          tag: latest
+      run:
+        path: sh
+        args:
+        - -exc
+        - |
+          # Clean up files created as part of this job
+          gsutil rm -r -f gs://test-bucket-for-terraform/((.:id))
```
hopkiw, line 149: i think we can drop this. gcs is cheap and the whole point of unique IDs is so we can go back and investigate.

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+            sleep 60
+          done
+
+          gsutil cat gs://test-bucket-for-terraform/((.:id))/run_result | grep -q "SUCCESS"
+
+  - task: destroy-sap-tf-environment
+    config:
+      platform: linux
+      image_resource:
+        type: docker-image
```
hopkiw, line 112:
```suggestion
        type: registry-image
```

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+          terraform init -upgrade
+          terraform apply -auto-approve
+
+  - task: wait-for-and-check-post-script-results
+    timeout: 30m
+    config:
+      platform: linux
+      image_resource:
+        type: docker-image
```
hopkiw, line 89:
```suggestion
        type: registry-image
```

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+          subnetwork = "default"
+          linux_image = "rhel-8-4-sap-ha"
+          linux_image_project = "rhel-sap-cloud"
+          instance_name = "hana-instance-((.:id))"
+          post_deployment_script = "gs://test-bucket-for-terraform/((.:id))/sap_post_script.sh"
+          }
+          EOT
+
+          terraform init -upgrade
+          terraform apply -auto-approve
```
hopkiw, line 82: does this block until completion? or return immediately?

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+        source:
+          repository: google/cloud-sdk
+          tag: latest
+      run:
+        path: sh
+        args:
+        - -exc
+        - |
+          # Loop until we see the result file in the bucket.
+          # This task has a timeout of 30 minutes. So either this loop will break or the task will abort.
```
hopkiw, line 99: we have to poll? the terraform commands above will return before the workflow is complete?

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+          cat <<EOT >> sap_hana.tf
+          terraform {
+            backend "gcs" {
+              bucket = "test-bucket-for-terraform"
+              prefix = "((.:id))/terraform"
+            }
+          }
+          module "sap_hana" {
+          source = "https://storage.googleapis.com/cloudsapdeploy/terraform/latest/terraform/sap_hana/sap_hana_module.zip"
+          project_id = "pneil-sandbox"
```
hopkiw, line 70: replace with prod project

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+          zone = "us-east1-b"
+          machine_type = "n1-highmem-32"
+          subnetwork = "default"
+          linux_image = "rhel-8-4-sap-ha"
+          linux_image_project = "rhel-sap-cloud"
+          instance_name = "hana-instance-((.:id))"
+          post_deployment_script = "gs://test-bucket-for-terraform/((.:id))/sap_post_script.sh"
```
hopkiw, line 77: which project owns this bucket? do we need to replace with a bucket owned by one of our prod projects?

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+          gsutil cp result.txt gs://test-bucket-for-terraform/((.:id))/run_result
+          EOT
+
+          gsutil cp sap_post_script.sh gs://test-bucket-for-terraform/((.:id))/sap_post_script.sh
+
+  - task: create-sap-tf-environment
+    config:
+      platform: linux
+      image_resource:
+        type: docker-image
```
hopkiw, line 52:
```suggestion
        type: registry-image
```

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+        source:
+          repository: google/cloud-sdk
+          tag: latest
+      run:
+        path: sh
+        args:
+        - -exc
+        - |
+          cat <<EOT >>sap_post_script.sh
```
hopkiw, line 37: let's check this file in, too

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+      image_resource:
+        type: registry-image
+        source:
+          repository: busybox
+      run:
+        path: sh
+        args:
+        - -exc
+        - |
+          # TODO - make more unique? guid?
```
hopkiw, line 16: look at our other examples - we often use date based IDs rather than random ones, which helps also with discovery

concourse/pipelines/sap-workload-test.yaml
```diff
@@ -0,0 +1,149 @@
...
+          terraform {
+            backend "gcs" {
+              bucket = "test-bucket-for-terraform"
+              prefix = "((.:id))/terraform"
+            }
+          }
+          EOT
+
+          terraform init
+          terraform destroy -auto-approve
```
hopkiw, line 132: what does this do? we have a new tf file and we use it for destroy? why not call destroy on the original file? and will this delete the bucket, or just the objects? (in which case, i would second advice from below: we don't need to clean up gcs objects)

hopkiw (changes requested):

concourse/pipelines/sap-workload-test.yaml
patelne: Done. Passed along the terraform state file instead of using the gcs backend.

concourse/pipelines/sap-workload-test.yaml
hopkiw: what is the *project* which owns this bucket?

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -4,6 +4,7 @@ local common = import '../templates/common.libsonnet';
 local daisy = import '../templates/daisy.libsonnet';
 local gcp_secret_manager = import '../templates/gcp-secret-manager.libsonnet';
 local lego = import '../templates/lego.libsonnet';
+local sap_test = import '../templates/sap-test.libsonnet';
```
hopkiw, line 7: i think this is only needed if we are gonna reuse the stuff, otherwise it can just be declared inline

concourse/pipelines/sap-workload-test.yaml
patelne: Oh yeah sorry, it should be `gcp-guest`.

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -385,16 +386,40 @@ local DebianImgPublishJob(image, env, workflow_dir) = imgpublishjob {
...
+};
+
+local RhelImgGroup(images) = imggroup {
+  group_name: 'rhel',
+  images: images,
+  extra_job_groups: [
+    'sap-workload-test-%s' % [image]
+    for image in std.filter(function(x) std.endsWith(x, '-sap') , images)
+  ]
 };
```
hopkiw, line 423: please restore this, you just want the rhel jobs to have extra jobs added we can do that with a composition right at the invocation point. i'll show an example of that.

concourse/templates/sap-test.libsonnet
```diff
@@ -0,0 +1,150 @@
...
+        ],
+        run: {
+          path: 'sh',
+          args: [
+            '-exc',
+            |||
+            cd guest-test-infra/concourse/workload-tests/SAP/
+            # We want to upload this actual script with the unique id
+            sed -i 's/$1/gcp-guest-test-outputs/g' sap_post_script.sh
+            sed -i 's/$2/((.:id))/g' sap_post_script.sh
```
hopkiw, line 50: took me a while to figure out what this was trying to do bc you used bash variables as the replacement markers. typically it's best practice to do e.g. `__BUCKET__` and `__RUNID__` instead in the template. i also think this is just about as ugly as a heredoc, but fine to start with.

concourse/templates/sap-test.libsonnet
```diff
@@ -0,0 +1,150 @@
...
+            cp guest-test-infra/concourse/workload-tests/SAP/sap_hana.tf tf-state/
+            cd tf-state
+
+            terraform init
+            terraform init -upgrade
+            terraform apply -auto-approve \
+              -var="instance_name=hana-instance-((.:id))" \
+              -var="post_deployment_script=gs://gcp-guest-test-outputs/workload-tests/sap/((.:id))/sap_post_script.sh" \
+|||
+            + "-var='linux_image=%(image)s-ha'" % {image: tl.image},
```
hopkiw, line 88: you can add the % to the end, no need to split this last line out of the multiline string block
```suggestion
            -var='linux_image=%s-ha'"
            ||| % {tl.image}
```

concourse/templates/sap-test.libsonnet
```diff
@@ -0,0 +1,150 @@
...
+        platform: 'linux',
+        image_resource: {
+          type: 'registry-image',
+          source: {
+            repository: 'hashicorp/terraform',
+            tag: 'latest',
+          },
+        },
+        inputs: [
+          { name: "tf-state" },
```
hopkiw, line 134: if you specify the path for this input as '.', it will be the current directory for the task. then you can use terraform args directly instead of a shell script (thankful the tf docker image even supports shell scripts)

```suggestion
          { name: "tf-state", path: "." },
```

concourse/workload-tests/SAP/sap_post_script.sh
```diff
@@ -0,0 +1,6 @@
+if cat /var/log/messages | grep -q "ERROR - Deployment Exited"; then
+    echo "ERROR" > result.txt
+else
+    echo "SUCCESS" > result.txt
+fi
+gsutil cp result.txt gs://$1/workload-tests/sap/$2/run_result
```
hopkiw, line 6: as mentioned above, `$1` is not a good replacement anchor

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -385,16 +386,40 @@ local DebianImgPublishJob(image, env, workflow_dir) = imgpublishjob {
   image_prefix: common.debian_image_prefixes[image],
 };
 
-local ImgGroup(name, images) = {
-  name: name,
+local SapWorkloadTestJob(image) = sap_test.saptestjob {
+  image: image,
+};
```
hopkiw, line 391: there's no logic in this function, so just replace it with direct use of the template

hopkiw (changes requested):

concourse/templates/sap-test.libsonnet
```diff
@@ -0,0 +1,150 @@
...
+        },
+        inputs: [
+          { name: "guest-test-infra" },
+        ],
+        run: {
+          path: 'sh',
+          args: [
+            '-exc',
+            |||
+            cd guest-test-infra/concourse/workload-tests/SAP/
```
hopkiw, line 47: 1. this could also be replaced by setting `path:` in the input
2. let's put the scripts in `concourse/scripts` not `concourse/workload-testing/SAP`

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -486,6 +635,13 @@ local ImgGroup(name, images) = {
           for env in envs
           for image in rhel_images
         ] +
+        [
+          // SAP related test jobs
+          saptestjob {
+            image: image
+          }
+          for image in std.filter(function(x) std.endsWith(x, '-sap'), rhel_images)
```
hopkiw, line 643: just list them out rather than this kind of thing, i can't tell what it will be by reading. if it's long, create a list with `local sap_images = [image,image];`

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -385,6 +385,155 @@ local DebianImgPublishJob(image, env, workflow_dir) = imgpublishjob {
...
+      outputs: [
+        { name: "tf-state" },
+      ],
+      run: {
+        path: 'sh',
+        args: [
+          '-exc',
+          |||
+          cp guest-test-infra/concourse/scripts/sap_hana.tf tf-state/
+          cd tf-state
```
hopkiw, line 467: this won't work, cp will complain

why make a copy? why not use `dir:` arg and avoid use of shell altogether?

hopkiw (changes requested):

concourse/pipelines/linux-image-build.jsonnet
patelne: It looks like having these separate named outputs and inputs causes issues. If I set output as below and set `
        dir: 'guest-test-infra/concourse/scripts'`, the scripts folder will now be empty (the task won't see any files there).
```
      outputs: [
        { name: "tf-state", path: "guest-test-infra/concourse/scripts" },
      ],
```

I think setting the outputs field will init the folder and clear it out. Do you know a way to avoid this without doing this cp/cd stuff?

hopkiw (approved):

PR #398, opened 2022-03-02
Use gcloud-package-operation task for promoting to staging

Quintonamore:
Previously some steps in our package build pipelines were using the `gcloud-promote-package` task to promote a package to staging, using this would require an engineer's time to manually call ARLE to release the package.

Instead, since promoting to staging isn't a protected action, we can use the `gcloud-package-operation` task with the `promoteToStaging` option so that the promotion to staging is done automatically without any manual action.


Quintonamore:
Closing this PR since the templates in #476 utilize the promote to staging workflow.

PR #401, opened 2022-03-04
Daisy build: Add pre-release tests

EricEdens:
Following this PR, the daisy artifacts will be published to the following locations:

- Container image:
   - Every build is tagged with the commit's SHA
   - If the tests pass, then the `release` and `latest` tags are applied to the build
- Windows, Linux, and OSX binaries:
   - If tests pass, upload to `gs://compute-image-tools/{latest,release}/{linux,windows,darwin}`

This is consistent with the [legacy cloud build](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/v20220209/cli_tools_cloudbuild.yaml#L30-L33) configuration that used to build daisy.

The following tests are added in this PR:
- Run all workflows in [compute-daisy/daisy_integration_tests](https://github.com/GoogleCloudPlatform/compute-daisy/tree/master/daisy_integration_tests) using the daisy test runner.
- Run one workflow with the staged container

### DIFF
--- a/concourse/pipelines/container-build.jsonnet
+++ b/concourse/pipelines/container-build.jsonnet
@@ -66,7 +66,7 @@ local buildcontainerimgjob = {
           {
             task: 'build-image',
             config: buildcontainerimgtask {
-              commit_sha: "((.:%s-commit-sha))" % job.image,
+              commit_sha: '((.:%s-commit-sha))' % job.image,
               destination: job.destination,
               dockerfile: job.dockerfile,
               context: job.context,
@@ -88,12 +88,22 @@ local BuildContainerImage(image) = buildcontainerimgjob {
 // Start of output.
 {
   local daisy_architectures = ['linux', 'windows', 'darwin'],
-
+  resource_types: [{
+    name: 'registry-image-forked',
+    type: 'registry-image',
+    source: { repository: 'gcr.io/compute-image-tools/registry-image-forked' },
+  }],
   resources: [
     common.GitResource('guest-test-infra'),
     common.GitResource('compute-image-tools'),
-    common.GitResource('compute-image-tools-trigger') {
-      source+: { paths: ['daisy_workflows/**'] },
+    {
+      name: 'compute-image-tools-trigger',
+      type: 'git',
+      source: {
+        uri: 'https://github.com/GoogleCloudPlatform/compute-image-tools.git',
+        branch: 'master',
+        paths: ['daisy_workflows/**'],
+      },
     },
     common.GitResource('compute-daisy'),
   ],
@@ -150,7 +160,7 @@ local BuildContainerImage(image) = buildcontainerimgjob {
       destination: 'gcr.io/compute-image-tools/daisy',
       image: 'daisy',
       input: 'compute-daisy',
-
+      passed: 'build-daisy-test-runner',
       extra_resources: ['compute-image-tools-trigger'],
       extra_steps:
         [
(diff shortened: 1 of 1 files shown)


EricEdens:
cc @hopkiw @adjackura

concourse/pipelines/container-build.jsonnet
```diff
@@ -194,9 +203,61 @@ local BuildContainerImage(image) = buildcontainerimgjob {
             },
           }
           for arch in daisy_architectures
-        ] +
-        //  Put three binaries using gsutil.
-        [
+        ],
+      plan:
+        super.plan + [
```
hopkiw, line 208:
```suggestion
      plan+: [
```

concourse/pipelines/container-build.jsonnet
```diff
@@ -215,8 +276,32 @@ local BuildContainerImage(image) = buildcontainerimgjob {
...
+                source: { repository: 'google/cloud-sdk', tag: 'alpine' },
+              },
+              run: {
+                path: 'sh',
+                args: [
+                  '-exc',
+                  'for t in latest release; do' +
+                  '  gcloud container images add-tag --quiet' +
+                  '    gcr.io/compute-image-tools/daisy:((.:compute-daisy-commit-sha))' +
+                  '    gcr.io/compute-image-tools/daisy:$t;' +
```
hopkiw, line 304: so it's your intention to keep both tags up to date, here? so each new image version gets 3?

concourse/pipelines/container-build.jsonnet
```diff
@@ -215,8 +276,32 @@ local BuildContainerImage(image) = buildcontainerimgjob {
...
+                  '    gsutil acl ch -u AllUsers:R gs://compute-image-tools/$t/$f/daisy;' +
+                  '  done;' +
+                  'done',
+                ],
+              },
+            },
+          },
+          //  Add release and latest tags to the staged container.
+          {
+            task: 'promote-container',
```
hopkiw, line 290: nit: tag-image

concourse/pipelines/container-build.jsonnet
```diff
@@ -194,9 +203,61 @@ local BuildContainerImage(image) = buildcontainerimgjob {
...
+                  tag: 'latest',
+                  google_auth: true,
+                },
+              },
+              run: {
+                path: '/daisy_test_runner',
+                args: [
+                  '-projects=compute-image-test-pool-001',
+                  '-zone=us-central1-c',
+                  'compute-daisy/daisy_integration_tests/daisy_e2e.test.gotmpl',
```
hopkiw, line 232: it's not 100% clear to me how this tests the newly built daisy without any reference to it; looking at it, it builds daisy test runner from source, and has no relation to the daisy image build. should we link them together somehow?

hopkiw (changes requested):

concourse/pipelines/container-build.jsonnet
EricEdens: Good question -- my intent was to keep the same behavior as the [current cloudbuild results](https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/cli_tools_cloudbuild.yaml). If I were creating this new, I'd propose two tags (the git sha and `latest`), but since existing people might be using `release`, I'd propose we keep publishing to that tag.

concourse/pipelines/container-build.jsonnet
EricEdens: Oh interesting -- since the daisy test runner is built at [L138](https://github.com/GoogleCloudPlatform/guest-test-infra/blob/5c7e8aec92201eee01a68f433ecff5c28b61d96c/concourse/pipelines/container-build.jsonnet#L138), my assumption was this would reference *that* build result. Is my assumption incorrect? If so, what would you suggest?

concourse/pipelines/container-build.jsonnet
EricEdens: done!

concourse/pipelines/container-build.jsonnet
EricEdens: Changed the task's name to `tag-image`

concourse/pipelines/container-build.jsonnet
hopkiw: These are currently independent jobs with no relationship, so presumably they could be run in any order or even simultaneously. To fix, if that test runner is also updated to tag with the git commit like we're doing here, we could then add a `passed:[]` type dependency to that job, and reference the image by tag rather than by 'latest' to ensure we're testing the same code. In fact why don't we update the base template to add the step to get the commit and also tag it, so this becomes a norm for all these images?

EricEdens: good idea! done.

hopkiw (approved):

PR #405, opened 2022-03-07
cds image build: correct input

hopkiw:
append rather than overwriting; the workflow itself is in compute-image-tools


wrigri (approved):
/lgtm

PR #410, opened 2022-03-09
generate CSR

hopkiw:
ACME expects *two* keys, one for the account and one for the final cert. Additionally, the LEGO tool won't reuse an existing key unless we use the CSR path.

* remove unneeded lego template
* standardize secret names
* add separate rhui TLS key secret
* generate CSR for lego


concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -230,25 +230,68 @@ local CDSImgBuildJob(image, workflow) = imgbuildjob {
...
+          source: { repository: 'alpine/openssl' },
+        },
+        run: {
+          path: 'openssl',
+          args: [
+            'req',
+            '-new',
+            '-key=./gcp-secret-manager/rhui_tls_key.pem',
+            '-subj=/CN=rhui.googlecloud.com',
+            '-out=./gcp-secret-manager/thecsr.pem',
```
EricEdens, line 257: You don't need to declare `gcp-secret-manager` as an input and output?

hopkiw: thanks for catching that!

PR #418, opened 2022-03-14
Fix Windows publish workflows reference.

zmarano:


zmarano:
Ah- should be the other way around.

PR #423, opened 2022-03-21
remove unused arg 'gcs-key.credential'

dntczdx:
We are not reading credentials in a hard-coded way, so these args are not useful anymore.


EricEdens:
seems fine to me; adding liam

/assign @hopkiw

hopkiw (approved):

PR #425, opened 2022-03-21
run ARM tests for guest agent based on test image

dntczdx:
Build the test image by installing the guest-agent arm64 lib, and run it.
To support that, we added machine-type and worker-image as the task args.

### DIFF
--- a/concourse/pipelines/guest-package-build.yaml
+++ b/concourse/pipelines/guest-package-build.yaml
@@ -199,11 +199,19 @@ jobs:
           gcs-package-path: gs://gcp-guest-package-uploads/guest-agent/google-guest-agent-((.:package-version))-g1.el8.x86_64.rpm
           machine-type: e2-medium
           worker-image: projects/compute-image-tools/global/images/family/debian-10-worker
-  - task: guest-agent-image-tests
-    file: guest-test-infra/concourse/tasks/image-test-args.yaml
-    vars:
-      images: projects/gcp-guest/global/images/debian-9-((.:build-id)),projects/gcp-guest/global/images/debian-10-((.:build-id)),projects/gcp-guest/global/images/debian-11-((.:build-id)),projects/gcp-guest/global/images/debian-11-arm64-((.:build-id)),projects/gcp-guest/global/images/centos-7-((.:build-id)),projects/gcp-guest/global/images/rhel-7-((.:build-id)),projects/gcp-guest/global/images/rhel-8-((.:build-id))
-      extra-args: "-exclude=(image)|(disk)|(security)|(oslogin)"
+  - in_parallel:
+      fail_fast: true
+      steps:
+      - task: guest-agent-image-tests-amd64
+        file: guest-test-infra/concourse/tasks/image-test-args.yaml
+        vars:
+          images: projects/gcp-guest/global/images/debian-9-((.:build-id)),projects/gcp-guest/global/images/debian-10-((.:build-id)),projects/gcp-guest/global/images/debian-11-((.:build-id)),projects/gcp-guest/global/images/centos-7-((.:build-id)),projects/gcp-guest/global/images/rhel-7-((.:build-id)),projects/gcp-guest/global/images/rhel-8-((.:build-id))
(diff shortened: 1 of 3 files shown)


dntczdx:
Hold it in order to wait for the publish of debian-11-worker-arm64 image.

concourse/pipelines/guest-package-build.yaml
```diff
@@ -149,40 +149,60 @@ jobs:
...
           dest-image: rhel-8-((.:build-id))
           gcs-package-path: gs://gcp-guest-package-uploads/guest-agent/google-guest-agent-((.:package-version))-g1.el8.x86_64.rpm
+          machine-type: e2-medium
+          worker-image: projects/compute-image-tools/global/images/family/debian-10-worker
   - task: guest-agent-image-tests
     file: guest-test-infra/concourse/tasks/image-test-args.yaml
     vars:
-      images: projects/gcp-guest/global/images/debian-9-((.:build-id)),projects/gcp-guest/global/images/debian-10-((.:build-id)),projects/gcp-guest/global/images/debian-11-((.:build-id)),projects/gcp-guest/global/images/centos-7-((.:build-id)),projects/gcp-guest/global/images/rhel-7-((.:build-id)),projects/gcp-guest/global/images/rhel-8-((.:build-id))
+      images: projects/gcp-guest/global/images/debian-9-((.:build-id)),projects/gcp-guest/global/images/debian-10-((.:build-id)),projects/gcp-guest/global/images/debian-11-((.:build-id)),projects/gcp-guest/global/images/debian-11-arm64-((.:build-id)),projects/gcp-guest/global/images/centos-7-((.:build-id)),projects/gcp-guest/global/images/rhel-7-((.:build-id)),projects/gcp-guest/global/images/rhel-8-((.:build-id))
```
hopkiw, line 205: i don't think this will work - the VMs are launched by default on e2 family instances. we can't intermix x86 and arm64 images in one test run unless we make some changes to CIT. so probably we just want to add a second task (can be in parallel with this one) that specifies the t2a machine type, the arm-enabled zone, and the arm image.

concourse/pipelines/guest-package-build.yaml
```diff
@@ -149,40 +149,60 @@ jobs:
...
           gcs-package-path: gs://gcp-guest-package-uploads/guest-agent/google-guest-agent_((.:package-version))-g1_amd64.deb
+          machine-type: e2-medium
+          worker-image: projects/compute-image-tools/global/images/family/debian-10-worker
       - task: build-package-image-debian-10
         file: guest-test-infra/concourse/tasks/daisy-build-package-image.yaml
         vars:
           source-image: projects/debian-cloud/global/images/family/debian-10
           dest-image: debian-10-((.:build-id))
           gcs-package-path: gs://gcp-guest-package-uploads/guest-agent/google-guest-agent_((.:package-version))-g1_amd64.deb
+          machine-type: e2-medium
```
hopkiw, line 160: just a note, having to specify the machine type and not being able to set a default is a limitation of separated task files, and goes away when we move this pipeline to a jsonnet template.

hopkiw (changes requested):

concourse/pipelines/guest-package-build.yaml
dntczdx: My understanding is that, the "manager"  is just a launcher for the test workflows. So, ideally, it can launch whatever machine type dynamically depending on the test image. I thought that has been done.

Let me put it in a separate task for now.

hopkiw: the 'manager' is an internal implementation detail of CIT; CIT will automatically detect the need to specify ARM binaries if you're using ARM machine type, but you can't launch all of these images on the same machine type, and the machine type is not automatically detected but passed in as an argument

dntczdx: done

hopkiw (approved):

dntczdx:
It's still pending on 2 things:
1. debian-11-worker-arm64 release to PROD.
2. successfully built the guest-agent package.

PR #426, opened 2022-03-22
upload ARM packages

dntczdx:
Upload ARM apckages.

Hold until the "upload" operation supports multiple files uploading.

### DIFF
--- a/concourse/pipelines/guest-package-build.yaml
+++ b/concourse/pipelines/guest-package-build.yaml
@@ -206,7 +206,7 @@ jobs:
       - task: upload-guest-agent-deb11
         file: guest-test-infra/concourse/tasks/gcloud-package-operation.yaml
         vars:
-          package_path: guest-agent/google-guest-agent_((.:package-version))-g1_amd64.deb
+          package_path: guest-agent/google-guest-agent_((.:package-version))-g1_amd64.deb;guest-agent/google-guest-agent_((.:package-version))-g1_arm64.deb
           universe: cloud-apt
           repo: google-guest-agent-bullseye
         params:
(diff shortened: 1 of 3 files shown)


concourse/pipelines/guest-package-build.yaml
```diff
@@ -206,7 +206,7 @@ jobs:
       - task: upload-guest-agent-deb11
         file: guest-test-infra/concourse/tasks/gcloud-package-operation.yaml
         vars:
-          package_path: guest-agent/google-guest-agent_((.:package-version))-g1_amd64.deb
+          package_path: guest-agent/google-guest-agent_((.:package-version))-g1_amd64.deb;guest-agent/google-guest-agent_((.:package-version))-g1_arm64.deb
```
hopkiw, line 209: i think we decided to use a real list in the RPC rather than a delimiter

hopkiw (changes requested):

concourse/pipelines/guest-package-build.yaml
dntczdx: yes. it's constructed in gcloud-package-operation.yaml now, since we don't want to put a complex json string here

hopkiw: this still has semicolons, is that right?

concourse/tasks/gcloud-package-operation.yaml
```diff
@@ -17,4 +17,13 @@ run:
...
   - |
-    gcloud pubsub topics publish $TOPIC --message "{\"type\": \"${TYPE}\", \"request\": {\"bucket\": \"gcp-guest-package-uploads\", \"object\": \"((package_path))\", \"universe\": \"((universe))\", \"repo\": \"((repo))\"}}"
+    paths=$(echo "((package_path))" | tr ";" "\n")
+    for path in $paths
+    do
+      if [[ "${objects}" != "" ]]; then
+        objects="${objects}, "
+      fi
+      objects="${objects}{\"bucket\":\"gcp-guest-package-uploads\",\"object\":\"${path}\"}"
+    done
```
hopkiw, line 27: can't we just write the input to match what this task expects? this is overly complicated

concourse/pipelines/guest-package-build.yaml
dntczdx: modified to directly passing in json string, no more

concourse/tasks/gcloud-package-operation.yaml
dntczdx: done

hopkiw (approved):

PR #431, opened 2022-03-23
build & publish windows-install-media

dntczdx:
This is missed on both previous build framework and current concourse. Add it to concourse to make it completely managed.

### DIFF
--- a/concourse/pipelines/windows-install-media-image-build.yaml
+++ b/concourse/pipelines/windows-install-media-image-build.yaml
@@ -16,6 +16,11 @@ resources:
   source:
     uri: https://github.com/GoogleCloudPlatform/guest-test-infra.git
     branch: master
+- name: windows-install-media-gcs
+  type: gcs
+  source:
+    bucket: artifact-releaser-prod-rtp
+    regexp: "windows-install-media/windows-install-media-v([0-9]+).tar.gz"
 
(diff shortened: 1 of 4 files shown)


dntczdx:
/hold
The compute-image-tools workflows are not ready yet. Hold.

concourse/pipelines/windows-install-media-image-build.yaml
```diff
@@ -0,0 +1,202 @@
...
+  - load_var: gcs-url
+    file: windows-install-media-gcs/url
+  - task: generate-build-date
+    file: guest-test-infra/concourse/tasks/generate-version.yaml
+  - load_var: build-date
+    file: publish-version/version
+  - task: daisy-build-windows-install-media
+    file: guest-test-infra/concourse/tasks/daisy-build-images-debian.yaml
+    vars:
+      wf: "windows_media/windows-install-media.wf.json"
```
bkatyl, line 53: When the files and workflow are added to compute-image-tools, please place them in daisy_workflows/image_build/windows

The ISO and update file references  need to be changed to be required values passed in by concourse as all the values are in secret manager.

The publish file needs to be placed in daisy_workflows/build-publish)/windows and we should meet about updating the publish project that is currently being used.

PR #436, opened 2022-03-29
comment out guest-agent-publish-coverage since it is broken 

dntczdx:
Will bring it back when we resolved the build issue on debian 11 arm64.


dntczdx:
reopen

hopkiw (approved):

PR #442, opened 2022-04-15
CIT: Support SSH connections to hosts that block SHA1

EricEdens:
The SSH tests in CIT use Go to establish SSH connections:

https://github.com/GoogleCloudPlatform/guest-test-infra/blob/99181c466e5d93095615cfc52edfa16499e71292/imagetest/utils/test_utils.go#L192

While running CIT against Ubuntu 22.04, I observed that the SSH tests were failing:

```
	<testsuite name="ssh-daily-ubuntu-2204-jammy-v20220415" tests="5" failures="2" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="ssh-daily-ubuntu-2204-jammy-v20220415" name="TestMatchingKeysInGuestAttributes" time="0.02"></testcase>
		<testcase classname="ssh-daily-ubuntu-2204-jammy-v20220415" name="TestHostKeysAreUnique" time="60.2">
			<failure type="">    host_key_test.go:59: connect to remote host at 1650057165249076651&#xA;    host_key_test.go:62: user test-user failed ssh to target host, vm2-ssh-3jtbh, err ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain</failure>
		</testcase>
		<testcase classname="ssh-daily-ubuntu-2204-jammy-v20220415" name="TestSSHInstanceKey" time="60.14">
			<failure type="">    image_ssh_test.go:33: connect to remote host at 1650057225390562509&#xA;    image_ssh_test.go:36: user test-user failed ssh to target host, vm2-ssh-3jtbh, err ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain</failure>
		</testcase>
		<testcase classname="ssh-daily-ubuntu-2204-jammy-v20220415" name="TestEmptyTest" time="0"></testcase>
		<testcase classname="ssh-daily-ubuntu-2204-jammy-v20220415" name="TestHostKeysNotOverrideAfterAgentRestart" time="0.12"></testcase>
	</testsuite>
```

This occurred since Ubuntu 22.04 updated to an OpenSSH version that blocks connections that use SHA-1: https://www.openssh.com/releasenotes.html

This PR updates the `golang.org/x/crypto` dependency, which brings in their PR to support SHA-2 keys: https://github.com/golang/go/issues/37278

It also does the following:
- Creates a go.mod file in the `imagetest` directory
- Updates the container build configurations to build relative to this directory


hopkiw:
/hold

so the main issue here is we need to confirm changing the shared go.mod doesn't break *any* of the containers, and if it does, we'll have to create per-project go modules either for the old or the new

EricEdens:
> so the main issue here is we need to confirm changing the shared go.mod doesn't break any of the containers, and if it does, we'll have to create per-project go modules either for the old or the new

@hopkiw 

Ahh, this creates a *new* go.mod and leaves the old one unchanged.

hopkiw (approved):

PR #451, opened 2022-04-22
RHUI roll-out: Wait for MIGs to be stable

EricEdens:
Within each region, we start a rolling update, then wait for the version target to be reached. The MIG exposes a "stable" state, which means that all of the nodes are passing their health check.

This PR adds an additional step to wait for this stable signal.

More info on the status fields: https://cloud.google.com/compute/docs/instance-groups/rolling-out-updates-to-managed-instance-groups#group_status

### DIFF
--- a/concourse/pipelines/rhui-release.yaml
+++ b/concourse/pipelines/rhui-release.yaml
@@ -34,7 +34,7 @@ jobs:
         - --region=us-west1
         - --project=google.com:rhel-infra
         path: gcloud
-    task: deploy-rhua
+    task: rhua-start-rolling-update
   - config:
       image_resource:
         source:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/rhui-release.yaml
```diff
@@ -54,7 +54,27 @@ jobs:
...
+          tag: alpine
+        type: registry-image
+      platform: linux
+      run:
+        args:
+        - compute
+        - instance-groups
+        - managed
+        - wait-until
+        - --stable
```
hopkiw, line 71: this is in the wait-for-version-target step?

hopkiw (changes requested):

concourse/pipelines/rhui-release.yaml
EricEdens: the jsonnet renderer puts the `task` key at the bottom of the object (I'm revisiting the jsonnet now to replace what's here)

hopkiw (approved):

PR #452, opened 2022-04-22
Use jsonnet for RHUI pipeline

EricEdens:
This migrates the `rhui-release` pipeline from raw yaml to jsonnet, including the wait-for-stable feature from https://github.com/GoogleCloudPlatform/guest-test-infra/pull/451 .

### DIFF
--- a/concourse/pipelines/pipeline-set-pipeline.yaml
+++ b/concourse/pipelines/pipeline-set-pipeline.yaml
@@ -38,7 +38,7 @@ jobs:
   - set_pipeline: container-build
     file: rendered/container-build.json
   - set_pipeline: rhui-release
-    file: rendered/rhui-release.yaml
+    file: rendered/rhui-release.json
   - set_pipeline: artifact-releaser-autopush
     file: guest-test-infra/concourse/pipelines/artifact-releaser-autopush.yaml
   - set_pipeline: partner-image-validations
--- a/concourse/pipelines/pipeline-set-pipeline.yaml
+++ b/concourse/pipelines/pipeline-set-pipeline.yaml
@@ -38,7 +38,7 @@ jobs:
   - set_pipeline: container-build
     file: rendered/container-build.json
   - set_pipeline: rhui-release
-    file: guest-test-infra/concourse/pipelines/rhui-release.yaml
+    file: rendered/rhui-release.yaml
   - set_pipeline: artifact-releaser-autopush
     file: guest-test-infra/concourse/pipelines/artifact-releaser-autopush.yaml
   - set_pipeline: partner-image-validations
--- a/concourse/pipelines/rhui-release.jsonnet
+++ b/concourse/pipelines/rhui-release.jsonnet
@@ -24,7 +24,7 @@ local gcloudmigtask = {
   region:: error 'must set region on gcloudmigtask',
   action:: error 'must set action on gcloudmigtask',
 
-  task: error 'must set task',
+  task: error 'must set task on gcloudmigtask',
   config: {
     image_resource: {
       source: {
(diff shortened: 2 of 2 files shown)


concourse/pipelines/rhui-release.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+    },
+  },
+};
+
+local deployjob = gatejob {
+  local job = self,
+
+  stage:: 'prod',
+  region:: error 'must set region on deployjob',
+  plan: super.plan + std.flattenArrays([
```
EricEdens, line 58: @hopkiw -- I know we talked about avoiding functions from `std` :)  The goal is to repeat the three tasks for each node type (the tasks being update, wait for version, wait for stable), and this was the only way I could think of to achieve that, since as far as I can tell jsonnet's object comprehension only allows a single object to be iterated.

concourse/pipelines/rhui-release.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+
+local gcloudmigtask = {
+  local task = self,
+
+  node:: error 'must set node on gcloudmigtask',
+  stage:: 'prod',
+  region:: error 'must set region on gcloudmigtask',
+  action:: error 'must set action on gcloudmigtask',
+
+  task: error 'must set task',
```
hopkiw, line 27: nit: must set task in gcloudimgtask

concourse/pipelines/rhui-release.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+      type: 'registry-image',
+    },
+    platform: 'linux',
+    run: {
+      args: [
+        'compute',
+        'instance-groups',
+        'managed',
+      ] + task.action + [
+        '--quiet',
```
hopkiw, line 43: `+` will only work if `task.action` is also an array. suggest:

```suggestion
        'compute',
        'instance-groups',
        'managed',
        task.action,
        '--quiet',
```

concourse/pipelines/rhui-release.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+        '%s-mig-%s-%s' % [task.node, task.stage, task.region],
+        '--region=' + task.region,
+        '--project=' + project,
+      ],
+      path: 'gcloud',
+    },
+  },
+};
+
+local deployjob = gatejob {
```
hopkiw, line 53: it doesn't seem like you actually want deployjob to begin on the base of gatejob. it's confusing bc they're very different job types.

concourse/pipelines/rhui-release.jsonnet
hopkiw, line 58: oh, no need for this, we can just write out the plan. i would say pretty much any use of flattenArrays is a code smell for our uses.

concourse/pipelines/rhui-release.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+        project: project,
+        family: nodeType,
+      },
+    }
+    for nodeType in ['cds', 'rhua']
+  ],
+  jobs: [
+    gatejob {
+      name: 'manual-trigger',
+      trigger: false,
```
hopkiw, line 116: this is the default

concourse/pipelines/rhui-release.jsonnet
```diff
@@ -0,0 +1,142 @@
...
+    gatejob {
+      name: 'gate-1',
+      passed: ['deploy-prod-' + region for region in ['europe-west1', 'us-central1']],
+    },
+    deployjob {
+      name: 'deploy-prod-asia-southeast1',
+      region: 'asia-southeast1',
+      passed: ['gate-1'],
+    },
+  ],
```
hopkiw, line 141: i can't read this and tell what the output looks like. for us the use of templates is readability, not absolute reduction in lines of configuration. suggest using a slightly more literal layout

hopkiw (changes requested):

concourse/pipelines/rhui-release.jsonnet
EricEdens: I removed the comprehensions in this version. On the one hand, I agree that it's more readable when we have a small number of regions. I suspect that there will be a lot of copy-and-paste when we add the full set of regions, but we can see what it looks like then.

concourse/pipelines/rhui-release.jsonnet
EricEdens: removed (it was clearer for me as a reader, but I see how it's verbose)

concourse/pipelines/rhui-release.jsonnet
EricEdens: `task.action` _is_ an array :)

concourse/pipelines/rhui-release.jsonnet
EricEdens: done

concourse/pipelines/pipeline-set-pipeline.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
   - set_pipeline: container-build
     file: rendered/container-build.json
   - set_pipeline: rhui-release
-    file: guest-test-infra/concourse/pipelines/rhui-release.yaml
+    file: rendered/rhui-release.yaml
```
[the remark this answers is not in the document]
EricEdens: updated the extension.

I like the idea of having set_pipeline update itself. Let's do that in a separate PR to avoid mixing concerns.

concourse/pipelines/rhui-release.jsonnet
hopkiw: it says done but.. it is not done? (i tried refreshing, lmk if it's on my end)

concourse/pipelines/rhui-release.jsonnet
hopkiw: ok, i see. i thought task.action would always be a single word, but i see it is almost always multiple words.

concourse/pipelines/rhui-release.jsonnet
EricEdens: whoops, sorry, I have it staged locally. Pushed.

hopkiw (approved):

PR #453, opened 2022-04-26
Insert image task

Quintonamore:
Added a new task "gcloud-insert-image" to send a pub/sub message that writes image release information to the ARLE DB. The ARLE autopush pipeline was also modified to test the new task.


concourse/tasks/gcloud-insert-image.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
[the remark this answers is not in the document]
EricEdens: @Quintonamore 

A quick guide to jsonnet:

1. Convert your yaml to JSON (I've been using [yq](https://github.com/mikefarah/yq) recently for that)
2. Paste the json into a file called `gcloud-insert-image.jsonnet`
3. Run `jsonnetfmt gcloud-insert-image.jsonnet`

To reference the task, use `file: rendered/gcloud-insert-image.json`

concourse/tasks/gcloud-insert-image.yaml
```diff
@@ -0,0 +1,22 @@
...
+params:
+  TOPIC: "projects/artifact-releaser-prod/topics/gcp-guest-image-release-prod"
+
+run:
+  path: sh
+  args:
+  - -exc
+  - |
+    wf=$(sed 's/\"/\\"/g' ./compute-image-tools/daisy_workflows/build-publish/((wf)) | tr -d '\n')
+    gcloud pubsub topics publish $TOPIC --message "{\"type\": \"insertImage\", \"request\": {\"image_name\": \"((image_name))\", \"gcs_image_path\": \"((gcs_image_path))\", \"image_publish_template\": \"${wf}\", \"source_version\": \"((source_version))\", \"publish_version\": \"((publish_version))\", \"release_notes\": \"((release_notes))\"}}"
```
EricEdens, line 22: From what I can tell, the purpose of this is to escape a text file into a json string?

I'm concerned about two things:
1. The readability and maintainability of the `sed` approach -- for me, at least, it's really cryptic.
2. The correctness of the escaping.

I'd suggest two alternative approaches: 
1. Write a small container that becomes an interface for communicating with artifact releaser
2. If that's an unjustifiably-large task, use a proper JSON library; I would assume that jq is smart enough to handle this.

concourse/pipelines/artifact-releaser-autopush.jsonnet
```diff
@@ -0,0 +1,133 @@
+local imgpublishjob = {
+  local job = self,
+
```
EricEdens, line 3: It's probably best to move this migration to its own PR  -- I think it's a good practice to have a single PR cover a single issue. "Insert image task" seems like a nice atomic change, and "convert artifact releaser to jsonnet" seems like another nice atomic change.

Quintonamore: Thanks for the review, I forgot to delete this PR after Liam's initial comment but thanks for the pointers for transitioning to jsonnet! I'll separate these out into separate PRs.

PR #459, opened 2022-05-11
Update RHUIv4 with all prod regions

EricEdens:
This was the clearest formulation that I could think of. I tried a couple of others, and they had drawbacks. Let me know if there's another approach that'd be clearer!

Other attempts:

1. Create the jobs without loops. This was a few hundred lines of blocks, and it was hard for me to verify that all of the regions were present.
2. Shuffle the list to randomize regions. With this approach, it was hard for me to verify that all of the regions were represented.

Testing:
- Set the pipeline in my dev environment.

image

### DIFF
--- a/concourse/pipelines/rhui-release.jsonnet
+++ b/concourse/pipelines/rhui-release.jsonnet
@@ -1,43 +1,43 @@
 local project = 'google.com:rhel-infra';
 
-// List of regions where RHUA is deployed.
-local deployed_regions = [
-  'asia-east1',
-  'asia-east2',
+// RHUA is deployed where NFS is available.
+local wave1 = [
   'asia-northeast1',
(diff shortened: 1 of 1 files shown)


concourse/pipelines/rhui-release.jsonnet
```diff
@@ -1,5 +1,44 @@
...
+  'us-east1',
+  'us-east4',
+  'us-west1',
+  'us-west2',
+  'us-west3',
+  'us-west4',
+];
+
+// Generate increasingly-larger waves.
+local wave1 = deployed_regions[0:2];
```
hopkiw, line 37: let's write these explicitly, but everything else looks great

hopkiw (changes requested):

PR #460, opened 2022-05-11
Release testing image to arle prod

Quintonamore:
Previously there were no tasks in the ARLE testing pipeline to test an image publish without publishing to a public cloud project.

Now referencing a static debian-11 build we can use this entry to test in production.

### DIFF
--- a/concourse/pipelines/artifact-releaser-autopush.yaml
+++ b/concourse/pipelines/artifact-releaser-autopush.yaml
@@ -157,6 +157,21 @@ jobs:
           publish_version: ((.:version))
           source_version: "v20211027"
           release_notes: "Disregard this release. Alma Linux 8 test."
+      - task: publish-prod-debian-11
+        config:
+          platform: linux
+          image_resource:
+            type: registry-image
+            source:
+              repository: google/cloud-sdk
+              tag: alpine
+          run:
+            path: /bin/sh
+            args:
+            - -exc
+            - |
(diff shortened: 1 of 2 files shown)


concourse/pipelines/artifact-releaser-autopush.yaml
```diff
@@ -157,6 +157,21 @@ jobs:
...
+            - |
+              wf="{\"Name\": \"debian-prod-test\", {{$work_project := printf \"%q\" \"gce-image-builder\" -}} {{$endpoint := \`\"https://www.googleapis.com/compute/alpha/projects/\"\` -}} {{$delete_after := \`\"24h*30*6\"\` -}} {{if eq .environment \"prod\" -}} \"WorkProject\": {{$work_project}}, \"PublishProject\": \"bct-prod-images\", \"ComputeEndpoint\": {{$endpoint}}, \"DeleteAfter\": {{$delete_after}}, {{- else -}} \"WorkProject\": {{$work_project}}, \"PublishProject\": {{$work_project}}, \"ComputeEndpoint\": {{$endpoint}}, \"DeleteAfter\": {{$delete_after}}, {{- end}} {{$time := trimPrefix .publish_version \"v\"}} \"Images\": [{ \"Prefix\": \"debian-11-bullseye\", \"Family\": \"debian-11-prod-test\", \"Description\": \"Debian, Debian GNU/Linux, 11 (bullseye), amd64 built on {{$time}}, supports Shielded VM features\", \"Licenses\": [ {{if (or (eq .environment \"staging\") (eq .environment \"oslogin-staging\")) -}} \"projects/bct-staging-functional/global/licenses/debian-11-bullseye\" {{- else -}} \"projects/debian-cloud/global/licenses/debian-11-bullseye\" {{- end}} ], \"GuestOsFeatures\": [\"UEFI_COMPATIBLE\", \"VIRTIO_SCSI_MULTIQUEUE\"]}]}"
```
hopkiw, line 173: can we extract this to a file? this is v difficult to read, haha

concourse/pipelines/artifact-releaser-autopush.yaml
```diff
@@ -157,6 +157,21 @@ jobs:
           publish_version: ((.:version))
           source_version: "v20211027"
           release_notes: "Disregard this release. Alma Linux 8 test."
+      - task: publish-prod-debian-11
```
hopkiw, line 160: hm, i don't think it's really publish prod, is it? my understanding of the PR is it is to publish for testing purposes

hopkiw (changes requested):

concourse/pipelines/artifact-releaser-autopush.yaml
Quintonamore: I wanted the workflow saved here to signal that this workflow should not be used anywhere else. Which is why the entire workflow is saved here instead of being read in from `compute-image-tools/daisy_workflows`.

There wasn't an existing directory for saving a file to `guest-test-infra/concourse` so I thought it was best to include directly in the task instead of creating a directory with a single file. If this is saved to a file I don't see a natural place for it, do you have any suggestions?

concourse/pipelines/artifact-releaser-autopush.yaml
Quintonamore: It's to test publish in ARLE's production environment, would `publish-debian-11-prod-env` work better?

concourse/pipelines/artifact-releaser-autopush.yaml
Quintonamore: Talked offline, and decided to save the workflow file to `compute-image-tools`.

https://github.com/GoogleCloudPlatform/compute-image-tools/pull/1923

bkatyl (approved):

PR #471, opened 2022-05-19
[arm] use us-central1-a instead of us-west1-a for guest agent image test for spporting ARM

dntczdx:
use us-central1-a instead of us-west1-a for guest agent image test for spporting ARM

This suppose to be the last change for ARM packages build/publish


dntczdx:
dup with #473

PR #473, opened 2022-05-23
run agent image tests in us-central1

hopkiw:


zmarano:
/lgtm
We may need to alter quota however.

PR #474, opened 2022-05-23
[guest-agent image test] fix: verify that shut down script executed no less than 80s

dntczdx:
The actual allowed running time period is 90s, not 120s. So let's set the threshold to 80s other than 110s.

### DIFF
--- a/imagetest/test_suites/metadata/shutdown_script_test.go
+++ b/imagetest/test_suites/metadata/shutdown_script_test.go
@@ -46,7 +46,7 @@ func TestShutdownUrlScript(t *testing.T) {
 
 // TestShutdownScriptTime test that shutdown scripts can run for around two minutes
 func TestShutdownScriptTime(t *testing.T) {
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/metadata/shutdown_script_test.go
```diff
@@ -45,7 +46,7 @@ func TestShutdownUrlScript(t *testing.T) {
 
 // TestShutdownScriptTime test that shutdown scripts can run for around two minutes
 func TestShutdownScriptTime(t *testing.T) {
-	bytes, err := ioutil.ReadFile("/shutdown.txt")
```
hopkiw, line 48: this path is set in setup.go, so if you want to change it here, you need to change it there as well.

dntczdx: I didn't want to change it. I thought this test forgot to use the contant in setup.go. Just read it again, realized it's different (shutdown_out.txt vs shutdown.txt")

hopkiw (approved):

PR #475, opened 2022-05-23
Strip the word stable from date

patelne:
The tag names are of the form 20220110 or stable-20220110. This change makes it so the `stable-` string is stripped before writing to the file.

### DIFF
--- a/concourse/tasks/get-last-stable-tag.yaml
+++ b/concourse/tasks/get-last-stable-tag.yaml
@@ -21,6 +21,6 @@ run:
     commit=$(cat repo/.git/refs/tags/stable)
     date=$(cd repo/.git/refs/tags; grep -l $commit stable-2*)
     [[ -z $date ]] && exit 1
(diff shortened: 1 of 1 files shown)


hopkiw:
i dont think this is what you want.

on every build of a new commit, we add the version tag, which looks a lot like a date. it is YYYYMMDD.NN, where .NN is a daily incrementing build counter, almost always `.00`. On every release to stable, we add the stable tag (moving with each release) and the stable-YYYYMMDD tag to help us see historically when we've made stable releases. nowhere do we need a simple YYYYMMDD tag.

patelne:
Currently this breaks the `publish-time-since-stable` task since it can't parse the date with the leading `stable-`.

And as far as the workflow goes, the only time `last-stable-tag/date` is used is for the publish-time task. https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/concourse/pipelines/guest-package-build.yaml#L539

hopkiw (approved):

PR #476, opened 2022-05-26
Use templates for guest package build pipeline

hopkiw:
Unlike previous adoption of templates, I wasn't able to guarantee that the output was unchanged. So this PR moves to templates but also fixes some inconsistencies, and establishes a normal process for all builds. The build step also uploads packages, either to unstable or to staging. Then promote steps follow naturally from there. The windows package builds in guest-agent and compute-image-windows repos are separated out by package.

Tested on devcluster to ensure it applies and dependencies are correct, but there is a risk of breakage due to unavoidable size of the change, so I'll test all jobs except for stable promotions after merging.

### DIFF
--- a/concourse/pipelines/guest-package-build.jsonnet
+++ b/concourse/pipelines/guest-package-build.jsonnet
@@ -31,13 +31,13 @@ local publishresulttask = {
   },
 };
 
-// job which builds a package - repos to build and individual upload tasks are passed in
+// job which builds a package - environments to build and individual upload tasks are passed in
 local buildpackagejob = {
   local tl = self,
 
   package:: error 'must set package in buildpackagejob',
   gcs_dir:: tl.package,
-  repos:: error 'must set repos in buildpackagejob',
+  builds:: error 'must set builds in buildpackagejob',
   uploads:: error 'must set uploads in buildpackagejob',
   extra_tasks:: [],
 
@@ -112,13 +112,13 @@ local buildpackagejob = {
       },
     },
     { load_var: 'package-version', file: 'package-version/version' },
-    // Invoke daisy build workflows for all specified repos.
+    // Invoke daisy build workflows for all specified builds.
     {
       in_parallel: {
         fail_fast: true,
         steps: [
           {
-            task: 'guest-package-build-%s-%s' % [tl.package, repo],
+            task: 'guest-package-build-%s-%s' % [tl.package, build],
             config: {
               platform: 'linux',
               image_resource: {
@@ -137,12 +137,12 @@ local buildpackagejob = {
                   '-var:version=((.:package-version))',
                   '-var:gcs_path=gs://gcp-guest-package-uploads/' + tl.gcs_dir,
                   '-var:build_dir=',
-                  'guest-test-infra/packagebuild/workflows/build_%s.wf.json' % underscore(repo),
+                  'guest-test-infra/packagebuild/workflows/build_%s.wf.json' % underscore(build),
                 ],
               },
             },
           }
-          for repo in tl.repos
+          for build in tl.builds
         ],
       },
     },
@@ -250,7 +250,7 @@ local promotepackagejob = {
     { load_var: 'last-stable-date', file: 'last-stable-tag/date' },
     // Run provided promotion tasks.
     { in_parallel: tl.promotions },
-    // Optionally tag the repo. This is optional because some repos produce multiple packages.
+    // Optionally tag the repo. This is optional because some produce multiple packages.
   ] + if tl.tag then [
     {
       put: '%s-tag' % tl.package,
@@ -407,7 +407,7 @@ local buildpackageimagetask = {
   jobs: [
     buildpackagejob {
       package: 'guest-agent',
-      repos: ['deb9', 'deb11-arm64', 'el7', 'el8', 'el9', 'goo'],
+      builds: ['deb9', 'deb11-arm64', 'el7', 'el8', 'el9', 'goo'],
       // The guest agent has additional testing steps to build derivative images then run CIT against them.
       extra_tasks: [
         {
@@ -610,7 +610,7 @@ local buildpackageimagetask = {
     },
     buildpackagejob {
       package: 'guest-oslogin',
-      repos: ['deb9', 'deb10', 'deb11', 'deb11-arm64', 'el7', 'el8', 'el9'],
+      builds: ['deb9', 'deb10', 'deb11', 'deb11-arm64', 'el7', 'el8', 'el9'],
       gcs_dir: 'oslogin',
       uploads: [
         uploadpackagetask {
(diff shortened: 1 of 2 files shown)


concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -0,0 +1,1177 @@
+local underscore(input) = std.strReplace(input, '-', '_');
+
+local buildpackagejob = {
+  local tl = self,
```
EricEdens, line 4: This is a pretty big review :)

For review's sake, I'd feel a lot more comfortable if this were split in two:
1. A port of the previous YAML file
2. Improvements / enhancements

In the PR's overview you mentioned that you weren't able to get the same YAML output as in previous reviews. What kind of challenges were there?

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -27,17 +27,54 @@ local buildpackagejob = {
...
+        image_resource: {
+          type: 'registry-image',
+          source: { repository: 'alpine/git' },
+        },
+
+        inputs: [{ name: tl.package, path: 'repo' }],
+        outputs: [{ name: 'package-version' }],
+
+        run: {
+          path: 'ash',
```
dorileo, line 60: did you mean `bash` or `sh` here?

Linskeyd:
Is it possible to share the fully rendered json here? That could help review as well.

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -164,22 +257,79 @@ local promotepackagejob = {
...
+    config: {
+      platform: 'linux',
+      image_resource: {
+        type: 'registry-image',
+        source: { repository: 'gcr.io/compute-image-tools/daisy' },
+      },
+      run: {
+        path: '/publish-job-result',
+        args: [
+          '--project-id=gcp-guest',
```
dorileo, line 269: Adding to your own comments elsewhere, this block could also become a shell script (it's repeated twice at least, and would reduce maintenance later).

dorileo: I'd say adding a TODO for now would be ok

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -164,22 +257,79 @@ local promotepackagejob = {
...
+
+  task: 'upload-' + tl.repo,
+  config: {
+    platform: 'linux',
+    image_resource: {
+      type: 'registry-image',
+      source: { repository: 'google/cloud-sdk', tag: 'alpine' },
+    },
+    run: {
+      path: 'gcloud',
```
dorileo, line 323: Is there a reason to why some cases you're using full path and others you aren't?

concourse/pipelines/guest-package-build.jsonnet
hopkiw: no this is intended to be 'ash', these args are copied from the original task 
https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/concourse/tasks/generate-package-version.yaml

hopkiw:
> Is it possible to share the fully rendered output here? That could help review as well.

this is available in the presubmit output

concourse/pipelines/guest-package-build.jsonnet
hopkiw: yes just copying from our existing task files

concourse/pipelines/guest-package-build.jsonnet
dorileo: ok

concourse/pipelines/guest-package-build.jsonnet
hopkiw: we had some lengthy discussion offline; my position is it's not practical to make it smaller without adding a lot more work.

in previous translations i was able to guarantee the output was identical because we'd already cleaned up the yaml versions significantly and because the jobs such as image build jobs were naturally more uniform. by contrast, the package build pipeline is more or less the same as the original version we checked in, and package builds have a wide amount of variations - architecture, destination, naming, etc. to make templating work and add value i had to accept that it would also involve changes, such as moving the upload steps into the build jobs, adopting the new promotion RPCs to replace deprecated ones, and unifying some GCS paths. to make it templatized but not make these changes would have been a much more complicated set of templates, which would themselves then go away shortly thereafter. it would more or less double the work involved.

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -0,0 +1,1177 @@
...
+        },
+
+        inputs: [{ name: tl.package, path: 'repo' }],
+        outputs: [{ name: 'package-version' }],
+
+        run: {
+          path: 'ash',
+          args: [
+            '-exc',
+            // Ugly way to produce multi-line script. TODO: maybe move to scripts?
```
Linskeyd, line 63: IMO moving to a script makes sense. Added bonus is then you'd then be able to have automated tests for scripts that get inserted into overall config?

concourse/pipelines/guest-package-build.jsonnet
Linskeyd: +1; every one of these only differs by tl.package. You could make this a construct at the top that gets re-used? Or the construct could be re-used across pipelines (since I expect other pipelines to be very similar in structure needs to the job metric gathering.

hopkiw: this one is a bit different than the ones where i suggested that. those ones have newlines being passed as one big string to a shell interpreter. this one has discrete arguments being passed to a binary. that said, i could extract this job to a template for reuse

concourse/pipelines/guest-package-build.jsonnet
hopkiw: would love to hear your thoughts on how to test scripts. OK to add this as a followup improvement and not in this PR?

concourse/pipelines/guest-package-build.jsonnet
hopkiw: extracted to template!

dorileo: great! thanks.

concourse/pipelines/guest-package-build.jsonnet
Linskeyd: Yeah that is fine to be a follow up; that is what I intended but should have stated.

Not sure if ash will work with bash here, but I found one framework that may be interesting: https://github.com/bats-core/bats-core

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -0,0 +1,1150 @@
...
+        promotepackagestabletask { universe: 'cloud-apt', repo: 'gce-google-compute-engine-oslogin-buster' },
+        promotepackagestabletask { universe: 'cloud-apt', repo: 'gce-google-compute-engine-oslogin-bullseye' },
+        promotepackagestabletask { universe: 'cloud-yum', repo: 'gce-google-compute-engine-oslogin-el7' },
+        promotepackagestabletask { universe: 'cloud-yum', repo: 'gce-google-compute-engine-oslogin-el8' },
+        promotepackagestabletask { universe: 'cloud-yum', repo: 'gce-google-compute-engine-oslogin-el9' },
+      ],
+    },
+    buildpackagejob {
+      package: 'osconfig',
+      repos: ['deb10', 'deb11-arm64', 'el7', 'el8', 'el9', 'goo'],
```
zmarano, line 662: shouldn't this also include deb9 and deb11 as there are packages defined below per version?

hopkiw: good catch. the error is actually in the uploads - we should be uploading the same package to multiple repos. updated.

zmarano: Ah yeah. Perhaps it is a good idea to call out this idea in in the definition of "repos". That sort of represents the build environment for the package more than the repo right?

PR #478, opened 2022-05-31
Add first windows CIT image boot test

patelne:
Adds the initial windows CIT test for testing secure boot.

Also removed unused SingleVMTest function.

```
2022/06/07 13:57:23 running test image_boot/windows-w-new-metadata-scripts (ID 6vdrs) in project pneil-sandbox
2022/06/07 13:57:23 running test image_boot/debian-11 (ID 7l8qz) in project pneil-sandbox
2022/06/07 14:01:12 finished test image_boot/debian-11 (ID 7l8qz) in project pneil-sandbox
2022/06/07 14:03:22 finished test image_boot/windows-w-new-metadata-scripts (ID 6vdrs) in project pneil-sandbox
<testsuites name="" errors="0" failures="0" disabled="0" skipped="0" tests="10" time="0">
        <testsuite name="image_boot-debian-11" tests="5" failures="0" errors="0" disabled="0" skipped="0" time="0">
                <testcase classname="image_boot-debian-11" name="TestGuestBoot" time="0"></testcase>
                <testcase classname="image_boot-debian-11" name="TestGuestReboot" time="0"></testcase>
                <testcase classname="image_boot-debian-11" name="TestGuestRebootOnHost" time="0"></testcase>
                <testcase classname="image_boot-debian-11" name="TestGuestSecureBoot" time="0"></testcase>
                <testcase classname="image_boot-debian-11" name="TestBootTime" time="0"></testcase>
        </testsuite>
        <testsuite name="image_boot-windows-w-new-metadata-scripts" tests="5" failures="0" errors="0" disabled="0" skipped="0" time="0">
                <testcase classname="image_boot-windows-w-new-metadata-scripts" name="TestGuestBoot" time="0"></testcase>
                <testcase classname="image_boot-windows-w-new-metadata-scripts" name="TestGuestReboot" time="0"></testcase>
                <testcase classname="image_boot-windows-w-new-metadata-scripts" name="TestGuestRebootOnHost" time="0"></testcase>
                <testcase classname="image_boot-windows-w-new-metadata-scripts" name="TestGuestSecureBoot" time="18.96"></testcase>
                <testcase classname="image_boot-windows-w-new-metadata-scripts" name="TestBootTime" time="0.01"></testcase>
        </testsuite>
</testsuites>
```

### DIFF
--- a/imagetest/testworkflow.go
+++ b/imagetest/testworkflow.go
@@ -351,11 +351,6 @@ func NewTestWorkflow(name, image, timeout string) (*TestWorkflow, error) {
 	t.wf.Name = strings.ReplaceAll(name, "_", "-")
 	t.wf.DefaultTimeout = timeout
 
-	t.osUnderTest = "linux"
-	if strings.Contains(t.Image, "windows") {
-		t.osUnderTest = "windows"
-	}
-
 	t.wf.DisableGCSLogging()
 	t.wf.DisableCloudLogging()
 	t.wf.DisableStdoutLogging()


patelne:
assign @bkatyl

imagetest/testworkflow.go
```diff
@@ -346,6 +341,7 @@ func NewTestWorkflow(name, image, timeout string) (*TestWorkflow, error) {
 	t.wf = daisy.New()
 	t.wf.Name = strings.ReplaceAll(name, "_", "-")
 	t.wf.DefaultTimeout = timeout
+	t.osUnderTest = osUnderTest
```
patelne, line 344: I also thought about just basing this variable off of the Image name. i.e.
```
	t.osUnderTest = "linux"
	if string.Contains(t.Image, "windows") {
		t.osUnderTest = "windows"
	}
```

This would allow us to avoid adding a new flag for imagetest and keep things a little tighter. But since it is doing string comparison against strings it'll be fairly brittle.

As it is having a new flag would not allow running imagetests against both linux and windows images at the same time.

Thoughts?

imagetest/cmd/manager/main.go
```diff
@@ -36,6 +36,7 @@ var (
 	filter        = flag.String("filter", "", "only run tests matching filter")
 	exclude       = flag.String("exclude", "", "skip tests matching filter")
 	machineType   = flag.String("machine_type", "", "machine type to use for test instances")
+	osUnderTest   = flag.String("os", "linux", "operating system that tests will be run on")
```
bkatyl, line 39: Is there any reason not to use `runtime.GOOS` to determine the OS.

Also instead of passing the value, I'd suggest having the test do the detection as it shouldn't need to be passed.

imagetest/testworkflow.go
hopkiw: i think it should deduce rather than require explicit input where possible. practically we are unlikely to need to test windows and linux at the same time however, so it isn't a hard blocker

imagetest/cmd/manager/main.go
hopkiw, line 39: aside from whether we want to add this flag, i don't think there's any benefit to adding it and not having anything use it. so i don't want to merge this PR until we have some of the implementation added.

hopkiw (changes requested):

imagetest/cmd/manager/main.go
hopkiw: we discussed this offline and i think the better strategy is to resolve compilation errors with OS-specific build constraints but leave test selection as usual.

imagetest/cmd/startup.ps1
```diff
@@ -0,0 +1,52 @@
+# This is temporary until the next windows build happens which will contain the new metadata-scripts package
```
hopkiw, line 1: let's just coordinate them then and not check in temp scripts. you can run your development testing against images which do have the necessary changes

imagetest/cmd/wrapper/main.go
```diff
@@ -65,6 +62,11 @@ func main() {
 	}
 	workDir = workDir + "/"
 
+	testPackage := "image_test"
+	if runtime.GOOS == "windows" {
+		testPackage = "image_test.exe"
```
hopkiw, line 67: can we make these values passed in via metadata?

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -24,6 +25,10 @@ func TestGuestBoot(t *testing.T) {
 }
 
 func TestGuestReboot(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Reboot not implemented on Windows")
```
hopkiw, line 29: why is reboot not implemented on windows?

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -54,6 +63,14 @@ func TestGuestRebootOnHost(t *testing.T) {
 }
 
 func TestGuestSecureBoot(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		testWindowsGuestSecureBoot(t)
```
hopkiw, line 67: as a rule we shouldn't pass `t` around, but use the usual error paradigm and have error handling in this function

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -76,6 +93,23 @@ func TestGuestSecureBoot(t *testing.T) {
...
+	output, err := cmd.Output(); 
+	
+	if err != nil {
+		t.Fatalf("failed to run SecureBoot command: %v", err)
+	}
+
+	// The output will return a string that is either 'True' or 'False'
+	// so we need to parse it and compare here.
+	trimmed_output := strings.TrimSpace(string(output))
+	if trimmed_output != "True" {
```
hopkiw, line 107: lines can be combined

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -76,6 +93,23 @@ func TestGuestSecureBoot(t *testing.T) {
...
+		t.Fatalf("Secure boot is not enabled as expected: %v", trimmed_output)
```
hopkiw, line 108: won't this be "Secure boot is not enabled as expected: False" in all cases? is it useful to include?

imagetest/testworkflow.go
```diff
@@ -283,8 +282,14 @@ func finalizeWorkflows(ctx context.Context, tests []*TestWorkflow, zone, gcsPref
...
-		twf.wf.Sources["wrapper"] = fmt.Sprintf("%s.%s", testWrapperPath, arch)
-		twf.wf.Sources["testpackage"] = fmt.Sprintf("/%s.%s.test", twf.Name, arch)
+		if twf.osUnderTest == "windows" {
+			twf.wf.Sources["testpackage"] = fmt.Sprintf("/%s.%s.test.exe", twf.Name, arch)
+			twf.wf.Sources["wrapper.exe"] = "/wrapper.exe"
+			twf.wf.Sources["startup.ps1"] = "/startup.ps1"
+		} else {
+			twf.wf.Sources["testpackage"] = fmt.Sprintf("/%s.%s.test", twf.Name, arch)
+			twf.wf.Sources["wrapper"] = fmt.Sprintf("%s.%s", testWrapperPath, arch)
+		}
```
hopkiw, line 292: this is fine but it may be a bit more readable to have a paradigm that just uses the string formatter as the decider, you can do that like:

```go
var suffix string
if twf.osUnderTest == "windows" {
    suffix = ".exe"
}
twf.wf.Sources["testpackage"] = fmt.Sprintf("/%s.%s.test%s", twd.Name, arch, exe)
```

main benefit is reducing complexity of reading. rather than two whole blocks for different OSes, OS is simply considered in the formatting step. just a suggestion

imagetest/testworkflow.go
```diff
@@ -347,6 +352,11 @@ func NewTestWorkflow(name, image, timeout string) (*TestWorkflow, error) {
 	t.wf.Name = strings.ReplaceAll(name, "_", "-")
 	t.wf.DefaultTimeout = timeout
 
+	t.osUnderTest = "linux"
+	if strings.Contains(t.Image, "windows") {
+		t.osUnderTest = "windows"
+	}
```
hopkiw, line 358: imo unless there are more complex steps, i think we should just use this logic directly wherever we are currently checking `t.osUndertest`

hopkiw (changes requested):

imagetest/test_suites/image_boot/image_boot_test.go
patelne: Ah, you're right. Fixed.

imagetest/testworkflow.go
patelne: That's fair. Don't have a strong opinion on this. We can come back to this if for some reason OS classification becomes more complex.

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: oh ok for tests which don't work on a platform yet, no need to add these permanent skips. better just in your testing to specify what tests you *do* want to run with combination of `-filter` and -`exclude`, wdyt?

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: yeah if you want to keep the functions split up as written, i suggest they don't take any args but return error, then this function simply uses the `if err != nil` paradigm and all the use of the `t` object remains in this function

imagetest/test_suites/image_boot/image_boot_test.go
patelne: Those flags only work on the whole package name right? I don't see a way to do that a per-test level within a package

imagetest/testworkflow.go
patelne: Went ahead and did this as it does make it much cleaner.

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: i see, i thought you were doing this for all tests in the 'suite'. what are the blockers to finishing the windows tests in this suite before merging? i'd rather not check in short-lived temporary skips if we can avoid it

patelne: Actually I was able to do the whole suite.

patelne:
Bumping this in case it got missed :)

hopkiw (approved):

PR #488, opened 2022-06-03
extract generate-timestamp to template

hopkiw:
there was a typo in the version used in promote jobs, so i've extracted the version from build jobs and used it as a template.


wrigri (approved):
/lgtm

PR #489, opened 2022-06-03
Fix last stable

hopkiw:
* move prep steps for stable tag into the optional block
* disable the last-stable-date stuff which is proving too fragile and needs redesign
* only let one of the resulting promotions tag stable since it's repo-specific not package-specific


patelne (approved):

wrigri (approved):
/lgtm

PR #522, opened 2022-07-13
build & publish windows-install-media 

dntczdx:
hold until the corresponding build files on compute-image-tools are merged.

### DIFF
--- a/concourse/pipelines/windows-image-build.jsonnet
+++ b/concourse/pipelines/windows-image-build.jsonnet
@@ -7,7 +7,7 @@ local gcp_secret_manager = import '../templates/gcp-secret-manager.libsonnet';
 local client_envs = ['testing', 'staging', 'internal'];
 local server_envs = ['testing', 'staging', 'internal', 'prod'];
 local sql_envs = ['testing', 'staging', 'prod'];
-local windows_install_media_envs = ['testing', 'staging', 'prod'];
+local windows_install_media_envs = ['testing', 'prod'];
 local underscore(input) = std.strReplace(input, '-', '_');
 
 // Templates.
@@ -355,7 +355,7 @@ local windowsinstallmediaimgbuildjob = {
     },
     {
       load_var: 'iso_path_2022',
-      file: 'gcp-secret-manager/' + 'win2022-64',
+      file: 'gcp-secret-manager/win2022-64',
     },
     {
       task: 'get-secret-iso-path-2019',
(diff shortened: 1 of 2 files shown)


concourse/pipelines/windows-image-build.jsonnet
```diff
@@ -7,6 +7,7 @@ local gcp_secret_manager = import '../templates/gcp-secret-manager.libsonnet';
 local client_envs = ['testing', 'staging', 'internal'];
 local server_envs = ['testing', 'staging', 'internal', 'prod'];
 local sql_envs = ['testing', 'staging', 'prod'];
+local windows_install_media_envs = ['testing', 'staging', 'prod'];
```
bkatyl, line 10: https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/build-publish/windows/windows-install-media.publish.json

Only includes testing and prod.

concourse/pipelines/windows-image-build.jsonnet
```diff
@@ -298,6 +299,138 @@ local containerimgbuildjob = {
...
+      load_var: 'gcs-url',
+      file: '%s-gcs/url' % job.image,
+    },
+    {
+      task: 'get-secret-iso-path-2022',
+      config: gcp_secret_manager.getsecrettask { secret_name: 'win2022-64' },
+    },
+    {
+      load_var: 'iso_path_2022',
+      file: 'gcp-secret-manager/' + 'win2022-64',
```
bkatyl, line 358: For these variables, they can be a single string instead of two string that are concatenated.

https://github.com/GoogleCloudPlatform/guest-test-infra/blob/1e1879aa7dbcde153218a1f6f80d0d2c1e3dbcf6/concourse/pipelines/windows-image-build.jsonnet#L97 is an example where the rest of this file uses a single string.

'gcp-secret-manager/' + 'win2022-64' --> 'gcp-secret-manager/win2022-64'

Same for the other secret manager variables when it's a set string.

concourse/pipelines/windows-image-build.jsonnet
dntczdx: done

concourse/pipelines/windows-image-build.jsonnet
dntczdx: done

bkatyl (approved):

PR #530, opened 2022-07-20
Test for windows support of MTU 1500 from previous 1460

KarnVadaliya:


KarnVadaliya:
Since changing this test would likely break all of the linux CIT tests so closing the PR.

PR #540, opened 2022-08-08
Add SQL Server 2022 Preview to Windows pipeline

jjerger:
Adds derivative build for SQL 2022 running on Windows Server 2022. Also had to add support for passing in SSMS version, as Microsoft moved away from a universal installer. We may need to pass different versions going forward.


jjerger:
I'm going to acknowledge the mix of "prerelease" and hardcoding sqlserver is confusing here. My intent is to get Concourse building SQL 22 Preview successfully while I figure out a better way to approach this type of scenario in the future.

bkatyl (approved):

PR #543, opened 2022-08-19
Boot time regression

KarnVadaliya:
Boot time regression test for linux


imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -123,5 +197,49 @@ func TestBootTime(t *testing.T) {
...
+
+	MAX_THRESHOLD := getThresholdValue(image)
+
+	if MAX_THRESHOLD == 0 {
+		t.Fatalf("unrecognized image, no threshold value found")
+	}
+
+	var foundGuestAgentAndSshd bool
+
+	for i := 0; i < 120; i++ {
```
dorileo, line 217: Where's the value 120 coming from? is this the number of retries? if so why 120?

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -123,5 +197,49 @@ func TestBootTime(t *testing.T) {
...
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	if !foundGuestAgentAndSshd {
+		t.Fatalf("Condition for guest agent and sshd process to start not reached within timeout")
+	}
+
+	// Reading the system uptime once both guest agent and sshd are found in the processes
+	uptimefile, err := os.ReadFile("/proc/uptime")
```
dorileo, line 230: uptimeContent or uptimeData seems to be a naming better reflecting its purpose for this variable (instead of uptimefile)

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -123,5 +197,49 @@ func TestBootTime(t *testing.T) {
...
+	t.Logf("image start time is %d", time.Now().Second()-startTime)
+}
+
+func TestBootTime(t *testing.T) {
+	image, err := utils.GetMetadata("image")
+	if err != nil {
+		t.Fatalf("couldn't get image from metadata")
+	}
+
+	MAX_THRESHOLD := getThresholdValue(image)
```
dorileo, line 209: why not maxThreashold instead of MAX_THREASHOLD?

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -123,5 +197,49 @@ func TestBootTime(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to convet start time %s", metadata)
 	}
-	t.Logf("image boot time is %d", time.Now().Second()-startTime)
+	t.Logf("image start time is %d", time.Now().Second()-startTime)
+}
+
+func TestBootTime(t *testing.T) {
+	image, err := utils.GetMetadata("image")
```
dorileo, line 204: What's is this utils.GetMetadata("image") returning? the image name? can you make it return os-family? if so getThresholdValue() can be simplified or wouldn't be required at all.

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -17,11 +18,84 @@ import (
...
+	sort.Sort(sort.Reverse(sort.StringSlice(names)))
+	for _, name := range names {
+		if strings.Contains(image, name) {
+			return imageFamilyBootTimeThresholdMap[name]
+		}
+	}
+	return 0
+}
+
+func lookForProcesses() error {
```
dorileo, line 52: Dropping some ideas, why instead of looking for the process you don't look for the systemd service? that way you can detect if the service has failed or not - distinguishing a service failure from a time regression.

Imagine the case when the ssh service has failed to start (for any reason) and you're testing it, the process will never show up you will assume that ssh is taking too long but in fact the service is broken, if you test using the systemd service you will be aware of the service status and distinguish between time regression and the service failure, we'll at least have a better view of the real problem.

dorileo (commented):
The implementation overall looks ok, I have some considerations to some decisions and dropped some ideas.

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: no, it cannot return the family. this is the standard paradigm for this kind of check, see other tests

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: i advised him to look at processes and not systemd service. we have a timeout on the check.

dorileo: How would the timeout impact using or not using/relying on systemd? One could even use systemd-analyze to determine every piece of the boot process time, then use blame to factor the time of the unit you desire, you can get much more accurate data from that.

imagetest/test_suites/image_boot/image_boot_test.go
KarnVadaliya: This is based on the highest value of the all thresholds so the maximum of all is 120 so the value is set to 120 here.

imagetest/test_suites/image_boot/image_boot_test.go
KarnVadaliya: Done

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -17,11 +18,84 @@ import (
...
+	"ubuntu":      75,
+	"ubuntu-pro":  110,
+}
+
 const (
 	markerFile     = "/boot-marker"
 	secureBootFile = "/sys/firmware/efi/efivars/SecureBoot-8be4df61-93ca-11d2-aa0d-00e098032b8c"
 )
 
+func getThresholdValue(image string) float64 {
```
hopkiw, line 38: i would rather we use explicit logic here as in other tests, rather than assuming we will always be able to create a name map that can return data using this ad hoc algorithm. this approach can easily introduce subtle bugs.

explicit logic means a switch/case statement checking images that we know we have support for.

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw, line 52: this function is pretty specific to the conditions of the new test - it checks for sshd and guest agent both being up. i suggest either extracting the process listing logic to its own function or at least renaming this function to indicate what it is doing.

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: can you add a comment explaining this reasoning?

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -123,5 +197,49 @@ func TestBootTime(t *testing.T) {
...
+	uptimeData, err := os.ReadFile("/proc/uptime")
+	if err != nil {
+		t.Fatalf("Failed to read uptime file")
+	}
+	fields := strings.Split(string(uptimeData), " ")
+	uptime, err := strconv.ParseFloat(fields[0], 64)
+	if err != nil {
+		t.Fatalf("Failed to read uptime numeric value")
+	}
+	t.Logf("The boot time is %v seconds", uptime)
```
hopkiw, line 239: i don't think we need logf in tests; the result is discarded. if you want this for troubleshooting you should probably be more specific than 'boot time', maybe 'found guest agent and sshd running at %v seconds'

imagetest/test_suites/image_boot/setup.go
```diff
@@ -43,12 +49,12 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		return nil
 	}
 
-	vm3, err := t.CreateTestVM("vm3")
+	vm4, err := t.CreateTestVM("vm4")
```
hopkiw, line 52: is there some reason we need a completely separate VM just for this test?

imagetest/test_suites/image_boot/setup.go
```diff
@@ -43,12 +49,12 @@ func TestSetup(t *imagetest.TestWorkflow) error {
...
 	if err != nil {
 		return err
 	}
-	vm3.AddMetadata("start-time", strconv.Itoa(time.Now().Second()))
-	vm3.EnableSecureBoot()
-	vm3.RunTests("TestGuestSecureBoot|TestBootTime")
+	vm4.AddMetadata("start-time", strconv.Itoa(time.Now().Second()))
+	vm4.EnableSecureBoot()
```
hopkiw, line 57: does this test need secure boot?

hopkiw (changes requested):

imagetest/test_suites/image_boot/image_boot_test.go
KarnVadaliya: So the value 120 is based on the maximum of all thresholds of all images so from the list it is sles-15. This is because we are taking into consideration the worst case for both sshd and guest agent to start for all images. The values are decided by taking enough samples by running the test multiple times. I can update the iteration to run based on each image's expected threshold value.  Also, the lookForProcesses() function it will return an error if both sshd and guest-agent are not found in the process filesystem in that case we wait for a delay of one second and check again giving the process the time to start. Currently we check for it 120 times (120 seconds + time to check (which is minimal in so ignoring)). So if for example for sles-15 image if we not able to find sshd and guest-agent in 120s that means it is taking longer than expected for these processes to starts and violates our expected boot-time conditions so the test fails

hopkiw: i do understand :) i suggest a comment on line 216 saying '120 is the current maximum number of seconds of any distro' or similar, so anyone reading the code understands where the number 120 came from

imagetest/test_suites/image_boot/setup.go
KarnVadaliya: To isolate this test to run separately not sure how other other test would effect the boot time.

imagetest/test_suites/image_boot/setup.go
KarnVadaliya: Well adding secure boot does not affect the boot time

imagetest/test_suites/image_boot/setup.go
KarnVadaliya: Added the test in the same VM as secure boot as it does not affect the boot time

imagetest/test_suites/image_boot/image_boot_test.go
hopkiw: i'm sorry, i somehow resolved this by accident. i think using systemd-analyze as you suggest would be useful for per-unit determinations, but we only care about the final goal state, a single measurement. and we want to be reliable to errors or misconfigurations in the systemd space. so talking to systemd is not appropriate for this particular test

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -17,11 +17,94 @@ import (
...
+	case strings.Contains(image, "sles-12"):
+		return imageFamilyBootTimeThresholdMap["sles-12"]
+	case strings.Contains(image, "sles-15"):
+		return imageFamilyBootTimeThresholdMap["sles-15"]
+	case strings.Contains(image, "ubuntu-pro"):
+		return imageFamilyBootTimeThresholdMap["ubuntu-pro"]
+	case strings.Contains(image, "ubuntu"):
+		return imageFamilyBootTimeThresholdMap["ubuntu"]
+	default:
+		return 0
```
hopkiw, line 57: i think the default should be an explicit error that this is an unsupported OS, or a reasonable default. it might mean moving the case statement into the main function

hopkiw (changes requested):

PR #544, opened 2022-08-25
Add Windows GooGet tests.

wrigri:


imagetest/cmd/manager/main.go
```diff
@@ -84,6 +85,14 @@ var (
 		"ubuntu-pro-1604-lts":     "projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-1604-lts",
 		"ubuntu-pro-1804-lts":     "projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-1804-lts",
 		"ubuntu-pro-2004-lts":     "projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-2004-lts",
+		"windows-2012-r2-core":    "projects/windows-cloud/global/images/family/windows-2012-r2-core",
```
dorileo, line 88: I'd love to see alphabetically ordered lists/lines :-)

imagetest/utils/test_utils.go
```diff
@@ -238,3 +242,83 @@ func GetInterface(index int) (net.Interface, error) {
...
+		t.Fatalf("couldn't get image from metadata")
+	}
+
+	if strings.Contains(image, "-x86") {
+		t.Skip(skipMsg)
+	}
+}
+
+// PowershellOutput holds stdout, stderr and the exit code from running a powershell command.
+type PowershellOutput struct {
```
dorileo, line 266: ProcessStatus would sound better than PoweshellOutput and would embrace future uses avoiding possible duplications

dorileo (commented):
PR looking good, just a minor comment - more a rant though :-).

imagetest/cmd/manager/main.go
wrigri: I can switch the -core files around on my next PR. Interestingly, the order here is just the order that came out when I did "gcloud images list"

The reason that they come out in that order is that they are ordered by image name and not by family, so
windows-server-2016-dc-core-v20220812
sorts before
windows-server-2016-dc-v20220812

PR #552, opened 2022-09-19
Skipping boot test for Alma Linux

KarnVadaliya:


imagetest/test_suites/image_boot/setup.go
```diff
@@ -38,6 +38,11 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		return nil
 	}
 
+	if strings.Contains(t.Image, "almalinux") {
+		// secure boot is not supported on Alma Linux
```
zmarano, line 42: Secure boot is supported both on Rocky Linux and AlmaLinux.... but is this test even about secure boot?

KarnVadaliya: Both the test (secure boot and boot time) runs on the same VM so skipping for both. Should I run them on separate VMs ?

zmarano: Oh I see.... we don't need to skip in either case- just add almalinux to the supported boot time tests.

KarnVadaliya: Also, skipping secure boot for rocky-linux-8 was already there before. As of now it is supported ?

zmarano: Yeah- it is supported so we should actually remove that skip (I thought I did already).

PR #554, opened 2022-09-20
Remove staging publication (except oslogin staging).

zmarano:


concourse/pipelines/windows-image-build.jsonnet
```diff
@@ -456,8 +456,8 @@ local imgpublishjob = {
       passed: [
         job.passed,
       ],
-      // Builds are automatically pushed to testing. Triggering staging will automatically progress to prod and internal.
-      trigger: if job.env == 'staging' then false else true,
+      // Builds are automatically pushed to testing.
```
bkatyl, line 459: We need it to stop after testing. Staging was useful because it was the common step between testing and "the rest" internal/internal, prod etc.

I'm not sure what the right condition for this is, but I'll think on it tonight.

zmarano: Yeah.... I was thinking about that- internal maybe?

bkatyl: sql_envs and windows_install_media_envs don't use the internal project. But maybe best approach is to create internal project(s) for them so it's a predictable pattern.

zmarano: How about this.

concourse/pipelines/windows-image-build.jsonnet
```diff
@@ -456,8 +456,8 @@ local imgpublishjob = {
...
+      trigger: if job.env == 'testing' then true else false,
```
zmarano, line 460: @bkatyl This is what we do for Linux where only testing is automated (after build).

zmarano: Altered this to be:
build -> auto publish to testing
trigger publish to prod -> auto publish to internal prod

bkatyl (approved):

PR #555, opened 2022-09-20
Add driver packages tests.

wrigri:


imagetest/test_suites/windows/driver_package_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+func getDriverList(remove bool) []string {
+	drivers := []string{
+		"google-compute-engine-driver-netkvm",
```
bkatyl, line 16: This is the Network VirtIO driver, so I believe you need to treat it like the gvnic driver were it doesn't get uninstalled.

imagetest/test_suites/windows/driver_package_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func TestNetworkDriverLoaded(t *testing.T) {
+	utils.WindowsOnly(t)
+	command := fmt.Sprintf("ipconfig /all | Select-String Description")
+	output, err := utils.RunPowershellCmd(command)
+	if err != nil {
+		t.Fatalf("Error running 'ipconfig': %v", err)
+	}
+	adapterName := "Google VirtIO Ethernet Adapter"
```
bkatyl, line 53: This description is good for the virtio case.

GVNIC has a slightly different description. It's "Google Ethernet Adapter"

We'll likely need to test virtio and gvnic in different instances since an instance can only have one or another.

imagetest/test_suites/windows/driver_package_test.go
wrigri: Done.

bkatyl (approved):

PR #558, opened 2022-10-04
add new OWNERS file

hopkiw:
for the imagetest/ dir, currently only one team member with relevant domain expertise, so let's mandate approval for now.

### DIFF
--- a/imagetest/OWNERS
+++ b/imagetest/OWNERS
@@ -1,0 +1,2 @@
+approvers:
+  - hopkiw


imagetest/OWNERS
```diff
@@ -0,0 +1,2 @@
+approvers:
+  - hopkiw
```
bkatyl, line 2: I strongly disagree with a single approver. This is a resource owned by the team.

hopkiw: this is just my first thought of how to ensure I approve all PRs in this specific subdirectory. If I were unavailable, you'd still be able to remove this OWNERS file if needed to bypass.

Do you have an alternative suggestion on how to accomplish this?

wrigri: I wasn't sure how I felt about this, but now that I've thought about it, I don't think this is necessary. While I understand wanting to review all of the changes in this subdirectory, the downsides of a not-quite-fully-reviewed PR are not as high as with some of the other code that we own. If a PR gets merged in and there were changes that would have been useful to make the code clearer or more correct, we can just send that in another PR. We don't need to create a whole new release if there is some minor change that comes in a separate PR.

It would perhaps be useful to have additional automated checks in this repo/subdir, but that's a separate discussion.

PR #559, opened 2022-10-05
Update Metadata tests to support Windows.

wrigri:
Also, refactor/consolidate tests to run on fewer VMs.


imagetest/fixtures.go
```diff
@@ -126,41 +126,36 @@ func (t *TestVM) AddMetadata(key, value string) {
...
+	t.testWorkflow.wf.Sources[sourceFile] = fileName
 
-	t.AddMetadata("shutdown-script-url", "${SOURCESPATH}/shutdown-script")
+	sourcesPath := "${SOURCESPATH}/" + sourceFile
+	t.AddMetadata(metadataKey, sourcesPath)
 	return nil
 }
 
-// SetStartupScript sets the `startup-script` metadata key for a VM.
-func (t *TestVM) SetStartupScript(script string) {
```
hopkiw, line 160: please restore the helper functions, which keep the constants in the library package rather than in each test

imagetest/fixtures.go
```diff
@@ -126,41 +126,36 @@ func (t *TestVM) AddMetadata(key, value string) {
...
-	t.AddMetadata("shutdown-script", script)
-}
-
-// SetShutdownScriptURL sets the`shutdown-script-url` metadata key for a VM.
-func (t *TestVM) SetShutdownScriptURL(script string) error {
-	fileName := fmt.Sprintf("/shutdown_script-%s", uuid.New())
+// SetMetadataScriptURL sets a URL metadata key for a VM. For Windows VMs,
+// the filename must have an extention which should be supplied with the
+// dot (e.g. ".ps1"). Otherwise, 'ext' can be an empty string.
+func (t *TestVM) SetMetadataScriptURL(metadataKey, script, ext string) error {
```
hopkiw, line 146: remove the 'ext' parameter, since you're not validating it anyway

imagetest/test_suites/metadata/setup.go
```diff
@@ -12,40 +12,189 @@ var Name = "metadata"
 
 const (
 	daemonScriptTemplate = `#!/bin/bash
-
-nohup sleep 3600 > /dev/null 2>&1 < /dev/null &
-echo $! > %s
-`
+	nohup sleep 3600 > /dev/null 2>&1 < /dev/null &
```
hopkiw, line 15: this looks like some reformatting your IDE did, but fyi it means the script will have this indentation. probably you want to restore the previous un-indented string literal for every script you're using backtics for in this file.

imagetest/test_suites/metadata/setup.go
```diff
@@ -12,40 +12,189 @@ var Name = "metadata"
...
+	ping -n 2 127.0.0.1 >NUL`
+	powershellScriptTemplate = `Set-Content -Value "%s" -Path %s
+	Start-Sleep 1`
+	startupContent  = "The startup script worked."
+	shutdownContent = "The shutdown script worked."
+
+	timeOutputPath        = "/shutdown.txt"
+	windowsTimeOutputPath = "C:\\shutdown.txt"
+
+	bashTimeScriptTemplate = `#!/bin/bash
```
hopkiw, line 32: this naming is more generic than 'shutdownScriptTime', and not clear why you've made it a template since it's only used once

imagetest/test_suites/metadata/setup.go
```diff
@@ -12,40 +12,189 @@ var Name = "metadata"
...
+	while [[ 1 ]]; do
+	date +%%s >> %s
+	  sync
+	  sleep 1
+	  done`
+
+	windowsTimeScriptTemplate = `while ($true) {
+		$time = (Get-Date).ToString("yyyyMMdd-HHMMss")
+		Add-Content -Path %s -Value $time
```
hopkiw, line 41: this is only used in one place, why make this a pass-in value? the naming for this script should also indicate it is specific to shutdown scripts

imagetest/test_suites/metadata/setup.go
```diff
@@ -12,40 +12,189 @@ var Name = "metadata"
 
 const (
 	daemonScriptTemplate = `#!/bin/bash
-
-nohup sleep 3600 > /dev/null 2>&1 < /dev/null &
-echo $! > %s
-`
+	nohup sleep 3600 > /dev/null 2>&1 < /dev/null &
+	echo $! > %s
```
hopkiw, line 16: this script is only used once, why template it?

imagetest/test_suites/metadata/setup.go
```diff
@@ -12,40 +12,189 @@ var Name = "metadata"
...
+
+func (ms metadataScript) script() string {
+	return fmt.Sprintf(ms.scriptTemplate, ms.outputContent, ms.outputPath)
+}
 
 // TestSetup sets up the test workflow.
 func TestSetup(t *imagetest.TestWorkflow) error {
+	daemonScript := fmt.Sprintf(daemonScriptTemplate, daemonOutputPath)
+	shutdownTimeScript := fmt.Sprintf(bashTimeScriptTemplate, timeOutputPath)
+	daemonMetadataKey := "startup-script"
```
hopkiw, line 191: variable is only used once, could omit

imagetest/test_suites/metadata/setup.go
```diff
@@ -56,63 +205,54 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 	if err != nil {
 		return err
 	}
-	vm2.SetShutdownScript(shutdownScript)
+	for _, ms := range metadataScripts {
+		if ms.url {
+			vm2.SetMetadataScriptURL(ms.metadataKey, ms.script(), ms.extension)
+		} else {
+			vm2.AddMetadata(ms.metadataKey, ms.script())
+		}
```
hopkiw, line 213: if you're already putting logic in method form you could just put this logic in the script() method (and omit extension param as mentioned above)

imagetest/test_suites/metadata/setup.go
```diff
@@ -56,63 +205,54 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 	if err != nil {
 		return err
 	}
-	vm2.SetShutdownScript(shutdownScript)
+	for _, ms := range metadataScripts {
```
hopkiw, line 208: the slice has a lot of keys for linux and windows, is it safe to add them all to a given vm? are you relying on some overwrite behavior, ordering, or client filtering?

imagetest/test_suites/metadata/setup.go
```diff
@@ -12,40 +12,189 @@ var Name = "metadata"
...
+	scriptTemplate string
+	outputPath     string
+	outputContent  string
+	extension      string
+}
+
+// A list of all metadata script types. Windows scripts should be listed
+// in the expected run order. Startup URL scripts are excluded because
+// they are used by the test runner and implicitly tested.
+var metadataScripts = []metadataScript{
```
hopkiw, line 66: you might want to put these in 2 top level structures and use a map[string]metadataScript with the key name as the map key. then you could choose to iterate over key names, as you're using this a few times below.

imagetest/test_suites/metadata/shutdown_script_test.go
```diff
@@ -1,58 +0,0 @@
-// +build cit
```
hopkiw, line 1: why deleted?

imagetest/test_suites/metadata/startup_shutdown_script_test.go
```diff
@@ -0,0 +1,127 @@
...
+const shutdownTime = 80
+
+// TestStartupShutdownScripts tests that all startup/shutdown scripts
+// appropriate for the platform ran successfully.
+func TestStartupShutdownScripts(t *testing.T) {
+	isWindows := runtime.GOOS == "windows"
+	for _, ms := range metadataScripts {
+		if ms.windows != isWindows {
+			continue
+		}
```
hopkiw, line 26: if you accept my suggestion for maps this would obviate the need for this per-key check, instead you would decide which map to iterate over

imagetest/test_suites/metadata/startup_shutdown_script_test.go
```diff
@@ -0,0 +1,127 @@
...
+			t.Fatalf("Couldn't get %s from metadata: %v", script.metadataKey, err)
+		}
+	}
+}
+
+// TestWindowsScriptOrder tests that Windows startup & shutdown scripts are
+// executed in the documented order.
+func TestWindowsScriptOrder(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("Script run order only checked on Windows.")
```
hopkiw, line 53: why don't we check order on linux? we have a defined order, just fewer keys

imagetest/test_suites/metadata/startup_shutdown_script_test.go
```diff
@@ -0,0 +1,127 @@
...
+	outputSlice := strings.Split(output.Stdout, "\n")
+
+	for _, outputFile := range outputSlice {
+		outputFile = strings.TrimSpace(outputFile)
+		if strings.HasPrefix(outputFile, "shutdown") {
+			actualShutdownOrder = append(actualShutdownOrder, outputFile)
+		} else if strings.HasPrefix(outputFile, "startup") {
+			actualStartupOrder = append(actualStartupOrder, outputFile)
+		} else if strings.HasPrefix(outputFile, "sysprep") {
+			actualStartupOrder = append(actualStartupOrder, outputFile)
```
hopkiw, line 84: most cases of repeated if/elseif in go are best written using switch/case

hopkiw (changes requested):

imagetest/test_suites/metadata/setup.go
wrigri: Yes, it's safe to add all of them to a given VM. The non-applicable metadata keys get ignored by the VM and if adding something like "windows-starutp-script-ps1" caused problems on Linux, I think we'd want to know about that.

imagetest/fixtures.go
wrigri: An extension is required for Windows URL startup scripts. This is not about validation, it's about creating a working startup script. If you set a metadata key like this:
windows-startup-script-url = "gs://mybucket/mystartupscript"
it will fail.

You must have a value like this:
windows-startup-script-url = "gs://mybucket/mystartupscript.cmd"

imagetest/test_suites/metadata/setup.go
wrigri: No, this was intentional. The indentation in the scripts don't cause any problems, and the un-indented strings were significantly harder to read. It was a judgement call of code readability over perfectly formatted scripts in metadata.

imagetest/test_suites/metadata/shutdown_script_test.go
wrigri: startup_script_test.go and shutdown_script_test.go were combined into startup_shutdown_script_test.go since most of the tests that were previously separate use the same logic for testing both the startup and shutdown scripts.

imagetest/test_suites/metadata/setup.go
wrigri: The template isn't used more than once, but the path variables are used more than once.

imagetest/test_suites/metadata/setup.go
wrigri: The script is only used once, but the output path value is used more than once.

imagetest/test_suites/metadata/setup.go
wrigri: Done

imagetest/test_suites/metadata/startup_shutdown_script_test.go
wrigri: Done.

imagetest/test_suites/metadata/setup.go
wrigri: Like the previous one, while the script itself is only used once, the path value is used more than once.

imagetest/test_suites/metadata/setup.go
wrigri: OK, I've looked at this for a bit and I don't see a way to put this logic into the script() method. AddMetadata and SetMatadataScriptURL do very different things, even though the values passed to them are similar.

imagetest/fixtures.go
wrigri: OK, I updated this to just automatically use ".ps1" for Windows URL scripts.

imagetest/test_suites/metadata/setup.go
wrigri: Done

imagetest/test_suites/metadata/startup_shutdown_script_test.go
wrigri: Done

bkatyl (approved):

PR #568, opened 2022-11-14
GVNIC test

KarnVadaliya:


imagetest/test_suites/network/gvnic_test.go
```diff
@@ -0,0 +1,54 @@
...
+	file := fmt.Sprintf(driverInfoPath, interfaceName)
+	_, err := os.Stat(file)
+	if os.IsNotExist(err) {
+		return err
+	}
+	fileData, err := os.ReadFile(file)
+	if err != nil {
+		return err
+	}
+	driver := driverRegex.FindString(string(fileData))
```
hopkiw, line 35: this is a very interesting approach, but i think it is easier to do using link resolution. the file /sys/class/net/$device/device/driver/module will be a symbolic link to the driver - this can be looked up using `os.Readlink()` without ever reading any file

imagetest/test_suites/network/setup.go
```diff
@@ -73,7 +73,17 @@ func TestSetup(t *imagetest.TestWorkflow) error {
...
+		return err
+	}
+	vm3.ChangeNicTypeToGVNIC()
+	if err := vm3.Reboot(); err != nil {
+		return err
+	}
+
 	vm1.RunTests("TestPingVMToVM|TestDHCP|TestDefaultMTU")
 	vm2.RunTests("TestAlias")
+	vm3.RunTests("TestGVNIC")
```
hopkiw, line 87: just to clarify - you only want to run this one test on the GVNIC-enabled instance now, right? Would it be possible to bundle it with either vm1 or vm2 to reduce resource usage?

imagetest/test_suites/network/setup.go
```diff
@@ -73,7 +73,17 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		return err
 	}
 
+	vm3, err := t.CreateTestVM("vm3")
+	if err != nil {
+		return err
+	}
+	vm3.ChangeNicTypeToGVNIC()
+	if err := vm3.Reboot(); err != nil {
```
hopkiw, line 81: what is the reason to have VM 3 reboot?

imagetest/fixtures.go
```diff
@@ -240,6 +241,19 @@ func (t *TestVM) EnableSecureBoot() {
 	}
 }
 
+// ChangeNicTypeToGVNIC changes the type of vNIC to be used to GVNIC
+func (t *TestVM) ChangeNicTypeToGVNIC() {
```
hopkiw, line 245: it's a nitpick, but i feel 'change' implies amending something that exists, but we are actually setting this at launch time. maybe 'UseGVNIC' or similar?

imagetest/test_suites/windows/gvnic_test.go
```diff
@@ -0,0 +1,35 @@
+//go:build cit
```
hopkiw, line 1: i don't think we need any test in the windows suite, do we? can we make the gvnic_test in the network suite multi platform?

hopkiw (changes requested):

imagetest/fixtures.go
KarnVadaliya: Agreed, UseGVNIC sounds more appropriate here

imagetest/test_suites/network/setup.go
KarnVadaliya: I just followed the previous approach used for VM 2 wherein after adding the AliasIpRanges to the network interface the VM is rebooted.

imagetest/test_suites/network/setup.go
KarnVadaliya: I can actually however since other tests test specific network configuration will using GVNIC affect those tests output in any way?

imagetest/test_suites/network/gvnic_test.go
KarnVadaliya: yes agreed `os.Readlink()` would be a more easier approach.

imagetest/test_suites/network/setup.go
hopkiw: the reboot is necessary for the test in that case - we have a test to ensure the alias IP ranges survive a reboot.

hopkiw (approved):

PR #571, opened 2022-12-08
Fix for gvnic error for debian-10

KarnVadaliya:


imagetest/test_suites/network/setup.go
```diff
@@ -74,6 +78,16 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 	}
 	vm2.UseGVNIC()
 	vm1.RunTests("TestPingVMToVM|TestDHCP|TestDefaultMTU")
-	vm2.RunTests("TestAlias|TestGVNIC")
+	vm2.RunTests("TestAlias")
+	if strings.Contains(t.Image, "debian-10") {
```
hopkiw, line 82: this is interesting. how should we maintain this going forward? one future improvement might be to  read the features from the image at this point

imagetest/test_suites/network/setup.go
```diff
@@ -74,6 +78,16 @@ func TestSetup(t *imagetest.TestWorkflow) error {
...
+		// GVNIC Guest OS Feature not set on Debian 10
+		return nil
+	}
+	vm3, err := t.CreateTestVM("vm3")
```
hopkiw, line 86: why are we adding a third VM? it doesn't seem related to the debian-10 change

hopkiw (changes requested):

imagetest/test_suites/network/setup.go
hopkiw: oh but why not do something like this:

```go
if strings.Contains(t.Image, "debian-10") {
  vm2.RunTests("TestAlias")
} else {
  vm2.RunTests("TestAlias|TestGVNIC")
}
```

KarnVadaliya: Oh yes that seems better. Thanks

imagetest/test_suites/network/setup.go
KarnVadaliya: I tried to read from the Image the GuestOsFeatures but got a null error https://pkg.go.dev/google.golang.org/api/compute/v1#GuestOsFeature

hopkiw: discussed offline, will follow up in another PR

hopkiw (approved):

PR #575, opened 2022-12-28
use guest attributes in packagebuild

hopkiw:
Using serial console scraping is unreliable. Move package builds to use the new guest attributes feature in Daisy.

* Replace SerialOutput with GuestAttribute in common workflow, using default values
* Update common.sh to emit attributes in success and fail conditions
* Set sanity keys `enable-oslogin=FALSE` and `block-project-ssh-keys=TRUE`

### DIFF
--- a/packagebuild/common.sh
+++ b/packagebuild/common.sh
@@ -13,12 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


hopkiw:
/hold

Depends on GoogleCloudPlatform/compute-daisy#18

hopkiw:
Our current EL9 build for the guest agent is failing because it triggers some breaking reload of logs mid-startup-script. So this change is needed to unblock those builds.

PR #581, opened 2023-01-10
Update timeout for wait-for-build

jjerger:
Because 20m isn't enough for EL9. Setting to 30 to allow cleanup; plan to investigate why this is needed so suddenly later on, but this will hopefully unblock a release that needs to happen this week.


wrigri (approved):
/lgtm

PR #585, opened 2023-01-25
syft test build

koln67:
Make a test image build to add the syft arg and verify manually.

After verifying this, the syft arg will be added to the other image builds.


koln67:
/hold
Doing a bit of testing on the side before making this change

bkatyl (approved):

PR #596, opened 2023-02-08
Concourse local var reference change

koln67:
Based on the concourse documentation, it looks like this is the correct way of referencing the local var and calling string replace.


koln67:
/hold

Changing the daisyimagetask might affect files such as those in windows, which do not yet have the sbom_destination supported yet.

PR #600, opened 2023-02-10
artifact-registry-yum-plugin for el9, arm64

ChaitanyaKulkarni28:
Build artifact-registry-yum-plugin for el9, arm64 and release it in OS specific repo

### DIFF
--- a/concourse/pipelines/guest-package-build.jsonnet
+++ b/concourse/pipelines/guest-package-build.jsonnet
@@ -932,22 +932,22 @@ local buildpackageimagetask = {
     },
     buildpackagejob {
       package: 'artifact-registry-yum-plugin',
-      builds: ['el7', 'el8', 'el9'],
+      builds: ['el7', 'el8', 'el8-arm64', 'el9', 'el9-arm64'],
       uploads: [
         uploadpackagetask {
           package_paths: '{"bucket":"gcp-guest-package-uploads","object":"yum-plugin-artifact-registry/yum-plugin-artifact-registry-((.:package-version))-g1.el7.noarch.rpm"}',
           universe: 'cloud-yum',
-          repo: 'yum-plugin-artifact-registry',
+          repo: 'yum-plugin-artifact-registry-el7',
         },
         uploadpackagetask {
-          package_paths: '{"bucket":"gcp-guest-package-uploads","object":"yum-plugin-artifact-registry/dnf-plugin-artifact-registry-((.:package-version))-g1.el8.noarch.rpm"}',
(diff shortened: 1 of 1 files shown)


concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -932,31 +932,32 @@ local buildpackageimagetask = {
...
       uploads: [
         uploadpackagetask {
           package_paths: '{"bucket":"gcp-guest-package-uploads","object":"yum-plugin-artifact-registry/yum-plugin-artifact-registry-((.:package-version))-g1.el7.noarch.rpm"}',
           universe: 'cloud-yum',
-          repo: 'yum-plugin-artifact-registry',
+          repo: 'yum-plugin-artifact-registry-el7',
         },
         uploadpackagetask {
-          package_paths: '{"bucket":"gcp-guest-package-uploads","object":"yum-plugin-artifact-registry/dnf-plugin-artifact-registry-((.:package-version))-g1.el8.noarch.rpm"}',
+          package_paths: '{"bucket":"gcp-guest-package-uploads","object":"yum-plugin-artifact-registry/dnf-plugin-artifact-registry-((.:package-version))-g1.el8.noarch.rpm"},{"bucket":"gcp-guest-package-uploads","object":"yum-plugin-artifact-registry/dnf-plugin-artifact-registry-((.:package-version))-g1.el8.aarch64.rpm"}',
```
zmarano, line 943: There shouldn't be an aarch64 rpm as this is a noarch rpm already.

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -932,31 +932,32 @@ local buildpackageimagetask = {
     },
     buildpackagejob {
       package: 'artifact-registry-yum-plugin',
-      builds: ['el7', 'el8', 'el9'],
+      builds: ['el7', 'el8', 'el8-arm64', 'el9', 'el9-arm64'],
```
zmarano, line 935: Since its noarch, we don't need to build for arm64 specifically.

concourse/pipelines/guest-package-build.jsonnet
zmarano: Oh actually its the other way around.... there should be two packages for EL8 and EL9 which use DNF. Instead of `g1.el8.noarch.rpm` it should be `g1.el8.x86_64.rpm` and `g1.el8.aarch64.rpm`

ChaitanyaKulkarni28: yeah, we have go binaries which are arch specific, updated to use x86_64 for all el7, el8 and el9

concourse/pipelines/guest-package-build.jsonnet
ChaitanyaKulkarni28: as discussed these are arch specific, so arm64 is required as well

PR #602, opened 2023-02-14
Load linux date into var, use for tar gz and sbom json naming

koln67:
With the current setup, concourse tasks can only create one file. However, we need a tar.gz and sbom.json file, and both need to have the same name so we can associate them.

To ensure they can have the same name, the linux date will be loaded into a local concourse variable. This can be reused in both generate-build-id and generate-build-id-sbom.

### DIFF
--- a/concourse/templates/daisy.libsonnet
+++ b/concourse/templates/daisy.libsonnet
@@ -43,6 +43,7 @@
     // Add additional overrideable attrs.
     build_date:: '',
     gcs_url:: error 'must set gcs_url in daisy image task',
+    sbom_destination:: '',
 
     workflow_prefix+: 'build-publish/',
     vars+: [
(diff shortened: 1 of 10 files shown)


koln67:
/hold

waiting for windows builds to finish

concourse/templates/daisy.libsonnet
```diff
@@ -54,7 +55,11 @@
     ] + if self.build_date == '' then
       []
     else
-      ['build_date=' + task.build_date],
+      ['build_date=' + task.build_date
+    ] + if self.sbom_destination == '' then
+      []
+    else
+      ['sbom_destination=' + task.sbom_destination],
```
dorileo, line 62: you are checking for `self.sbom_destination` but using `task.sbom_destination` is the intentional?

koln67: Yeah, I'm copying the format for what was used for "self.build_date" a few lines earlier.

bkatyl (approved):

PR #607, opened 2023-02-15
Sbom dest

koln67:
Add sbom destination to linux image builds. 

For windows, the sbom_destination variable has no effect, so set it to any value other than the empty string.


koln67:
/hold
Temporary hold so testing can be done later

bkatyl (approved):

PR #621, opened 2023-03-02
Removing RHUI image build and release process

bkatyl:
Removing RHUI image build and release process


Quintonamore (approved):
There's an [unused lego import](https://github.com/GoogleCloudPlatform/guest-test-infra/blob/082302a302e37990fcb80839186953bcf54faeb6/concourse/pipelines/linux-image-build.jsonnet#L6) that should be removed while we're clearing out the RHUI build and release process.

PR #624, opened 2023-03-03
Fix missing variable in worker builds.

zmarano:


koln67:
Looks like this is the format which worked the previous time

  - task: generate-build-id
    file: guest-test-infra/concourse/tasks/generate-build-id.yaml
    vars:
      prefix: "rhel-7-metal"
      id: ((.:id))
  - put: rhel-7-metal-gcs

PR #648, opened 2023-04-21
Add bookworm to Concourse image build

Quintonamore:
Adds debian-12 and debian-12-arm64 image build and test pipeline.


zmarano:
Oh, I think we need to add it here as well: https://github.com/GoogleCloudPlatform/guest-test-infra/blob/e3e31926e5f559e9a8ad60bd1d5251673c22b740/concourse/templates/common.libsonnet#L7

PR #650, opened 2023-04-26
Update go version in the go integration container

vorakl:
There is one more missing place where Go version wasn't updated.
Addition to changes in PR #620


dorileo:
as we talked in private, we must guarantee backward compatibility, I'll inspect the current tests and see if we have the proper guardrails and/or if this is one of them.

PR #667, opened 2023-05-17
Fix custom hostname test.

zmarano:


dorileo:
are we considering old use cases or are we testing it all based on the transient behavior?

zmarano:
I'm looking at that question now for some of the older distros (EL7 and SLES 12 in particular).

zmarano:
Custom host names do work correctly on almost all distros (except SLES). The CLI tools used to get the FQDN have changed a lot and some tools give you static hostnames, some transient hostnames, some the kernel's hostname. For what we support today `hostname -f` will get us a result for this test case but we do need to revisit hostnames and how we handle them in the future.

PR #670, opened 2023-05-17
sbom: move it all to common.sh

dorileo:
This change moves the implementation of the steps to deploy sbom and generating the actual sbomutil to common.sh and adapts rpm and deb scripts to call the routines defined in there.


jjerger (approved):
/lgtm

PR #690, opened 2023-05-31
Add windows sbom generation

koln67:
Add windows sbom generation.

Do not merge until compute image tools has been updated with the sbom-util-gcs-root variable for windows images.


koln67:
replaced witn #692

PR #700, opened 2023-06-07
Create package jobs and all resources (including images) in Jsonnet

drewhli:


concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -0,0 +1,423 @@
...
+    },
+    run: {
+      path: 'gcloud',
+      args: [
+        'pubsub',
+        'topics',
+        'publish',
+        tl.topic,
+        '--message',
+        '{"type:" "%s", "request": {"gcsfiles": ["%s"], "universe": "%s", "repo": "%s"}}' % [
```
Quintonamore, line 57:
```suggestion
        '{"type": "%s", "request": {"gcsfiles": [%s], "universe": "%s", "repo": "%s"}}' % [
```

These could be causing the issues when verifying with `fly`.

concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -0,0 +1,423 @@
...
+  topic:: 'projects/artifact-releaser-autopush/topics/gcs-guest-package-upload-autopush',
+  type:: error 'must set type in upload_arle_autopush_task',
+  universe:: error 'must set universe in upload_arle_autopush_task',
+  repo:: error 'must set repo in upload_arle_autopush_task',
+
+  task: 'upload_arle_autopush_task',
+  config: {
+    platform: 'linux',
+    image_resource: {
+      type: 'image-registry',
```
Quintonamore, line 46:
```suggestion
      type: 'registry-image',
```

I don't know if this would cause the validation to fail, but we should use the right type anyways.

concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -0,0 +1,616 @@
...
+
+// Get the universe given the build.
+local get_universe(build) = if std.startsWith(build, 'deb') then 'cloud-apt'
+else if std.startsWith(build, 'el') then 'cloud-yum'
+else 'cloud-yuck';
+
+// Change '-' to '_', mainly used for images
+local underscore(input) = std.strReplace(input, '-', '_');
+
+local upload_arle_autopush_task = {
```
dorileo, line 18: I wonder if we can't reuse whatever is defined within `arle.libsonnet` both here and for the `_stagging` variant, or even maybe have their definition moved to a common place such as `arle.libsonnet` or something.

concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -0,0 +1,616 @@
...
+    },
+  },
+};
+
+
+local upload_arle_autopush_staging = {
+  local tl = self,
+
+  package:: error 'must set package in upload_arle_autopush_staging',
+  gcs_dir:: tl.package + '_',
```
dorileo, line 58: why the `_` suffix in the `gcs_dir`? Why not something more meaningful if we really need it?

concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -0,0 +1,616 @@
...
+  ],
+};
+
+{
+  // Image names.
+  local almalinux = ['almalinux-8', 'almalinux-9'],
+  local debian = ['debian-10', 'debian-11', 'debian-11-arm64', 'debian-12', 'debian-12-arm64'],
+  local centos = ['centos-7', 'centos-stream-8', 'centos-stream-9'],
+  local rhel = [
+    'rhel-7',
```
dorileo, line 296: Side Note: these are some kinds of repetition I'd love if we could stop propagating and start adding a more generic set of configurations and abstraction that make sense in a more global sense, `rhel` or enterprise linux distribution names are stable and equal across all contexts.

dorileo (commented):
Good job here, I had some minor comments but I'm not blocking this PR based on them, I'd have it merged and iterate over it.

concourse/pipelines/artifact-releaser-test.jsonnet
drewhli: The name is definitely misleading. I think it'd be better to rename it to 'gcs_file_name' instead (the name of the file up until where the package name is defined). The reason for this underscore is because of certain packages (looking at you, _compute-image-windows_) where the package version is not preceded by an underscore like most other packages.

concourse/pipelines/artifact-releaser-test.jsonnet
drewhli: Acknowledged.

concourse/pipelines/artifact-releaser-test.jsonnet
drewhli: The defined tasks in `arle.libsonnet` can't really be used for packages as far as I understand things; they seem more catered to images. However, the package upload task could definitely be moved to a more common 'library' so-to-speak.

Quintonamore (approved):

PR #706, opened 2023-06-26
Use bullseye base containers due to incompatibilites with bookworm based containers.

zmarano:


zmarano:
https://github.com/docker-library/golang/issues/467 has one of the issues but there seem to be many

PR #717, opened 2023-07-05
daisy: build daisy statically & update container image

dorileo:
Make sure we are building daisy statically; additionally we are
updating the container image to be consistent across the other
build environments.


concourse/pipelines/container-build.jsonnet
```diff
@@ -208,11 +208,11 @@ local BuildContainerImage(image) = buildcontainerimgjob {
               platform: 'linux',
               image_resource: {
                 type: 'registry-image',
-                source: { repository: 'golang' },
+                source: { repository: 'golang:bullseye' },
```
zmarano, line 211: golang:1.19-bullseye

PR #720, opened 2023-07-07
Fix various issues with artifact-releaser-test

drewhli:


zmarano (approved):

drewhli:
Decided to put all the fixes in the same PR. I've found a problem where the file endings did not exactly match that found on GCS, so this PR should fix that. This PR also introduces a rather unelegant, but working solution to the `compute-image-windows` package file endings. The solution will be reverted when a standard is decided on for at least those GooGet packages regarding file endings.

PR #725, opened 2023-07-13
Fix more minor issues with artifact releaser testing pipeline

drewhli:
Fix certgen file name, and make the tag filter even more stringent to avoid cases where the tag is, for example, 20180708a, which would make the pipeline fail.

### DIFF
--- a/concourse/pipelines/artifact-releaser-test.jsonnet
+++ b/concourse/pipelines/artifact-releaser-test.jsonnet
@@ -380,7 +380,7 @@ local pkggroup = {
   ],
   local compute_image_windows_packages = ['certgen', 'auto-updater', 'powershell', 'sysprep', 'ssh'],
   local compute_image_windows_gcs_names = [
-    'certgen.x86_64.x86_64',
+    'certgen.x86_64',
     'google-compute-engine-auto-updater.noarch',
     'google-compute-engine-powershell.noarch',
     'google-compute-engine-sysprep.noarch',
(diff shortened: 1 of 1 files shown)


concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -445,7 +445,7 @@ local pkggroup = {
                    owner: 'GoogleCloudPlatform',
                    repository: package,
                    access_token: '((github-token.token))',
-                   tag_filter: '([0-9]+.*)',
+                   tag_filter: '([0-9]+\\.[0-9]+)',
```
dorileo, line 448: is this expression change meant to match both `certgen.x86_64.x86_64` and `certgen.x86_64`?

concourse/pipelines/artifact-releaser-test.jsonnet
```diff
@@ -390,8 +390,8 @@ local pkggroup = {
     '.0@1.goo',
     '@1.goo',
     '@1.goo',
-    '.0@1.goo',
     '@1.goo',
+    '.0@1.goo',
```
dorileo, line 394: I'm curious to know/understand this ordering change impact.

concourse/pipelines/artifact-releaser-test.jsonnet
drewhli: This is for matching versions, so it won't be matching the file names. According to the errors in concourse right now, it is getting the `20180522a` tag for `compute-image-tools`. The extra `a` messes things up.

concourse/pipelines/artifact-releaser-test.jsonnet
drewhli: So every index of file endings is matched with the same index in (in this case) gcs package names. I mismatched them, causing some errors in the testing pipeline.

PR #727, opened 2023-07-13
Allow filtering on ready images

ChaitanyaKulkarni28:


container_images/gce-img-resource/check/check_command.go
```diff
@@ -39,10 +39,19 @@ func Run(request Request) (Response, error) {
 	}
 
 	call := computeService.Images.List(request.Source.Project)
+
+	var filter string
+
+	if request.Source.ReadyOnly {
```
dorileo, line 45: Can you update the comment starting on line 13?

container_images/gce-img-resource/check/check_command.go
```diff
@@ -39,10 +39,19 @@ func Run(request Request) (Response, error) {
...
+		filter = "(status = READY) "
+	}
+
 	if request.Source.Family != "" {
-		call = call.Filter(fmt.Sprintf("family = %s", request.Source.Family))
+		filter = filter + fmt.Sprintf("(family = %s)", request.Source.Family)
```
dorileo, line 50: Did you check if this format is correct/valid? `(status = READY)(family = family-name)` if I were to guess I'd have guessed `(status = READY, family = family-name)`.

dorileo: Nitpick: if the space between the tokens are relevant you could use a `strings.Join([]string{...}, " ")` instead.

ChaitanyaKulkarni28: yep, that's the right format. Reference: https://cloud.google.com/compute/docs/reference/rest/v1/images/list

container_images/gce-img-resource/check/check_command.go
ChaitanyaKulkarni28: Done

container_images/gce-img-resource/check/check_command.go
ChaitanyaKulkarni28: > Nitpick: if the space between the tokens is relevant you could use a `strings.Join([]string{...}, " ")` instead.

Done

PR #730, opened 2023-07-14
Use commands for dhcp test

koln67:
The previous dhcp test checks for a dhcp client, which is deprecated on most systems.

Instead, use the commands which are available on the image to check for dhcp ipv4.


koln67:
Currently still investigating the nmcli output and why some images are failing the dhcp test for that command.

PR #732, opened 2023-07-19
Use commands for DHCP test

koln67:
Check if the dhcp IPv4 is found using commands. If any of the cases pass, pass the test.

### DIFF
--- a/imagetest/test_suites/network/dhcp_test.go
+++ b/imagetest/test_suites/network/dhcp_test.go
@@ -26,21 +26,21 @@ func TestDHCP(t *testing.T) {
 	var err error
 
 	// Run every case: if one command or check succeeds, the test passes.
-	if checkCmdExists(networkctlCmd) {
+	if utils.CheckLinuxCmdExists(networkctlCmd) {
 		cmd = exec.Command(networkctlCmd, "status")
 		if err = parseNetworkctlOutput(cmd); err == nil {
 			return
 		}
 	}
 
-	if checkCmdExists(nmcliCmd) {
+	if utils.CheckLinuxCmdExists(nmcliCmd) {
 		cmd = exec.Command(nmcliCmd, "device", "show")
 		if err = parseNmcliOutput(cmd); err == nil {
 			return
 		}
 	}
 
-	if checkCmdExists(wickedCmd) {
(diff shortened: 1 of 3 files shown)


imagetest/test_suites/network/dhcp_test.go
```diff
@@ -13,40 +15,151 @@ import (
...
+			return nil
 		}
 	}
-	t.Fatalf("failed finding dhcp client process")
+	return fmt.Errorf("failed finding dhclient process")
+}
+
+// accepts IP addresses in the form of a.b.c.d or a.b.c.d/IPNET
+func validIPOrCIDR(token string) bool {
+	fmt.Printf("token is %s", token)
```
dorileo, line 145: please use `t.Logf()`

imagetest/test_suites/network/dhcp_test.go
```diff
@@ -13,40 +15,151 @@ import (
...
+	IPAddress, _, err := net.ParseCIDR(token)
+	if IPAddress != nil && err == nil {
+		return true
+	}
+
+	return false
+}
+
+func checkCmdExists(cmd string) bool {
+	_, err := exec.LookPath(cmd)
```
dorileo, line 159: if you are checking for cmd (or executables) please also test it's executable.

imagetest/test_suites/network/dhcp_test.go
koln67: this print statement was used for debugging but is unnecessary in the test, removed.

imagetest/test_suites/network/dhcp_test.go
koln67: https://pkg.go.dev/os/exec#example-LookPath

exec.LookPath() searches for an executable in the path. I believe this checks if the cmd variable is executable, was there a different specific check you had in mind?

dorileo: `exec.LookPath()` will not check if the file has `+x` mode, you need to check if yourself, you can use the file's `Mode` to check it.

koln67: these tests run with root/sudo permissions, would this check be necessary?

PR #737, opened 2023-07-26
Create GVE performance test

drewhli:
Formatting will be fixed in the next commit. I have my VIM tab set to 2 spaces, which doesn't match the tab characters being used here.


imagetest/test_suites/network/gvnic_test.go
```diff
@@ -45,8 +50,27 @@ func CheckGVNICPresentWindows(interfaceName string) error {
 	return errors.New("GVNIC not present")
 }
 
+func CheckGVNICPerformance() (string, error) {
+	// Wait for iperf to finish.
+	for i := 0; i <= maxRetries; i++ {
```
dorileo, line 55: If I understand this code correctly you are waiting until the `iperf` finishes, right? In that case is the `10 * (60 * time.Second)` duration aligned with the iperf test timeout?

imagetest/test_suites/network/setup.go
```diff
@@ -10,10 +10,17 @@ import (
 var Name = "network"
 
 const (
-	vm1Name = "vm1"
-	vm2Name = "vm2"
-	vm1IP   = "192.168.0.2"
-	vm2IP   = "192.168.0.3"
+	vm1Name    = "vm1"
```
dorileo, line 13: with all these constants I wonder if wouldn't be more elegant to introduce a `type InstanceConfig struct` and wrap it all with it.

imagetest/test_suites/network/gvnic_test.go
drewhli: Yea. As for the second question, I will discuss with you privately. Basically, the same interval was used in a similar test somewhere else, so I've decided to reuse the same interval.

dorileo: Should be fine. The retry logic looks correct, just wonder if the times are realistic - but they can be throttled later with more "real data".

imagetest/test_suites/network/setup.go
drewhli: Done.

imagetest/test_suites/network/gvnic_test.go
zmarano, line 1: In order to prevent this from blocking releases, we could make a new test suite for performance tests or create an exclusion in the release pipeline (https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/concourse/pipelines/linux-image-build.jsonnet#L35) .... I think I would prefer the dedicated test suite only because we can then instrument it as needed. Although, either way will make sure we're not (yet) blocking releases on performance tests?

imagetest/test_suites/network/gvnic_test.go
```diff
@@ -45,8 +50,27 @@ func CheckGVNICPresentWindows(interfaceName string) error {
...
+		results, err := utils.GetMetadataGuestAttribute("testing/results")
```
zmarano, line 56: The test wrapper uploads its results to GCS (https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/imagetest/cmd/wrapper/main.go#L104) and thus we already have a good place to stick test data that will stick around. Instead of using guest attributes, we could either directly upload the perf results to the same location or merge them back into the junit output that gets uploaded there at the end.

imagetest/test_suites/network/gvnic_test.go
dorileo: +1

drewhli: Done.

imagetest/test_suites/network/gvnic_test.go
drewhli: The results are being merged with the junit output with this method, since we're outputting the results of the test as part of the test. The iperf results are viewable from the results uploaded to GCS.

drewhli:
Weird issue where the networks aren't getting cleaned up correctly due to them being in use by a firewall rule at the time of deletion. Investigating. But otherwise good to go.

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -508,7 +508,7 @@ local buildpackageimagetask = {
                       '-zone=us-central1-a',
                       '-test_projects=compute-image-test-pool-002,compute-image-test-pool-003,compute-image-test-pool-004,compute-image-test-pool-005',
                       '-images=projects/gcp-guest/global/images/debian-10-((.:build-id)),projects/gcp-guest/global/images/debian-11-((.:build-id)),projects/gcp-guest/global/images/debian-12-((.:build-id)),projects/gcp-guest/global/images/centos-7-((.:build-id)),projects/gcp-guest/global/images/rhel-7-((.:build-id)),projects/gcp-guest/global/images/rhel-8-((.:build-id)),projects/gcp-guest/global/images/rhel-9-((.:build-id))',
-                      '-exclude=(image)|(disk)|(security)|(oslogin)|(storage_perf)',
+                      '-exclude=(image)|(disk)|(security)|(oslogin)|(storage_perf)|(gveperf)',
```
zmarano, line 511: Let's go with network_perf so that we can represent other network devices in the future (at least virtio.

imagetest/test_suites/gveperf/startupscripts/netserver_startup.sh
```diff
@@ -0,0 +1,37 @@
...
+# to test the network bandwidth between the two VMs.
+
+sleepduration=120
+
+if [[ -f /usr/bin/apt ]]; then
+  echo "apt found Installing iperf."
+  sudo apt update && sudo apt install iperf
+elif [[ -f /bin/apt-get ]]; then
+  echo "apt found Installing iperf."
+  sudo apt-get update && sudo apt-get install iperf
```
zmarano, line 13: apt and apt-get are functionally the same from this point of view, we can remove the apt-get case (it will never be called).

imagetest/test_suites/gveperf/startupscripts/netserver_startup.sh
```diff
@@ -0,0 +1,37 @@
...
+elif [[ -f /bin/dnf ]]; then
+  echo "dnf found Installing iperf."
+  os=$(cat /etc/redhat-release)
+  if [[ "$os" == *"release 9."* ]]; then
+    sudo dnf -y install https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/Packages/i/iperf-2.1.6-2.el9.x86_64.rpm
```
zmarano, line 18: These package calls will likely break over time as versions move. Instead, for EPEL it is better to install the epel-release package and then install iperf from the repo. https://docs.fedoraproject.org/en-US/epel/#how_can_i_use_these_extra_packages

imagetest/test_suites/gveperf/startupscripts/netserver_startup.sh
```diff
@@ -0,0 +1,37 @@
...
+  sudo dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
+  sudo sudo dnf makecache && sudo dnf -y install iperf
+elif [[ ! -f /bin/dnf ]] && [[ -f /bin/yum ]]; then
+  echo "yum found Installing iperf."
+  yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  sudo sudo yum makecache && sudo yum -y install iperf
+elif [[ -f /usr/bin/zypper ]]; then
+  echo "zypper found Installing iperf."
+  sudo zypper --no-gpg-checks refresh
+  sudo zypper --no-gpg-checks --non-interactive install https://iperf.fr/download/opensuse/iperf-2.0.5-14.1.2.x86_64.rpm
```
zmarano, line 33: Similar issue with this- but let's see if I can find a better solution.

concourse/pipelines/guest-package-build.jsonnet
drewhli: Done.

imagetest/test_suites/gveperf/startupscripts/netserver_startup.sh
drewhli: Done.

imagetest/test_suites/network_perf/startupscripts/netserver_startup.sh
```diff
@@ -0,0 +1,32 @@
...
+  fi
+  sudo sudo dnf makecache && sudo dnf -y install iperf
+elif [[ -f /bin/yum ]]; then
+  echo "yum found Installing iperf."
+  yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  sudo sudo yum makecache && sudo yum -y install iperf
+elif [[ -f /usr/bin/zypper ]]; then
+  echo "zypper found Installing iperf."
+  sudo zypper --no-gpg-checks refresh
+  sudo zypper --no-gpg-checks --non-interactive install iperf
```
zmarano, line 28: For SLES, I found we can use the PackageHub repo which is version and arch dependent.

For SLES 15 SP5 we have to do the following.

```
SUSEConnect --product PackageHub/15.5/x86_64
zypper install iperf
```

For SLES 12 SP5
```
SUSEConnect --product PackageHub/12.5/x86_64
zypper install iperf
```

Note that those versions and architectures will have to be handled.

imagetest/test_suites/network_perf/setup.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+	if err := clientVM.SetPrivateIP(clientServerNetwork, clientConfig.ip); err != nil {
+		return err
+	}
+	clientVM.AddMetadata("enable-guest-attributes", "TRUE")
+	clientVM.AddMetadata("iperftarget", serverConfig.ip)
+	clientVM.SetStartupScript(string(clientStartup))
+
+	if strings.Contains(t.Image, "debian-10") || strings.Contains(t.Image, "rhel-7-7-sap") || strings.Contains(t.Image, "rhel-8-1-sap") {
```
dorileo, line 83: Can you make this kind of check top down rather than bottom up like this? Can we make it more declarative on the tests registration what image families are supported etc? Ideally make it central? so we can toggle these behaviors in a single place?

imagetest/test_suites/network_perf/startupscripts/netserver_startup.sh
drewhli: Done.

imagetest/test_suites/network_perf/setup.go
drewhli: Yes. Done. For the centralized part, that can be done as some sort of cleanup work. Won't do in this PR.

PR #738, opened 2023-07-26
Support multiple disks in a VM

koln67:
Allow multiple disks to be added in the create VM step

### DIFF
--- a/imagetest/testworkflow.go
+++ b/imagetest/testworkflow.go
@@ -83,7 +83,12 @@ func (t *TestWorkflow) appendCreateVMStep(disks []DiskInfo, hostname string) (*d
 
 	for _, disk := range disks {
 		diskname, disktype := disk.Name, disk.Type
-		attachedDisk := &compute.AttachedDisk{Source: diskname, Type: disktype}
+		diskInitParams := &compute.AttachedDiskInitializeParams{}
+		// set the disk type for the api, default is pd-standard if not set.
+		if disktype != "" {
+			diskInitParams.DiskType = disktype
+		}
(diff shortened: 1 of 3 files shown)


koln67:
Currently unsure about AttachedDisk and Disk, so pr is on hold

imagetest/testworkflow.go
```diff
@@ -69,7 +80,13 @@ func (t *TestWorkflow) appendCreateVMStep(name, hostname string) (*daisy.Step, *
 	instance.StartupScript = fmt.Sprintf("wrapper%s", suffix)
 	instance.Name = name
 	instance.Scopes = append(instance.Scopes, "https://www.googleapis.com/auth/devstorage.read_write")
-	instance.Disks = append(instance.Disks, attachedDisk)
+
+	for _, disk := range disks {
```
dorileo, line 84: are you sure we want do add all items in the disks slice? I guess we want to skip `root` disk (AKA `disks[0]`), right?

koln67: before this PR, there was only one item passed in as the disk, and that disk was appended.

I'm assuming that should be added, however I'm trying to figure out if it should be of type "Disk" or "AttachedDisk."

It is also possible that the first disk should be of type "AttachedDisk" and the second and future disks should be of type "Disk"

koln67: It appears that instance.Disks starts out as a slice with length 0. This means we want to append all the disks.

koln67:
/hold

It seems that using a map from vm name to Attached Disk Params might be easier

imagetest/testworkflow.go
dorileo: sounds good.

PR #742, opened 2023-07-28
Allow test vm to have multiple disks

koln67:
Functionality of the current tests should not change. This should allow a future change to create a test vm with 2 or more disks.


jjerger (approved):
/lgtm

PR #746, opened 2023-08-03
Non-blocking Testsuite

koln67:
proposed changes to add test suite(s) which give performance metrics but do not block image releases.


imagetest/cmd/manager/main.go
```diff
@@ -13,6 +13,7 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/test_suites/disk"
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/test_suites/fioperf"
```
zmarano, line 16: I think we should create a generic performance test suite and then within it have the different tests for storage, networking, etc. Also, let's use storage_perf instead of fio so it is inclusive of other tools (fio is just one possible tool).

imagetest/cmd/manager/main.go
```diff
@@ -199,6 +215,12 @@ func main() {
 		},
 	}
 
+	// These tests are excluded by default so they do not block releases.
+	skippedNonblockingTests := make(map[string]bool)
+	for _, skippedNonblockingTestName := range []string{fioperf.Name} {
```
drewhli, line 220: It may be a good idea to make the array of blocking test names a module-level variable. That way it'll be easier to find and update the list of tests whenever a new one is added.

imagetest/cmd/manager/main.go
```diff
@@ -38,6 +39,7 @@ var (
 	parallelCount = flag.Int("parallel_count", 5, "TestParallelCount")
 	filter        = flag.String("filter", "", "only run tests matching filter")
 	exclude       = flag.String("exclude", "", "skip tests matching filter")
+	extraTests    = flag.String("extra_tests", "", "run extra tests which would normally not be run and do not block releases")
```
zmarano, line 42: I'm wondering if this flag makes sense since we already have filter (which is basically include only) and exclude. The choice of which test suites to run is left to the place it is being instrumented (in our CI pipeline or elsewhere for example).

koln67: storage_perf is now excluded in the concourse files

imagetest/cmd/manager/main.go
koln67: renamed to storage_perf

imagetest/cmd/manager/main.go
koln67: the test suites which are excluded are now in the concourse files

PR #749, opened 2023-08-08
guest-oslogin: exclude storage perf tests

dorileo:
We accidentally added test in the filter flag rather than the exclude, since we want to have it excluded this patches applies the change.


concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -751,7 +751,7 @@ local buildpackageimagetask = {
                       '-zone=us-central1-a',
                       '-test_projects=oslogin-testing-project',
                       '-images=projects/gcp-guest/global/images/debian-10-((.:build-id)),projects/gcp-guest/global/images/debian-11-((.:build-id)),projects/gcp-guest/global/images/debian-12-((.:build-id)),projects/gcp-guest/global/images/centos-7-((.:build-id)),projects/gcp-guest/global/images/rhel-7-((.:build-id)),projects/gcp-guest/global/images/rhel-8-((.:build-id)),projects/gcp-guest/global/images/rhel-9-((.:build-id))',
-                      '-filter=(oslogin)|(storage_perf)',
+                      '-filter=(oslogin) -exclude=(storage_perf)',
```
drewhli, line 754: I think we can just remove this line. So it'll just look like `-filter=oslogin`. The reason being the filter already excludes all other tests except those that match the filter, so the exclude flag would be superfluous.

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -773,7 +773,7 @@ local buildpackageimagetask = {
                       '-test_projects=oslogin-testing-project',
                       '-images=projects/gcp-guest/global/images/debian-11-arm64-((.:build-id)),projects/gcp-guest/global/images/debian-12-arm64-((.:build-id)),projects/gcp-guest/global/images/rocky-linux-8-optimized-gcp-arm64-((.:build-id)),projects/gcp-guest/global/images/rhel-9-arm64-((.:build-id))',
                       '-machine_type=t2a-standard-2',
-                      '-filter=(oslogin)|(storage_perf)',
+                      '-filter=(oslogin) -exclude=(storage_perf)',
```
drewhli, line 776: Ditto previous comment.

concourse/pipelines/guest-package-build.jsonnet
dorileo: will do, thanks.

concourse/pipelines/guest-package-build.jsonnet
dorileo: done, thanks.

PR #750, opened 2023-08-08
fio read test

koln67:
Add mount steps, and eventually fio test outputs to the storage perf test.

### DIFF
--- a/imagetest/test_suites/storage_perf/iops_print_test.go
+++ b/imagetest/test_suites/storage_perf/iops_print_test.go
@@ -4,9 +4,12 @@
 package storageperf
 
 import (
+	"encoding/json"
+	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
@@ -17,14 +20,55 @@ const (
 	testreadOutputDir = "/mnt/disks/mount_dir"
 )
 
-// TestIOPSPrint is a placeholder test which prints out info about iops
-func TestIOPSPrint(t *testing.T) {
-	mountDiskSymlink := "/dev/disk/by-id/google-" + mountDiskName
-	symlinkRealPath, err := filepath.EvalSymlinks(mountDiskSymlink)
+type BlockDeviceList struct {
+	BlockDevices []BlockDevice `json:"blockdevices,omitempty"`
+}
+
+type BlockDevice struct {
+	Name string `json:"name,omitempty"`
+	Size string `json:"size,omitempty"`
+	Type string `json:"type,omitempty"`
+	// Other fields are not currently used.
+	X map[string]interface{} `json:"-"`
+}
+
+func getMountDiskPartition(diskExpectedSizeGb int) (string, error) {
+	lsblkCmd := "lsblk"
+	if !utils.CheckLinuxCmdExists(lsblkCmd) {
+		return "", fmt.Errorf("could not find lsblk")
+	}
+	lsblkout, err := exec.Command(lsblkCmd, "-o", "name,size,type", "--json").Output()
 	if err != nil {
-		t.Fatalf("symlink could not be resolved: %v", err)
+		return "", fmt.Errorf("failed to execute lsblk: %v", err)
 	}
 
+	var blockDevices BlockDeviceList
+	if err := json.Unmarshal(lsblkout, &blockDevices); err != nil {
+		return "", fmt.Errorf("failed to unmarshal lsblk output: %v", err)
+	}
+
+	diskExpectedSizeGbString := strconv.Itoa(diskExpectedSizeGb) + "G"
+	for _, blockDev := range blockDevices.BlockDevices {
+		if strings.ToLower(blockDev.Type) == "disk" && blockDev.Size == diskExpectedSizeGbString {
+			return blockDev.Name, nil
+		}
+	}
+
+	return "", fmt.Errorf("disk block with size not found")
+}
+
+// TestIOPSPrint is a placeholder test which prints out info about iops
+func TestIOPSPrint(t *testing.T) {
+	//mountDiskSymlink := "/dev/disk/by-id/google-" + mountDiskName
+	//symlinkRealPath, err := filepath.EvalSymlinks(mountDiskSymlink)
+	//if err != nil {
+	//	t.Fatalf("symlink could not be resolved: %v", err)
+	//}
+	diskPartition, err := getMountDiskPartition(HyperdiskSize)
+	if err != nil {
+		t.Fatalf("did not find mount disk partition: %v", err)
+	}
+	symlinkRealPath := "/dev/" + diskPartition
 	if !utils.CheckLinuxCmdExists(mkfsCmd) {
 		t.Fatalf("could not format mount disk: %s cmd not found", mkfsCmd)
 	}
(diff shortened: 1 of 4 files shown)


imagetest/test_suites/storage_perf/iops_print_test.go
```diff
@@ -4,10 +4,131 @@
...
+		}
+	}
+
+	return "", fmt.Errorf("disk block with size not found")
+}
+
+func installFio() error {
+	var installFioCmd *exec.Cmd
+	if utils.CheckLinuxCmdExists("apt") {
+		installFioCmd = exec.Command("apt", "install", "-y", "fio")
```
zmarano, line 64: always run `apt update` before trying to install (there's no guarantee that a local package cache exists yet).

imagetest/test_suites/storage_perf/iops_print_test.go
```diff
@@ -4,10 +4,131 @@
...
+		symlinkRealPath, err = getMountDiskPartitionSymlink()
+		if err != nil {
+			t.Fatalf("failed to find symlink to mount disk with any method: error %v", err)
+		}
+	}
+
+	if !utils.CheckLinuxCmdExists(mkfsCmd) {
+		t.Fatalf("could not format mount disk: %s cmd not found", mkfsCmd)
+	}
+	mkfsFullCmd := exec.Command(mkfsCmd, "-m", "0", "-E", "lazy_itable_init=0,lazy_journal_init=0,discard", symlinkRealPath)
```
zmarano, line 115: There are two ways to benchmark performance- one is against the raw block device, other is against a mounted filesystem.  We should benchmark against the block device (so that the filesystem is not part of the equation). Filesystem types, configurations, options, and even versions can differ greatly. For our purpose we should look at the block device performance: https://cloud.google.com/compute/docs/disks/benchmarking-pd-performance#raw-disk

koln67: changed to use the filesystem command

imagetest/test_suites/storage_perf/iops_print_test.go
koln67: changed to run update before install

imagetest/test_suites/storage_perf/iops_read_test.go
```diff
@@ -0,0 +1,56 @@
...
+		if err := installFio(); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// Arbitrary file read size, less than the size of hte hyperdisk in GB.
+	fileReadSizeString := strconv.Itoa(HyperdiskSize/10) + "G"
+	iopsJson, err := exec.Command("fio", "--name=read_iops_test", "--filename="+symlinkRealPath, "--filesize="+fileReadSizeString, "--time_based", "--ramp_time=2s", "--runtime=1m", "--ioengine=libaio", "--direct=1", "--verify=0", "--randrepeat=0", "--bs=4k", "--iodepth=256", "--rw=randread", "--iodepth_batch_submit=256", "--iodepth_batch_complete_max=256", "--output-format=json").Output()
+	if err != nil {
+		t.Fatalf("fio command failed with %v", err)
```
jjerger, line 41: Being a little nitpicky here, but the formatting could be adjusted a bit to match how we usually pass errors up the chain. Maybe something like:

"fio command failed: %v"

Or maybe (verbose):
"fio command failed with error: %v"

imagetest/test_suites/storage_perf/iops_read_test.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}
+
+	if !utils.CheckLinuxCmdExists("fio") {
+		if err := installFio(); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// Arbitrary file read size, less than the size of hte hyperdisk in GB.
```
zmarano, line 37:
```suggestion
	// Arbitrary file read size, less than the size of the hyperdisk in GB.
```

imagetest/test_suites/storage_perf/storage_perf_utils.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+const (
+	vmName = "vm"
+	// iopsErrorMargin allows for a small difference between iops found in the test and the iops value listed in public documentation.
+	iopsErrorMargin = 0.97
+	// hyperdiskSize is used to determine which partition is the mounted hyperdisk.
+	hyperdiskSize = 100
+	bootdiskSize  = 10
```
zmarano, line 20: Boot disk size varies by OS. For EL based images, it is 20GB by default, and Windows uses 30GB. I would suggest making this either larger than the biggest public image (50GB should do it) or allowing the API to create the default size (which is whatever the image size is).

imagetest/test_suites/storage_perf/storage_perf_utils.go
```diff
@@ -0,0 +1,128 @@
...
+type blockDevice struct {
+	name       string `json:"name,omitempty"`
+	size       string `json:"size,omitempty"`
+	deviceType string `json:"type,omitempty"`
+	// Other fields are not currently used.
+	X map[string]interface{} `json:"-"`
+}
+
+func getMountDiskPartition(diskExpectedSizeGb int) (string, error) {
+	lsblkCmd := "lsblk"
```
zmarano, line 55: Even if we do it in future iterations, it's worth stubbing out where we need to split Linux and Windows logic. This would be one place. And we could either do this with the function (this one is Linux specific) or inside the function.

imagetest/test_suites/storage_perf/iops_read_test.go
koln67: changed

imagetest/test_suites/storage_perf/storage_perf_utils.go
koln67: changed to 50gb

imagetest/test_suites/storage_perf/storage_perf_utils.go
koln67: currently commented so that windows logic can be added in a future iteration

imagetest/test_suites/storage_perf/storage_perf_utils.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// installFio tries to install fio with any of multiple package managers, and returns an error if all the package managers were not found or failed.
+func installFio() error {
+	var updateCmd, installFioCmd *exec.Cmd
+	if utils.CheckLinuxCmdExists("apt") {
+		updateCmd = exec.Command("apt", "-y", "update")
+		installFioCmd = exec.Command("apt", "install", "-y", "fio")
+	} else if utils.CheckLinuxCmdExists("yum") {
+		updateCmd = exec.Command("yum", "update")
```
zmarano, line 92: Probably avoid doing this- it may take a long time. This updates the entire system and isn't necessary.

koln67: done

imagetest/test_suites/storage_perf/storage_perf_utils.go
```diff
@@ -0,0 +1,123 @@
...
+		installFioCmd = exec.Command("yum", "-y", "install", "fio")
+	} else if utils.CheckLinuxCmdExists("zypper") {
+		updateCmd = exec.Command("zypper", "refresh")
+		installFioCmd = exec.Command("zypper", "--non-interactive", "install", "fio")
+	} else if utils.CheckLinuxCmdExists("dnf") {
+		updateCmd = exec.Command("dnf", "upgrade")
```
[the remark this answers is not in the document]
koln67: done

PR #752, opened 2023-08-10
Create gvnic jumbo frames performance test

drewhli:
- Removing rather irrelevant outputs from the client side of the `iperf` tests.
- Adding a method for creating a new network with a specified MTU
- Update startup scripts to handle running a second `iperf` test using the 8896 MTU network.
- Nits

Issues:
- The default network still randomly fails to get cleaned up after tests. However, the jumbo frames networks cleanup just fine...


imagetest/test_suites/network_perf/setup.go
```diff
@@ -32,22 +35,37 @@ const (
 func TestSetup(t *imagetest.TestWorkflow) error {
 	if strings.Contains(t.Image, "debian-10") || strings.Contains(t.Image, "rhel-7-7-sap") || strings.Contains(t.Image, "rhel-8-1-sap") {
```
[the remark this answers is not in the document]
drewhli: Done.

drewhli:
Regarding the most recent commit, these are the reasons why:
- `iperf3` was having very weird interactions with `netcat`, which was causing problems for `iperf` to actually perform and finish properly.
- For specifically SLE12 SP5, there was an error with the server rejecting packets sent by the client, related to having too many connections. The threshold seemed to be 14, so reducing to 12 should avoid that error, and shouldn't have any significant impacts on the results of the tests.

drewhli:
Most recent commit also:
- Fixed an issue with using image families with the `images` flag.
- Added a hard timeout of 5min to the client VM's check for the server VM's iperf server.

imagetest/test_suites/network_perf/targets.txt
```diff
@@ -0,0 +1,154 @@
...
+        "n1-highcpu-2":   10,
+        "n1-highcpu-4":   10,
+        "n1-highcpu-8":   10,
+        "n1-highcpu-16":  16,
+        "n1-highcpu-32":  32,
+        "n1-highcpu-48":  32,
+        "n1-highcpu-64":  32,
+        "n1-highcpu-80":  32,
+        "n1-highcpu-96":  32,
+        "f1-core":  1,
```
zmarano, line 73: We can remove f1 and g1 as well

imagetest/test_suites/network_perf/targets.txt
```diff
@@ -0,0 +1,154 @@
+{
```
zmarano, line 1: In general, to keep the scale of test runs more manageable, we should test the points when performance is expected to change. That being, the smallest machine shape at a given performance level. There may be a reason to also test performance with the largest machine shape in a series (TBD).

imagetest/test_suites/network_perf/targets.txt
```diff
@@ -0,0 +1,154 @@
...
+        "e2-highmem-2" : 4,
+        "e2-highmem-4" : 8,
+        "e2-highmem-8" : 16,
+        "e2-highmem-16": 16,
+        "e2-highcpu-2" : 4,
+        "e2-highcpu-4" : 8,
+        "e2-highcpu-8" : 16,
+        "e2-highcpu-16": 16,
+        "e2-highcpu-32": 16,
+        "e2-micro":  1,
```
zmarano, line 41: Let's leave out the shared cored machine types such as e2-micro and e2-small largely because of their memory limitations which will not run some OSes.

imagetest/test_suites/network_perf/performance_test.go
```diff
@@ -0,0 +1,22 @@
+//go:build cit
+// +build cit
+
+package gveperf
```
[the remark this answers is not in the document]
drewhli: Done.

imagetest/test_suites/network_perf/targets.txt
drewhli: Done.

imagetest/test_suites/network_perf/targets.txt
drewhli: Done.

imagetest/test_suites/network_perf/targets.txt
drewhli: Acknowledged. Will think of a way to do this efficiently.

drewhli: Done.

PR #756, opened 2023-08-17
Change imagetest's image map to a project map

drewhli:


zmarano:
Do we know what this is used for? There are a bunch of images in that list which are long since EOL etc... I doubt this map is fully used anywhere.

drewhli:
This is used for the `-images` flag. Users can take an easy route and instead of typing out the full image path, e.g. `projects/debian-cloud/global/images/debian-11-bullseye-v20230711`, they can just pass in `debian-11` and it'll get the respective image from the image family.

zmarano:
> This is used for the `-images` flag. Users can take an easy route and instead of typing out the full image path, e.g. `projects/debian-cloud/global/images/debian-11-bullseye-v20230711`, they can just pass in `debian-11` and it'll get the respective image from the image family.

I actually think we should remove this entirely (it's a maintenance burden). And instead abstract the two parts of the image request - project and family, or project and image. WDYT

drewhli:
I think that's probably the way to go. I have a few ideas about how to go about it:
- Use the map as a project map instead. So any images containing `debian will be mapped to the `debian-cloud`, for example, and the image/family path can be formed from there.
  - Testing specific versions would need some regex-ing, hopefully image names follow a convention with relation to versioning
- Enforce a convention on the `-images` flag to something like `$project/$family` or `$project/$image`, and we can simply extract the important bits.

I'm leaning more towards the first method, since the projects and image prefixes are highly unlikely to be changed.

imagetest/cmd/manager/main.go
```diff
@@ -44,62 +44,19 @@ var (
...
-		"windows-2019-core":                "projects/windows-cloud/global/images/family/windows-2019-core",
-		"windows-2019-core-for-containers": "projects/windows-cloud/global/images/family/windows-2019-core-for-containers",
-		"windows-2019-for-containers":      "projects/windows-cloud/global/images/family/windows-2019-for-containers",
-		"windows-2022":                     "projects/windows-cloud/global/images/family/windows-2022",
-		"windows-2022-core":                "projects/windows-cloud/global/images/family/windows-2022-core",
+	projectMap = map[string]string{
+		"centos":      "centos-cloud",
+		"cos":         "cos-cloud",
+		"debian":      "debian-cloud",
+		"fedora":      "fedora-coreos-cloud",
```
zmarano, line 51:
```suggestion
                "fedora-cloud": "fedora-cloud",
		"fedora-coreos": "fedora-coreos-cloud",
```

imagetest/cmd/manager/main.go
```diff
@@ -44,62 +44,19 @@ var (
...
+		"rhel":        "rhel-cloud",
+		"rhel-sap":    "rhel-sap-cloud",
+		"rocky-linux": "rocky-linux-cloud",
+		"sles":        "suse-cloud",
+		"sles-sap":    "suse-sap-cloud",
```
zmarano, line 56:
```suggestion
		"sles-sap":    "suse-sap-cloud",
		"sql-": "windows-sql-cloud",
```

imagetest/cmd/manager/main.go
```diff
@@ -44,62 +44,19 @@ var (
...
-		"windows-2022":                     "projects/windows-cloud/global/images/family/windows-2022",
-		"windows-2022-core":                "projects/windows-cloud/global/images/family/windows-2022-core",
+	projectMap = map[string]string{
+		"centos":      "centos-cloud",
+		"cos":         "cos-cloud",
+		"debian":      "debian-cloud",
+		"fedora":      "fedora-coreos-cloud",
+		"rhel":        "rhel-cloud",
+		"rhel-sap":    "rhel-sap-cloud",
+		"rocky-linux": "rocky-linux-cloud",
```
zmarano, line 54:
```suggestion
		"rocky-linux": "rocky-linux-cloud",
		"opensuse": "opensuse-cloud",
```

imagetest/cmd/manager/main.go
drewhli: Done.

imagetest/cmd/manager/main.go
drewhli: Done.

imagetest/cmd/manager/main.go
drewhli: Done.

PR #758, opened 2023-08-22
add startup scripts

koln67:
Work on generic startup scripts to install fio before tests run, and work on windows support

### DIFF
--- a/imagetest/test_suites/storage_perf/startupscripts/install_fio.ps1
+++ b/imagetest/test_suites/storage_perf/startupscripts/install_fio.ps1
@@ -1,3 +1,9 @@
 cd C:\
 gsutil cp gs://gce-image-build-resources/windows/fio-3.35-x64.msi C:\
-msiexec.exe /A "C:\fio-3.35-x64.msi" /quiet /norestart
+try {
+	msiexec.exe "C:\fio-3.35-x64.msi" /quiet /norestart
+}
+catch {
+	Write-Output "msi file failed to resolve"
+	Write-Output $_
+}
--- a/imagetest/test_suites/storage_perf/startupscripts/install_fio.ps1
+++ b/imagetest/test_suites/storage_perf/startupscripts/install_fio.ps1
@@ -1,0 +1,3 @@
+cd C:\
+gsutil cp gs://gce-image-build-resources/windows/fio-3.35-x64.msi C:\
+msiexec.exe /A "C:\fio-3.35-x64.msi" /quiet /norestart
--- a/imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
+++ b/imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
@@ -6,4 +6,10 @@ elif [[ -f /bin/yum ]]; then
 	yum -y install fio
 elif [[ -f /usr/bin/zypper ]]; then
 	zypper --non-interactive install fio
+else 
+	echo "No package managers found to install fio"
 fi
+
+errorcode=$?
+if [[ errorcode != 0 ]]; then
+	echo "Error running install fio command: code $errorcode"
--- a/imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
+++ b/imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
@@ -1,16 +1,9 @@
 if [[ -f /usr/bin/apt ]]; then
-	sudo apt -y update && sudo apt -y install fio
-fi
-
-if [[ -f /bin/yum ]]; then
-	sudo yum -y install fio
-fi
-
-if [[ -f /usr/bin/zypper ]]; then
+	apt -y update && apt -y install fio
+elif [[ -f /bin/dnf ]]; then 
+	dnf -y install fio
+elif [[ -f /bin/yum ]]; then
+	yum -y install fio
+elif [[ -f /usr/bin/zypper ]]; then
 	sudo zypper --non-interactive install fio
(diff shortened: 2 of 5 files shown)


imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
```diff
@@ -0,0 +1,16 @@
+if [[ -f /usr/bin/apt ]]; then
```
zmarano, line 1: You won't need sudo in this case (startup scripts run as root).

imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [[ -f /bin/yum ]]; then
+	sudo yum -y install fio
+fi
+
+if [[ -f /usr/bin/zypper ]]; then
+	sudo zypper --non-interactive install fio
+fi
+
+if [[ -f /bin/dnf ]]; then
```
zmarano, line 13: A bit inconsequential but put the dnf before yum (yum in EL8/9 exists as a symlink to dnf).

imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
zmarano, line 1: It may be better to do this with if, elif. In the odd case that more than one of these tools exists (it really shouldn't but let's say they do), one of the calls would fail. The only case that is like this is yum/dnf where there is a symlink to yum on dnf based systems and therefore you'll end up calling the install twice.

koln67: done

imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
koln67: done. also changed it in the golang code to install this, if it becomes necessary again due to script order behavior

imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
koln67: done

imagetest/test_suites/storage_perf/startupscripts/install_fio.ps1
```diff
@@ -0,0 +1,3 @@
+cd C:\
+gsutil cp gs://gce-image-build-resources/windows/fio-3.35-x64.msi C:\
+msiexec.exe /A "C:\fio-3.35-x64.msi" /quiet /norestart
```
jjerger, line 3: Small detail: /A shouldn't be needed since this will run as SYSTEM with administrative privileges.

You may also want to enclose the msiexec inside of a try/catch block, or use /L "file_path" to log to a file and then Get-Content the file to help surface errors.

koln67: done

imagetest/test_suites/storage_perf/startupscripts/install_fio.sh
```diff
@@ -0,0 +1,9 @@
+if [[ -f /usr/bin/apt ]]; then
+	apt -y update && apt -y install fio
+elif [[ -f /bin/dnf ]]; then 
+	dnf -y install fio
+elif [[ -f /bin/yum ]]; then
+	yum -y install fio
+elif [[ -f /usr/bin/zypper ]]; then
+	zypper --non-interactive install fio
+fi
```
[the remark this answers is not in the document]
koln67: done: currently does an echo in the error case

imagetest/test_suites/storage_perf/startupscripts/install_fio.ps1
koln67: it seems to be quite difficult to verify errors here, so I'll plan to use golang code to run these commands and print errors from there.

this script would be updated to be a clone of the golang code once that is done.

jjerger: If the script is going to call a go binary we could handle all the logging and error messages in there and pass it up instead of handling it with powershell.

Alternatively, I believe msiexec will log to Event Viewer even in this context, so if the powershell script were to retrieve the Application Event Log after attempting installation, that might also work, assuming the error in msiexec doesn't block the rest of the script (try/catch would probably be needed for that though. Might be too complex for what it needs to do)

jjerger: You could try something like we did for installing SQL but substituting msiexec:
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/master/daisy_workflows/image_build/sqlserver/sql_install.ps1#L286

jjerger: Just some ideas to consider. Not saying any one of these approaches is more or less correct.

koln67: yeah at least for now, the idea is that in storage_perf_utils.go, there is some code which installs fio on linux using go code.

The startup script install_fio.sh was based on those exact go code lines, and those go lines can print the error message if any command fails.

For windows, the idea would be to have go lines just like what is in storage_perf_utils.go. This is quite useful right now as well, because it seems that some of these windows startup commands might be failing, but it's hard to debug this ps1 script directly. After the go code to install fio on windows succeeds, the ps1 script would be a direct copy of what succeeded.

PR #759, opened 2023-08-24
Update image_boot test to support Windows boot testing

jjerger:
Splits necessary tests into separate functions in order to run the same battery against Linux and Windows

### DIFF
--- a/imagetest/test_suites/image_boot/image_boot_test.go
+++ b/imagetest/test_suites/image_boot/image_boot_test.go
@@ -80,7 +80,100 @@ func lookForSshdAndGuestAgentProcess() error {
 		return nil
 	}
 
-	return fmt.Errorf("not found")
+	return fmt.Errorf("Guest Agent and/or sshd not found.")
+}
+
+func lookForGuestAgentProcessWindows() error {
+	command := `$agentservice = Get-Service GCEAgent
+	$agentservice.Status`
+	output, err := utils.RunPowershellCmd(command)
+	if err != nil {
+		return fmt.Errorf("Failed to find Guest Agent service.")
+	}
+
+	agentStatus := strings.TrimSpace(output.Stdout)
+	if agentStatus == "Running" {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -189,65 +282,36 @@ func TestStartTime(t *testing.T) {
...
-	uptime, err := strconv.ParseFloat(fields[0], 64)
-	if err != nil {
-		t.Fatalf("Failed to read uptime numeric value")
-	}
-	t.Logf("found guest agent and sshd running at %v seconds", int(uptime))
-
-	//Validating the uptime against the allowed threshold value
-	if int(uptime) > maxThreshold {
-		t.Errorf("Boot time too long: %v is beyond max of %v", uptime, maxThreshold)
+	if runtime.GOOS == "windows" {
```
Quintonamore, line 306: This conditional comes up a lot through out these tests. Can we have a helper test function that returns a boolean if the runtime is windows so that there is less of a chance to have a typo?

Quintonamore (approved):

imagetest/test_suites/image_boot/image_boot_test.go
```diff
@@ -80,7 +80,100 @@ func lookForSshdAndGuestAgentProcess() error {
 		return nil
 	}
 
-	return fmt.Errorf("not found")
+	return fmt.Errorf("Guest Agent and/or sshd not found.")
```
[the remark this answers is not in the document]
jjerger: Whoops it looks like it merged these already. But that's a good idea - I'll open a follow up PR to be clearer about which one is missing.

Windows is a weird case though, since it won't be running sshd

imagetest/test_suites/image_boot/image_boot_test.go
jjerger: I'm thinking I'll add an IsWindows() function to our utils because this is going to come up alot more in other test suites, yeah.

PR #760, opened 2023-08-24
Add IsWindows to test utils

jjerger:
Simplifies having to import and run runtime.GOOS in every test suite targeting windows and linux.

### DIFF
--- a/imagetest/utils/test_utils.go
+++ b/imagetest/utils/test_utils.go
@@ -291,10 +291,7 @@ func Skip32BitWindows(t *testing.T, skipMsg string) {
 
 // IsWindows returns true if the detected runtime environment is Windows.
 func IsWindows() bool {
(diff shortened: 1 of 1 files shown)


imagetest/utils/test_utils.go
```diff
@@ -289,6 +289,14 @@ func Skip32BitWindows(t *testing.T, skipMsg string) {
 	}
 }
 
+// IsWindows returns true if the detected runtime environment is Windows.
+func IsWindows() bool {
+	if runtime.GOOS == "windows" {
```
Quintonamore, line 294: Thanks for adding this helper! Since the conditional already resolves to a boolean we can return the conditional directly.

```
return runtime.GOOS == "windows"

Quintonamore (approved):

PR #761, opened 2023-08-25
Rename test suites to avoid underscores

koln67:
Rename tests for consistency. For example, image_boot was renamed to imageboot.


imagetest/README.md
```diff
@@ -57,7 +57,9 @@ JUnit format XML will also be output.
 Tests are organized into go packages in the test\_suites directory and are
 written in go. Each package must at a minimum contain a setup file (by
 conventioned named setup.go) and at least one test file (by convention named
-$packagename\_test.go).
+$packagename\_test.go). The name of the test suite should not contain an
```
a-crate, line 60: This is specifically to make the suite folder name match the package name, right? I think this paragraph should state that.

PR #764, opened 2023-08-28
Add tier1 networking test to network perf test suite

drewhli:


imagetest/test_suites/networkperf/performance_test.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
+func TestEmpty(t *testing.T) {
```
koln67, line 16: Is TestEmpty here so that golang doesn't give the error that the test vms are declared but not used?

drewhli: Yes, exactly. I was getting null exception errors because there weren't any tests to run on the unsupported image VMs.

koln67: you may want to find a way to skip these tests in a follow up PR. For example, I think in the recently added sql test suite, the tests are skipped if the image is not windows.

PR #773, opened 2023-08-30
Create sql test suite

jjerger:
Initial commit of basic sql version testing.  Verifies 3 test cases:

1. Installed Sql Server edition matches the image (i.e. Web/Standard/Entierprise)
2. Installed Sql Server version matches the image (i.e. Sql Server 2022)
3. Installed Windows Server version matches the image (i.e. Windows Server 2022)


jjerger:
Before we merge this, a few things to think about:

- Is this directory location reasonable or should it be inside windows
- I am very bad at variable names. Please recommend alternatives if some don't make sense

PR #775, opened 2023-09-01
write iops

koln67:
Still working on windows fio write: seems like some parameters might be causing an error if the exact same parameters from read are used

### DIFF
--- a/imagetest/test_suites/storageperf/iops_read_test.go
+++ b/imagetest/test_suites/storageperf/iops_read_test.go
@@ -76,7 +76,7 @@ func TestReadIOPS(t *testing.T) {
 
 	var fioOut FIOOutput
 	if err = json.Unmarshal(readIOPSJson, &fioOut); err != nil {
-		t.Fatalf("fio output could not be unmarshalled with error: %v", err)
+		t.Fatalf("fio output %s could not be unmarshalled with error: %v", string(readIOPSJson), err)
 	}
 
 	finalIOPSValue := fioOut.Jobs[0].ReadResult.IOPS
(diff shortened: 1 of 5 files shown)


imagetest/test_suites/storageperf/iops_read_test.go
```diff
@@ -17,15 +17,11 @@ import (
 
 const (
 	commonFIOReadOptions = "--name=read_iops_test --filesize=10G --numjobs=8 --time_based --runtime=60s --ramp_time=2s --direct=1 --verify=0 --bs=4K --iodepth=256 --randrepeat=0 --rw=randread --group_reporting=1 --iodepth_batch_submit=256  --iodepth_batch_complete_max=256 --output-format=json"
-	windowsDriveLetter   = "F"
 )
 
 func RunFIOReadWindows() ([]byte, error) {
 	testdiskDrive := windowsDriveLetter + ":\\"
```
jjerger, line 23: I think windowsDriveLetter is undefined after being removed in one of these commits

imagetest/test_suites/storageperf/setup.go
```diff
@@ -34,6 +34,7 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 
 	vm.AddMetadata("enable-guest-attributes", "TRUE")
 	if strings.Contains(t.Image, "windows") {
+		vm.AddMetadata("windowsDriveLetter", windowsDriveLetter)
```
jjerger, line 37: I think windowsDriveLetter is undefined (was it moved from somewhere else?)

jjerger: Nvm just saw it in utils

imagetest/test_suites/storageperf/iops_read_test.go
jjerger: Nvm saw it in utils

jjerger (approved):
/lgtm

PR #777, opened 2023-09-07
Update networkperf test suite to support windows images

drewhli:
It's weird how the network performance test results don't reach anywhere close to the targets, especially in the non-Tier1 network tests (specifically for Windows)

### DIFF
--- a/imagetest/test_suites/networkperf/performance_test.go
+++ b/imagetest/test_suites/networkperf/performance_test.go
@@ -5,7 +5,6 @@ package networkperf
 
 import (
 	"encoding/json"
-	"runtime"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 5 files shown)


drewhli:
Seems like mostly a windows2016 issue.

imagetest/test_suites/networkperf/performance_test.go
koln67, line 1: TestEmpty should be removed

imagetest/test_suites/networkperf/performance_test.go
```diff
@@ -19,45 +18,43 @@ func TestEmpty(t *testing.T) {
```
koln67, line 18: TestEmpty should be removed

drewhli: Done.

PR #782, opened 2023-09-12
Add shapevalidation test

a-crate:
* Test that the expected amount of memory, number of CPUs, and number of NUMA nodes appear

* Exclude this test from existing CIT pipelines that trigger frequently

* Add a new release pipeline that will run all tests on an image, triggered manually

* Add API for tests to force a specific zone and VM shape when it is required for a test, and log when they do so

* Test new API functions


koln67:
Unrelated to merging this PR, do we want the shape validation tests to run on a 12 hour interval, or would it potentially use more quota than we want?

If so, we might want to exclude it from the cit-periodics dashboard here in a different PR:
https://github.com/GoogleCloudPlatform/oss-test-infra/blob/2f7280c9d7d4a3cf96c156fa60a52011188d3454/prow/prowjobs/GoogleCloudPlatform/gcp-guest/gcp-guest-config.yaml#L412

a-crate:
Definitely not, that would be too much quota and isn't likely to change once it's passing. I have a change ready to exclude that.

PR #790, opened 2023-09-14
Fix nil error in build

koln67:
It seems like the order of checking the steps can result in a nil pointer in the build, if machineType is not set.

### DIFF
--- a/imagetest/testworkflow.go
+++ b/imagetest/testworkflow.go
@@ -355,12 +355,12 @@ func finalizeWorkflows(ctx context.Context, tests []*TestWorkflow, zone, gcsPref
 
 		twf.wf.Zone = zone
 
-		createVMsStep := twf.wf.Steps[createVMsStepName]
-		for _, vm := range createVMsStep.CreateInstances.Instances {
-			if vm.Zone != "" && vm.Zone != twf.wf.Zone {
-				log.Printf("VM %s zone is set to %s, differing from workflow zone %s for test %s, not overriding\n", vm.Name, vm.Zone, twf.wf.Zone, twf.Name)
-			}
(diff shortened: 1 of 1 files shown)


imagetest/testworkflow.go
```diff
@@ -355,12 +355,12 @@ func finalizeWorkflows(ctx context.Context, tests []*TestWorkflow, zone, gcsPref
 
 		twf.wf.Zone = zone
 
-		createVMsStep := twf.wf.Steps[createVMsStepName]
-		for _, vm := range createVMsStep.CreateInstances.Instances {
-			if vm.Zone != "" && vm.Zone != twf.wf.Zone {
-				log.Printf("VM %s zone is set to %s, differing from workflow zone %s for test %s, not overriding\n", vm.Name, vm.Zone, twf.wf.Zone, twf.Name)
-			}
-			if machineType != "" {
+		if machineType != "" {
```
a-crate, line 358: This will avoid warning about mismatched zones unless the machine_type flag is set, and could still panic if the machine_type flag was set when createVMsStep is nil

koln67: When support for machinetypes was introduced, it seems that the createVMsStep was not nil as long as machinetype was set, so I don't think that should be a major concern here as it was not broken in the past.

Another possible fix would be to investigate which calls in the networkperf test suite are causing createVmsStep to be nil @drewhli 

https://github.com/GoogleCloudPlatform/guest-test-infra/pull/390

koln67: fixed so that the only change is to check if createVMsStep is nil before deferencing

the other logic should be the same as before for logging the zone and machine type

PR #791, opened 2023-09-15
Benchmark

koln67:
Hard code the expected performance values, and compare the iops that we get from fio to the hard coded expected performance

### DIFF
--- a/imagetest/test_suites/hotattach/hot_attach_test.go
+++ b/imagetest/test_suites/hotattach/hot_attach_test.go
@@ -173,10 +173,7 @@ func TestFileHotAttach(t *testing.T) {
 	}
 	// mount again, then read from the file
 	if runtime.GOOS == "windows" {
-		procStatus, err := utils.RunPowershellCmd("Initialize-Disk -PartitionStyle GPT -Number 1 -PassThru | New-Partition -DriveLetter " + windowsMountDriveLetter + " -UseMaximumSize | Format-Volume -FileSystem NTFS -NewFileSystemLabel 'Perf-Test' -Confirm:$false")
-		if err != nil {
-			t.Fatalf("failed to initialize disk on reattach for windows: errors %v %s %s", err, procStatus.Stdout, procStatus.Stderr)
-		}
+		t.Log("windows disk was successfully reattached")
 	} else {
 		if err := mountLinuxDiskToPath(linuxMountPath, true); err != nil {
 			t.Fatalf("failed to mount linux disk to path %s on reattach: error %v", linuxMountPath, err)
--- a/imagetest/test_suites/hotattach/hot_attach_test.go
+++ b/imagetest/test_suites/hotattach/hot_attach_test.go
@@ -41,9 +41,9 @@ func getProjectAndZone() (string, string, error) {
 	return projectZoneSlice[1], projectZoneSlice[3], nil
 }
 
-func getLinuxMountPath(mountDiskSizeGb int, mountDiskName string) (string, error) {
+func getLinuxMountPath(mountDiskSizeGB int, mountDiskName string) (string, error) {
 	symlinkRealPath := ""
-	diskPartition, err := utils.GetMountDiskPartition(mountDiskSizeGb)
+	diskPartition, err := utils.GetMountDiskPartition(mountDiskSizeGB * bytesInGB)
 	if err == nil {
 		symlinkRealPath = "/dev/" + diskPartition
 	} else {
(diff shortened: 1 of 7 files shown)


imagetest/test_suites/storageperf/iops_write_test.go
```diff
@@ -92,12 +92,21 @@ func TestRandomWriteIOPS(t *testing.T) {
...
+	}
+
+	expectedRandWriteIOPSString = strings.TrimSpace(expectedRandWriteIOPSString)
+	var expectedRandWriteIOPS float64
+	if expectedRandWriteIOPS, err := strconv.ParseFloat(expectedRandWriteIOPSString, 64); err != nil {
+		t.Fatalf("benchmark iops string %f was not a float: err %v", expectedRandWriteIOPS, err)
 	}
-	t.Logf("iops test pass with %f iops, expected at least %f", finalIOPSValue, expectedHyperdiskIOPS)
+	if finalIOPSValue < iopsErrorMargin*expectedRandWriteIOPS {
+		t.Fatalf("iops average was too low: expected close to %f, got  %f", expectedRandWriteIOPS, finalIOPSValue)
```
a-crate, line 106: Might be worth printing the calculated minimum or the margin here so anyone debugging the test doesn't have to guess what the actual target is

imagetest/test_suites/storageperf/iops_write_test.go
```diff
@@ -92,12 +92,21 @@ func TestRandomWriteIOPS(t *testing.T) {
...
+	}
+
+	t.Logf("iops test pass with %f iops, expected at least %f", finalIOPSValue, expectedRandWriteIOPS)
```
a-crate, line 109: This might be a confusing pass message, finalIOPSValue could be less than expectedRandWriteIOPS but still pass and print this. Printing the calculated actual minimum or printing "within margin %d of %s iops" would be more clear.

imagetest/test_suites/storageperf/iops_write_test.go
```diff
@@ -118,14 +127,21 @@ func TestSequentialWriteIOPS(t *testing.T) {
...
+	var expectedSeqWriteIOPS float64
+	if expectedSeqWriteIOPS, err := strconv.ParseFloat(expectedSeqWriteIOPSString, 64); err != nil {
+		t.Fatalf("benchmark iops string %f was not a float: err %v", expectedSeqWriteIOPS, err)
 	}
-	t.Logf("iops test pass with %f iops, expected at least %f", finalIOPSValue, expectedHyperdiskIOPS)
+	if finalIOPSValue < iopsErrorMargin*expectedSeqWriteIOPS {
+		t.Fatalf("iops average was too low: expected close to %f, got  %f", expectedSeqWriteIOPS, finalIOPSValue)
+	}
+
+	t.Logf("iops test pass with %f iops, expected at least %f", finalIOPSValue, expectedSeqWriteIOPS)
```
a-crate, line 146: Same comments as random write pass message here.

imagetest/test_suites/hotattach/hot_attach_test.go
```diff
@@ -173,10 +173,7 @@ func TestFileHotAttach(t *testing.T) {
 	}
 	// mount again, then read from the file
 	if runtime.GOOS == "windows" {
-		procStatus, err := utils.RunPowershellCmd("Initialize-Disk -PartitionStyle GPT -Number 1 -PassThru | New-Partition -DriveLetter " + windowsMountDriveLetter + " -UseMaximumSize | Format-Volume -FileSystem NTFS -NewFileSystemLabel 'Perf-Test' -Confirm:$false")
-		if err != nil {
-			t.Fatalf("failed to initialize disk on reattach for windows: errors %v %s %s", err, procStatus.Stdout, procStatus.Stderr)
-		}
+		t.Log("windows disk was successfully reattached")
```
jjerger, line 176: Is this just skipping the Windows test for now?

imagetest/test_suites/storageperf/iops_write_test.go
koln67: Fixed so that it prints what percent of the target was expected

imagetest/test_suites/storageperf/iops_write_test.go
koln67: Changed so that it prints what decimal factor (0.95 in this case) of the expected value is required to pass the test

imagetest/test_suites/storageperf/iops_write_test.go
koln67: same fix as previous comments

imagetest/test_suites/hotattach/hot_attach_test.go
koln67: It's now passing the windows test. It turns out that after first initializing the disk, then detaching and reattaching, the powershell command to initialize the disk again is not necessary because it remembers that the disk was initialized.

This test does verify that the text file can still be read after hot attach.

PR #800, opened 2023-09-21
Refactor Testworkflow structure...

a-crate:
...to provide a compute.Image and a compute.Service instead of image name

Use these for determining architecture and supported OS features instead of curated lists of names when possible

Retain the partial URL attribute (renamed) as this is needed by daisy and not readily available anywhere in the compute.Image object


a-crate:
/cc @zmarano 
Zach could you take a look at this? I'm not looking for LGTM yet, it has gofmt problems, needs to check GVNIC in more places, and I want a more comprehensive CIT presubmit before I make large changes. 

I just want to get feedback on whether this is a good direction/too large to change without discussion first/etc.
I'm trying to make the Right Thing mentioned in this [PR](https://github.com/GoogleCloudPlatform/guest-test-infra/pull/390#discussion_r816333901) a long time ago easier to do.

zmarano:
Generally I think this is a great change. It also facilitates image families. It is worth updating the README or even adding a section on the ways to reference and use images, features, and architectures: https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/imagetest/README.md

a-crate:
/hold
This is ready for review and it's generating workflows fine but I want to make sure they are running to completion before removing the hold.

PR #802, opened 2023-09-22
Skip the c3d family during storageperf testing on windows 2012 and 2016

a-crate:
Timing out on testgrid: https://testgrid.k8s.io/googleoss-gcp-guest#cit-periodics-performance

Not 100% sure I skipped this correctly, lmk if this is a problem.

### DIFF
--- a/imagetest/test_suites/storageperf/setup.go
+++ b/imagetest/test_suites/storageperf/setup.go
@@ -37,6 +37,9 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		if machineTypeParam, foundKey := paramMap["machineType"]; foundKey {
 			machineType = machineTypeParam
 		}
+		if strings.HasPrefix(machineType, "c3d") && (strings.Contains(t.Image, "windows-2012") || strings.Contains(t.Image, "windows-2016")) {
+			continue
+		}
 		bootDisk := compute.Disk{Name: vmName + machineType, Type: imagetest.PdBalanced, SizeGb: bootdiskSizeGB}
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/storageperf/setup.go
```diff
@@ -37,6 +37,9 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		if machineTypeParam, foundKey := paramMap["machineType"]; foundKey {
 			machineType = machineTypeParam
 		}
+		if strings.HasPrefix(machineType, "c3d") && (strings.Contains(t.Image, "windows-2012") || strings.Contains(t.Image, "windows-2016")) {
+			continue
```
zmarano, line 41: There's a skip function so that they show up correctly as skipped tests similar to
https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/imagetest/test_suites/cvm/setup.go#L18

However.... would doing it this way just ignore running this test on these machine types? If so, I think it is better- because it won't instrument the VM at all. Correct me if I'm wrong about that.

a-crate: My goal was not to skip storageperf entirely on windows 2012/2016 but to skip instrumenting the C3D vm and only check storage performance on the other two families (c3/n2).

Continuing the loop like this will avoid creating the CreateVMs and RunTests daisy steps in the final workflow, so should avoid instrumenting the VM all the same.

zmarano: Cool- then I think this makes sense.

PR #805, opened 2023-09-23
Add CIT presubmit container

a-crate:
There are multiple things wrong with the way presubmits are running on CIT:

1) The globbing on the go cli is package globbing not file globbing, so it doesn't pick up the imagetest folder in gobuild because it has it's own go.mod file

2) Even if it did run the logic in those cases it wouldn't make a lot of sense because the "test code" in CIT isn't really test code in the usual sense so running `go test` on it is incorrect, it just needs to build instead of pass

3) The test code that is typical test code (testworkflow_test.go, etc). isn't being run during gotest because of issue 1.

The only thing that does get picked up is gofmt, because that uses `find . -name *.go` to select files instead of the go cli globbing.

### DIFF
--- a/concourse/pipelines/container-build.jsonnet
+++ b/concourse/pipelines/container-build.jsonnet
@@ -121,6 +121,7 @@ local BuildContainerImage(image) = buildcontainerimgjob {
     BuildContainerImage('jsonnet-go'),
     BuildContainerImage('fly-validate-pipelines') { passed: 'build-jsonnet-go' },
     BuildContainerImage('pytest'),
+    BuildContainerImage('citpresubmit'),
 
     // Non-standard dockerfile location and public image.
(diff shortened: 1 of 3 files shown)


a-crate:
I went back and forth trying to make gobuild and gotest more universal but decided to implement a CIT specific container because the compiling has a lot more steps, takes longer, and requires a bunch of cit-specific logic to handle it's "test" code that doesn't apply to anything else in the repo. I'm open to feedback if you think this is not the right approach.

koln67:
this would be an improvement on the presubmit coverage. One minor point is that building the entire dockerfile for all the OSes would make the presubmit take around 30 minutes, which is quite long for a presubmit.

I think we can push this change through and revisit it if the presubmit takes too long.

PR #810, opened 2023-09-27
Install all dependencies for dockerfile

koln67:
Make the citpresubmit Dockerfile match the one for gotest, in case some dependency prevents go-junit-report from being installed


a-crate:
I think the latest failed presubmits are just using a stale container, concourse has it rebuilt now. Can't hurt to do this though.
/lgtm

PR #816, opened 2023-09-28
Windows write test fix

koln67:
Do not initialize the mount disk on windows. This fixes the fio write permissions issue.


jjerger (approved):
/lgtm

PR #817, opened 2023-09-28
Pass down machine types through TestWorkflow

drewhli:
This PR also
- Does some minor cleanup in `fixtures.go` regarding `CreateNetwork`, making the MTU setting a separate method instead of a copy-paste of `CreateNetwork` with one extra parameter.
- Refactors `networkperf` test suite to take advantage of the now-available machine type field. We pass the target number directly instead of a large map string.
  - Also use the `MachineType` to prevent `Tier_1` tests from running if the machine type is not supported.


a-crate:
This looks great, much cleaner way to handle things.
/lgtm

PR #819, opened 2023-09-29
Add guest attributes

koln67:
This change should CIT tests to pass if either serial output matches or the guest attribute is set.

### DIFF
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -17,10 +17,14 @@ import (
 )
 
 func main() {
-	// These are placeholders until daisy supports guest attributes.
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("failed to create cloud storage client: %v", err)
+	}
 	log.Printf("FINISHED-BOOTING")
 	defer func() {
-		if err := utils.PutMetadataGuestAttribute(utils.GuestAttributeTestNamespace, utils.GuestAttributeTestKey); err != nil {
+		if err := utils.PutMetadataGuestAttribute(ctx, utils.GuestAttributeTestNamespace, utils.GuestAttributeTestKey); err != nil {
 			log.Printf("failed to put test completed key in guest attribute namespace")
 		}
 		for f := 0; f < 5; f++ {
@@ -28,12 +32,6 @@ func main() {
 			time.Sleep(1 * time.Second)
 		}
 	}()
-
-	ctx := context.Background()
-	client, err := storage.NewClient(ctx)
-	if err != nil {
-		log.Fatalf("failed to create cloud storage client: %v", err)
-	}
 
 	daisyOutsPath, err := utils.GetMetadataAttribute("daisy-outs-path")
 	if err != nil {
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -27,6 +27,7 @@ func main() {
 		if err := utils.PutMetadataGuestAttribute(ctx, utils.GuestAttributeTestNamespace, utils.GuestAttributeTestKey); err != nil {
 			log.Printf("failed to put test completed key in guest attribute namespace")
 		}
+		log.Printf("successfully placed guest attribute for test completion")
 		for f := 0; f < 5; f++ {
 			log.Printf("FINISHED-TEST")
 			time.Sleep(1 * time.Second)
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -20,6 +20,9 @@ func main() {
 	// These are placeholders until daisy supports guest attributes.
 	log.Printf("FINISHED-BOOTING")
 	defer func() {
+		if err := utils.PutMetadataGuestAttribute(utils.GuestAttributeTestNamespace, utils.GuestAttributeTestKey); err != nil {
+			log.Printf("failed to put test completed key in guest attribute namespace")
+		}
 		for f := 0; f < 5; f++ {
 			log.Printf("FINISHED-TEST")
 			time.Sleep(1 * time.Second)
--- a/imagetest/testworkflow.go
+++ b/imagetest/testworkflow.go
@@ -216,9 +216,14 @@ func (t *TestWorkflow) addWaitStep(stepname, vmname string, stopped bool) (*dais
 	instanceSignal.Name = vmname
 	instanceSignal.Stopped = stopped
 
+	guestAttribute := &daisy.GuestAttribute{}
+	guestAttribute.Namespace = utils.GuestAttributeTestNamespace
+	guestAttribute.KeyName = utils.GuestAttributeTestKey
+
 	// Waiting for stop and waiting for success match are mutually exclusive.
 	if !stopped {
 		instanceSignal.SerialOutput = serialOutput
+		instanceSignal.GuestAttribute = guestAttribute
 	}
 
 	waitForInstances := &daisy.WaitForInstancesSignal{instanceSignal}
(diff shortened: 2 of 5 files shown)


imagetest/utils/test_utils.go
```diff
@@ -105,6 +109,38 @@ func GetMetadataHTTPResponse(path string) (*http.Response, error) {
...
+	return PutMetadata("guest-attributes/" + namespace + "/" + attribute)
+}
+
+// PutMetadata returns a metadata value for the specified key if it is present, and error if not.
+func PutMetadata(path string) error {
+	resp, err := PutMetadataHTTPResponse(path)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
```
dorileo, line 123: Can you please handle `resp.StatusCode` in `PutMuetadataHTTPResponse()` instead.

imagetest/utils/test_utils.go
```diff
@@ -105,6 +109,38 @@ func GetMetadataHTTPResponse(path string) (*http.Response, error) {
 	return resp, nil
 }
 
+// PutMetadataGuestAttribute sets the guest attribute in the namespace, and returns an error if this operation fails.
+func PutMetadataGuestAttribute(namespace, attribute string) error {
+	return PutMetadata("guest-attributes/" + namespace + "/" + attribute)
+}
+
+// PutMetadata returns a metadata value for the specified key if it is present, and error if not.
+func PutMetadata(path string) error {
```
dorileo, line 118: if you handle `resp.StatusCode` in `PutMuetadataHTTPResponse()` you won't even need this wrapper.

imagetest/utils/test_utils.go
```diff
@@ -105,6 +109,38 @@ func GetMetadataHTTPResponse(path string) (*http.Response, error) {
...
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("http response code is %v", resp.StatusCode)
+	}
+	return nil
+}
+
+// PutMetadataHTTPResponse returns http response for the specified key without checking status code.
+func PutMetadataHTTPResponse(path string) (*http.Response, error) {
```
dorileo, line 130: do we have a context we could use here?

dorileo: Huum actually we do, in the wrapper's `main()`, we have this:

```
	ctx := context.Background()
	client, err := storage.NewClient(ctx)
```

Please change this `PutMetadataHTTPResponse()` function and pass down the context.

No need to address it in this PR but we are not handling signals in the wrapper, ideally we should, the sig handler would have to cancel the context and nicely terminate the program - even if we don't expect to have the program being terminated or aborted it's a good practice.

koln67: currently passed it down, bug created

imagetest/utils/test_utils.go
koln67: done

imagetest/utils/test_utils.go
koln67: done

imagetest/cmd/wrapper/main.go
```diff
@@ -17,21 +17,22 @@ import (
 )
 
 func main() {
-	// These are placeholders until daisy supports guest attributes.
```
a-crate, line 20: If the placeholder comment is no longer accurate can we just delete all of the placeholder stuff and start relying solely on guest attributes to signal that testing is finished?

imagetest/testworkflow.go
```diff
@@ -216,9 +216,14 @@ func (t *TestWorkflow) addWaitStep(stepname, vmname string, stopped bool) (*dais
...
 	instanceSignal.Stopped = stopped
 
+	guestAttribute := &daisy.GuestAttribute{}
+	guestAttribute.Namespace = utils.GuestAttributeTestNamespace
+	guestAttribute.KeyName = utils.GuestAttributeTestKey
+
 	// Waiting for stop and waiting for success match are mutually exclusive.
 	if !stopped {
 		instanceSignal.SerialOutput = serialOutput
+		instanceSignal.GuestAttribute = guestAttribute
```
a-crate, line 226: This will wait for both the stopped signal and the test-complete key when stopped is true. I don't think this is the right pattern of behavior, when `addWaitStep` is called in fixtures.go with `stopped=true` it's called as a part of rebooting where the instance stopping is the critical signal rather than the tests being finished.
Maybe this should be broken out into two different functions with clearer meanings (ie wait for a stopped VM vs wait for tests to be finished).

koln67: The new change has a waitstep and a waitstoppedstep, where the SerialOutput and GuestAttrbute are only checked in the waitstep case.

It appears that if both are set, the wait step finishes if either the serial log or guest attribute matches.

imagetest/cmd/wrapper/main.go
koln67: I'm not sure what the original author of this comment had in mind with the serial log being the placeholder.

My idea for now was to have both guest attributes and the serial log be a condition for finishing the test, so that in corner cases where guest attributes might fail (for example if we need to manually set guest attributes to be true on a VM, though this doesn't seem to be an issue), the tests can continue to finish with the serial log.

After a few daily test runs of CIT, if guest attributes end up being as reliable as we hope they are, we can remove the serial log match.

imagetest/utils/test_utils.go
```diff
@@ -105,6 +109,34 @@ func GetMetadataHTTPResponse(path string) (*http.Response, error) {
 	return resp, nil
 }
 
+// PutMetadataGuestAttribute sets the guest attribute in the namespace, and returns an error if this operation fails.
+func PutMetadataGuestAttribute(ctx context.Context, namespace, attribute string) error {
+	path := "guest-attributes/" + namespace + "/" + attribute
```
dorileo, line 114: Can you use [url.JoinPath()](https://pkg.go.dev/net/url#JoinPath) instead?

imagetest/cmd/wrapper/main.go
dorileo: I'm ok keeping both methods for a while as long as we keep tracking it and clean it up as soon as we have enough data to make a decision.

imagetest/utils/test_utils.go
koln67: done

imagetest/cmd/wrapper/main.go
koln67: cool, I created a bug for tracking this as well

imagetest/testworkflow.go
a-crate: Ok so it will still work for reboots, but then it won't correctly signal when tests are finished, right?

a-crate: Nevermind I see you fixed it in the latest commit.

koln67: i haven't tested all the suites which call reboot(), though when I ran the imageboot suite, which calls reboot(), the test suite completed successfully.

imagetest/cmd/wrapper/main.go
a-crate: Will we be able to see that data? Daisy doesn't prefer one signal method, so we might see a signal match from serial logs even though guest attributes would have worked depending on the race.

koln67: I think for the first few days, not seeing any tests break from this change should be good enough.

I've added an additional log statement when the guest attribute is placed from the HTTP request, so we can confirm in the logs that this succeeded.

dorileo: We are not setting interval for metadata signal so it will be min interval = 6s, if we are not setting interval for serial then it will be 10s meaning metadata signal will always happen first.

dorileo: But we can disable serial signal and watch, then remove it entirely if when we are confident it's working as expected.

PR #823, opened 2023-10-03
Add wait for vm quota test workflow step, and functions to append quotas

a-crate:
During testing I ran into a lot of problems with finding free quota that was taken by the time vm creation started. Some of this is unavoidable because daisy takes a second to switch between steps, but I mitigated the problem by staggering parallel workers. Most of this can be avoided by just waiting 60-90 seconds after starting a workflow.

The other thing I did was break this up from one large wait-for-quota step that all creation steps depend on, to quota steps for each create step. But the only create step that actually needs enough quota to worry about right now is the create VMs step, so I just took the others out avoid to avoiding leaving in unused code.

If it's not clear enough that the existing API is only for quota that gates VM creation I can add some comments expanding on this in the code.

### DIFF
--- a/concourse/pipelines/linux-image-build.jsonnet
+++ b/concourse/pipelines/linux-image-build.jsonnet
@@ -59,7 +59,7 @@ local prepublishtesttask = {
       '-test_projects=compute-image-test-pool-005',
       // Run tests not ran in publish-to-testing
       // TODO enable oslogin
-      '-filter=(shapevalidation)|(hotattach)',
+      '-filter=(shapevalidation)',
       '-images=' + task.images,
     ] + task.extra_args,
   },
@@ -319,6 +319,10 @@ local imgpublishjob = {
             },
             attempts: 3,
           }
+          else
+          {
+          },
+          if tl.env == 'prod' then
           // Prod releases use a different final publish step that invokes ARLE.
           {
             task: 'publish-' + tl.image,


imagetest/test_suites/shapevalidation/setup.go
```diff
@@ -31,6 +34,7 @@ var x86shapes = map[string]*shape{
 		numa:  4,
 		disks: []*compute.Disk{{Name: "C3", Type: imagetest.PdBalanced, Zone: "us-east1-b"}},
 		zone:  "us-east1-b",
+		quota: []*daisy.QuotaAvailable{&daisy.QuotaAvailable{Units: 176.0, Metric: "C3_CPUS", Region: "us-east1"}},
```
dorileo, line 37: Do we want to use a hard coded region?

imagetest/test_suites/shapevalidation/setup.go
```diff
@@ -40,41 +44,47 @@ var x86shapes = map[string]*shape{
 		disks:      []*compute.Disk{{Name: "C3D", Type: imagetest.PdBalanced, Zone: "us-east4-c"}},
 		zone:       "us-east4-c",
 		exceptions: []*regexp.Regexp{regexp.MustCompile("windows")},
+		quota:      []*daisy.QuotaAvailable{&daisy.QuotaAvailable{Units: 360.0, Metric: "CPUS", Region: "us-east4"}}, // C3D metric not available over public API yet but it's such a large chunk of our usage in us-east4 that this should work for now
```
dorileo, line 47: Same here.

imagetest/test_suites/shapevalidation/setup.go
```diff
@@ -86,6 +96,7 @@ var armshapes = map[string]*shape{
 		numa:  1,
 		disks: []*compute.Disk{{Name: "T2A", Type: imagetest.PdStandard, Zone: "us-central1-a"}},
 		zone:  "us-central1-a",
+		quota: []*daisy.QuotaAvailable{&daisy.QuotaAvailable{Units: 48.0, Metric: "T2A_CPUS", Region: "us-central1"}},
```
dorileo, line 99: same (hard coded region).

imagetest/test_suites/shapevalidation/setup.go
dorileo: I don't know if we currently have flags for it though.

a-crate: We are already hard-coding the region, this is just adding the quota code. Hard-coding the region is not a perfect solution, but it's necessary for now to be able to run tests on VMs that are only available in certain zones. There is no region with enough quota to support every machine type we need for these suites so they will have to diverge somehow.

A better solution would be the ability to auto-determine region based on the requirements of the workflow but this would required a pretty big refactor to incorporate the compute API more deeply into the workflow. We mostly parse strings for information right now.

VMs without hard coded regions still respect the flag passed on the command line.

imagetest/testworkflow.go
```diff
@@ -566,11 +597,17 @@ func RunTests(ctx context.Context, storageClient *storage.Client, testWorkflows
...
+	if err != nil {
+		return nil, err
+	}
+
 	var wg sync.WaitGroup
 	for i := 0; i < parallelCount; i++ {
 		wg.Add(1)
 		go func(id int) {
 			defer wg.Done()
+			time.Sleep(time.Duration(int64(id) * int64(stagger)))
```
dorileo, line 610: why not `time.Sleep(stagger * time.Duration(i))` ?

a-crate:
Force pushed wrong branch.

PR #826, opened 2023-10-04
Add back guest attributes with fix for metadata shutdown test

koln67:
Previously, guest attributes were leading to the metadata shutdown script test failing due to reboot behaviors. This should fix the issue.

### DIFF
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -23,7 +23,7 @@ func main() {
 		log.Fatalf("failed to create cloud storage client: %v", err)
 	}
 	log.Printf("FINISHED-BOOTING")
-	defer func() {
+	defer func(ctx context.Context) {
 		if err := utils.PutMetadataGuestAttribute(ctx, utils.GuestAttributeTestNamespace, utils.GuestAttributeTestKey); err != nil {
 			log.Printf("failed to put test completed key in guest attribute namespace")
 		}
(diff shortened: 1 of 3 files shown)


dorileo:
I'm holding this PR until we have releases done this week.

koln67:
It seems that the issue was not the interval, but that the guest attribute stayed active on reboot.

If a test needed to run after the reboot finished, it could not do that, as the daisy workflow would immediately end. This is because the guest attribute was already set on the previous boot.

imagetest/cmd/wrapper/main.go
```diff
@@ -23,12 +24,21 @@ func main() {
...
+		if err := utils.QueryMetadataGuestAttribute(ctx, utils.GuestAttributeTestNamespace, utils.GuestAttributeTestKey, http.MethodPut); err != nil {
+			log.Printf("failed to put test completed key in guest attribute namespace")
+		}
+		log.Printf("successfully placed guest attribute for test completion")
 		for f := 0; f < 5; f++ {
 			log.Printf("FINISHED-TEST")
 			time.Sleep(1 * time.Second)
 		}
-	}()
+		time.Sleep(10 * time.Second)
```
dorileo, line 36: I understand why this is needed here, however I believe the actual fix should go in the metadata script runner - we can further discuss it offline.

dorileo: We can get rid of it when we have a fix landed to metadata script runner.

PR #827, opened 2023-10-04
OsLogin Tests

drewhli:
Also move from using the deprecated `ioutil` to using the `os` package for reading files.

The tests include
- Checking 2FA enabled and appropriate headers related to turning 2FA on.
- SSH-ing with OSLogin enabled works as intended
- TODO: SSH-ing with OSLogin 2FA enabled works as intended.

### DIFF
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -377,6 +377,11 @@ func (t *TestVM) AddMetadata(key, value string) {
 	}
 }
 
+// AddScope adds the specified auth scope to the service account on the VM.
+func (t *TestVM) AddScope(scope string) {
+	t.instance.Scopes = append(t.instance.Scopes, scope)
+}
+
 // RunTests runs only the named tests on the testVM.
 //
 // From go help test:
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -379,7 +379,11 @@ func (t *TestVM) AddMetadata(key, value string) {
 
 // AddScope adds the specified auth scope to the service account on the VM.
 func (t *TestVM) AddScope(scope string) {
-	t.instance.Scopes = append(t.instance.Scopes, scope)
+	if t.instance != nil {
+		t.instance.Scopes = append(t.instance.Scopes, scope)
+	} else if t.instancebeta != nil {
+		t.instancebeta.Scopes = append(t.instancebeta.Scopes, scope)
+	}
 }
 
 // RunTests runs only the named tests on the testVM.
--- a/imagetest/test_suites/oslogin/oslogin_test.go
+++ b/imagetest/test_suites/oslogin/oslogin_test.go
@@ -4,127 +4,18 @@
 package oslogin
 
 import (
-	"context"
-	"errors"
-	"fmt"
 	"os"
 	"os/exec"
-	"strconv"
 	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
-// const testUsername = "sa_105020877179577573373"
-// const testUUID = "3651018652"
-const (
-	testUsername = "sa_115308896887453382826"
-	testUUID     = "2260770985"
-)
-
-var testUserEntry = fmt.Sprintf("%s:*:%s:%s::/home/%s:", testUsername, testUUID, testUUID, testUsername)
-
-func getTwoFactorAuth(ctx context.Context, root string, elem ...string) (bool, error) {
-	data, err := utils.GetMetadata(ctx, append([]string{root}, elem...)...)
-	if err != nil {
-		return false, err
-	}
-
-	flag, err := strconv.ParseBool(data)
-	if err != nil {
-		return false, fmt.Errorf("failed to parse enable-oslogin-2fa metadata entry: %+v", err)
-	}
-
-	return flag, nil
-}
-
-func isTwoFactorAuthEnabled(ctx context.Context) (bool, error) {
-	var (
-		instanceFlag, projectFlag bool
-		err                       error
-	)
-
-	elem := []string{"attributes", "enable-oslogin-2fa"}
-
-	instanceFlag, err = getTwoFactorAuth(ctx, "instance", elem...)
-	if err != nil && !errors.Is(err, utils.ErrMDSEntryNotFound) {
-		return false, err
-	}
-
-	projectFlag, err = getTwoFactorAuth(ctx, "project", elem...)
-	if err != nil && !errors.Is(err, utils.ErrMDSEntryNotFound) {
-		return false, err
-	}
-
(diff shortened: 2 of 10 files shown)


imagetest/fixtures.go
```diff
@@ -377,6 +377,11 @@ func (t *TestVM) AddMetadata(key, value string) {
 	}
 }
 
+// AddScope adds the specified auth scope to the service account on the VM.
+func (t *TestVM) AddScope(scope string) {
+	t.instance.Scopes = append(t.instance.Scopes, scope)
```
a-crate, line 382: Can you add cases for if `t.instancebeta != nil` in in addition to `t.instance`

imagetest/test_suites/oslogin/oslogin_test.go
```diff
@@ -16,8 +16,12 @@ import (
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
-const testUsername = "sa_105020877179577573373"
-const testUUID = "3651018652"
+// const testUsername = "sa_105020877179577573373"
```
a-crate, line 19: Comments are a duplicate of the code below them. Is there any non-corp documentation for where these came from that we can link to publicly?

imagetest/test_suites/oslogin/setup.go
```diff
@@ -7,20 +7,36 @@ import (
 // Name is the name of the test package. It must match the directory name.
 var Name = "oslogin"
 
+const (
+	computeScope  = "https://www.googleapis.com/auth/compute"
+	platformScope = "https://www.googleapis.com/auth/cloud-platform"
+)
+
 // TestSetup sets up the test workflow.
 func TestSetup(t *imagetest.TestWorkflow) error {
```
a-crate, line 16: Should this be skipping windows images? Seems like it.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -0,0 +1,322 @@
...
+	session, err := client.NewSession()
+	if err != nil {
+		return fmt.Errorf("failed to create ssh session: %v", err)
+	}
+	data, err := session.Output("cat /etc/nsswitch.conf")
+	if err != nil {
+		return fmt.Errorf("failed to read /etc/nsswitch.conf: %v", err)
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
```
dorileo, line 109: This block is repeated, can we have a generic function to read these config files? The difference between this and the other I saw is the reader.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -0,0 +1,322 @@
...
+			if strings.HasPrefix(line, "[sudo] password for ") {
+				_, err = in.Write([]byte(sudo + "\n"))
+				if err != nil {
+					break
+				}
+			}
+		}
+	}(in, out, &output)
+
+	_, err = session.Output("sudo ls /var/google-sudoers.d")
```
dorileo, line 174: Are you checking only the existence of the file? Shouldn't we check the content as well?

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -0,0 +1,322 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get session stdin: %v", err)
+	}
+	out, err := session.StdoutPipe()
+	if err != nil {
+		return "", fmt.Errorf("failed to get session stdout: %v", err)
+	}
+
+	// Monitor the output stream of the ssh session.
+	go func(in io.WriteCloser, out io.Reader, output *[]byte) {
```
dorileo, line 149: I'm having a hard time understanding why this is implemented this way... are we running a remote command? Can't we simply copy the file via ssh?

dorileo (changes requested):

imagetest/test_suites/oslogin/setup.go
```diff
@@ -2,25 +2,47 @@ package oslogin
...
-	vm2, err := t.CreateTestVM("vm2")
+	twofa, err := t.CreateTestVM("twofa")
 	if err != nil {
 		return err
 	}
-	vm2.AddMetadata("enable-oslogin", "false")
-	vm2.RunTests("TestOsLoginDisabled")
+	twofa.AddScope(computeScope)
+	twofa.AddMetadata("enable-oslogin", "true")
+	twofa.AddMetadata("enable-oslogin-2fa", "true")
```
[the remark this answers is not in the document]
drewhli: Yes, that's what `defaultVM` is doing.

imagetest/test_suites/oslogin/ssh_test.go
drewhli: So when we create the SSH client, it turns out we do not have the permissions to access the file. The only way I found to access the `google-sudoers.d` file is to use `sudo`, which requires the user's sudo password to be inputted. As for copying the file, I would imagine we'd run into the same permissions issue, which means we wouldn't be able to copy the file without sudo.

From local testing on test VMs, this was the case -- we needed root permissions to access and read the file.

imagetest/test_suites/oslogin/oslogin_utils.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	flag, err := strconv.ParseBool(data)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse enable-oslogin-2fa metadata entry: %+v", err)
+	}
+	return flag, nil
+}
+
+// Gets the given secret.
+func getSecret(ctx context.Context, client *secretmanager.Client, secretName string) (string, error) {
```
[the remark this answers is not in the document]
drewhli: I don't believe any other tests are using secrets right now, but I understand what you mean. I will move it to the more general utils file since it's a more general function rather than something that is specifically built for OSLogin testing.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -0,0 +1,336 @@
...
+	session, err := client.NewSession()
+	if err != nil {
+		return fmt.Errorf("failed to create ssh session: %v", err)
+	}
+	data, err := session.Output("cat /etc/nsswitch.conf")
+	if err != nil {
+		return fmt.Errorf("failed to read /etc/nsswitch.conf: %v", err)
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
```
dorileo, line 123: why are we not using the introduced generic function here?

dorileo (changes requested):

imagetest/test_suites/oslogin/ssh_test.go
drewhli: No reason. Implemented now.

dorileo (approved):

PR #830, opened 2023-10-04
Fork Concourse windows image build pipeline for MBR

jjerger:
Adds MBR builds as a separate pipeline to be run ad-hoc

### DIFF
--- a/concourse/pipelines/windows-image-build-mbr.jsonnet
+++ b/concourse/pipelines/windows-image-build-mbr.jsonnet
@@ -276,25 +276,13 @@ local ImgGroup(name, images, environments) = {
 
 // Start of output.
 {
-  local windows_2012_images = [
+  local images = [
     'windows-server-2012-r2-dc-mbr',
     'windows-server-2012-r2-dc-core-mbr',
-  ],
-  local windows_2016_images = [
     'windows-server-2016-dc-mbr',
-    'windows-server-2016-dc-core-mbr',
-  ],
-  local windows_2019_images = [
     'windows-server-2019-dc-mbr',
-    'windows-server-2019-dc-core-mbr',
-  ],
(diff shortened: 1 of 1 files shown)


jjerger:
Question for Brian and Zach: Does it make sense to define the pipeline assuming that only the "testing" env will ever be passed in? I had removed alot of the branching logic from the original pipeline that stemmed from there being 4 publishing environments.

concourse/pipelines/windows-image-build-mbr.jsonnet
```diff
@@ -0,0 +1,346 @@
...
+          ImgPublishJob(image, env, 'windows', 'windows-bios')
+          for image in windows_server_images
+          for env in envs
+        ],
+
+  groups: [
+    ImgGroup('windows-2012-mbr', windows_2012_images, envs),
+    ImgGroup('windows-2016-mbr', windows_2016_images, envs),
+    ImgGroup('windows-2019-mbr', windows_2019_images, envs),
+    ImgGroup('windows-2022-mbr', windows_2022_images, envs),
```
zmarano, line 344: I don't believe we need to build 2022 as an MBR image- it never existed as MBR. WDYT?

bkatyl: We need it because customers can import Windows Server 2022 MBR.

bkatyl:
As we talked about in chat, lets remove the core images since these will only be for boot tests and we don't need or want to encourage double testing.

concourse/pipelines/windows-image-build-mbr.jsonnet
zmarano: Good point, or in place upgrades. Ok.

bkatyl (approved):

PR #831, opened 2023-10-04
Add wait for vm quota test workflow step, and functions to append quotas

a-crate:
During testing I ran into a lot of problems with finding free quota that was taken by the time vm creation started. Some of this is unavoidable because daisy takes a second to switch between steps, but I mitigated the problem by staggering parallel workers. Most of this can be avoided by just waiting 60-90 seconds after starting a workflow.

The other thing I did was break this up from one large wait-for-quota step that all creation steps depend on, to quota steps for each create step. But the only create step that actually needs enough quota to worry about right now is the create VMs step, so I just took the others out avoid to avoiding leaving in unused code.

If it's not clear enough that the existing API is only for quota that gates VM creation I can add some comments expanding on this in the code.

Previous PR before I reset my branch and dropped changes on master: https://github.com/GoogleCloudPlatform/guest-test-infra/pull/823 This one is not identical, I went back and revised the API a bit.

### DIFF
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -82,11 +82,6 @@ func (t *TestWorkflow) LockProject() {
 
 // WaitForVMQuota appends a list of quotas to the wait for vm quota step. Quotas with a blank region will be populated with the region corresponding to the workflow zone.
 func (t *TestWorkflow) WaitForVMQuota(qa ...*daisy.QuotaAvailable) error {
-	for _, q := range qa {
-		if q == nil {
-			return fmt.Errorf("not append a nil pointer to quota")
-		}
-	}
 	step, ok := t.wf.Steps[waitForVMQuotaStepName]
 	if !ok {
 		step, err := t.wf.NewStep(waitForVMQuotaStepName)
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -27,6 +27,7 @@ import (
 )
 
 const (
+	waitForVMQuotaStepName   = "wait-for-vm-quota"
 	createVMsStepName        = "create-vms"
 	createDisksStepName      = "create-disks"
 	createNetworkStepName    = "create-networks"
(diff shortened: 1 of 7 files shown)


imagetest/testworkflow.go
```diff
@@ -565,6 +584,7 @@ func RunTests(ctx context.Context, storageClient *storage.Client, testWorkflows
 		wg.Add(1)
 		go func(id int) {
 			defer wg.Done()
+			time.Sleep(time.Duration(id) * stagger)
```
a-crate, line 587: @dorileo 
> why not time.Sleep(stagger * time.Duration(i)) ?

I kept `id` because of this: https://go.dev/blog/loopvar-preview whole thing. `go vet` will fail it in the presubmit too.
Got rid of the unnecessary casts though.

dorileo: Ah yeah, for sure, you need to manage the go routine context access - that's fine, I meant to ask about the casts.

imagetest/fixtures.go
```diff
@@ -79,6 +80,25 @@ func (t *TestWorkflow) LockProject() {
 	t.lockProject = true
 }
 
+// WaitForVMQuota appends a list of quotas to the wait for vm quota step. Quotas with a blank region will be populated with the region corresponding to the workflow zone.
+func (t *TestWorkflow) WaitForVMQuota(qa ...*daisy.QuotaAvailable) error {
```
dorileo, line 84: Right now we are passing in a single qa, do you see us needing more than one?

imagetest/fixtures.go
```diff
@@ -79,6 +80,25 @@ func (t *TestWorkflow) LockProject() {
...
+	for _, q := range qa {
```
dorileo, line 85: You are only handling this condition because the caller is not taking care of it, if you avoid calling this method if quota is nil then you don't need to handle it here.

imagetest/test_suites/shapevalidation/setup.go
```diff
@@ -109,6 +119,9 @@ Familyloop:
 				continue Familyloop
 			}
 		}
+		if err := t.WaitForVMQuota(shape.quota); shape.quota != nil && err != nil {
```
dorileo, line 122: All the other cases you have the guarantee that you have quota, this is the only case you'd have to deal with exceptions:

This would look cleaner (and would remove the need of validating qa inside WaitForVMQuota()):

```
if shape.quota != nil {
  if err := t.WaitForVMQuota(shape.quota); err != nil {
    return err
  }
}
```

imagetest/fixtures.go
a-crate: It's a variadic argument, this function can handle multiple quotas.

imagetest/test_suites/shapevalidation/setup.go
a-crate: Done

imagetest/fixtures.go
dorileo: I know, my question is do we have such a use case?

a-crate: I misread your comment, I'm sorry. I don't think we have any use right now, everything that needs quotas needs multiple quotas but they're all stored in such a way that we call for them one by one.

dorileo: Yep, just that the method signature should reflects its use case (ideally).

a-crate: Makes sense. Done.

PR #840, opened 2023-10-09
Check for gve in gvnic tests as well

drewhli:


imagetest/test_suites/networkperf/gvnic_test.go
```diff
@@ -26,8 +26,8 @@ func CheckGVNICPresent(interfaceName string) error {
 	}
 	s := strings.Split(data, "/")
 	driver := s[len(s)-1]
-	if driver != "gvnic" {
-		errMsg := fmt.Sprintf("Driver set as %v want gvnic", driver)
+	if driver != "gvnic" && driver != "gve" {
```
zmarano, line 29: I think this is supposed to be || right?

drewhli: Yes.

drewhli: No, wait, it is supposed to be && because if it's not either of them, then we fail.

imagetest/test_suites/network/gvnic_test.go
```diff
@@ -26,8 +26,8 @@ func CheckGVNICPresent(interfaceName string) error {
 	}
 	s := strings.Split(data, "/")
 	driver := s[len(s)-1]
-	if driver != "gvnic" {
-		errMsg := fmt.Sprintf("Driver set as %v want gvnic", driver)
+	if driver != "gvnic" && driver != "gve" {
```
zmarano, line 29: ditto

PR #841, opened 2023-10-09
Test pdbalanced performance on ARM64 images in storageperf

a-crate:
Perf targets from https://cloud.google.com/compute/docs/disks/performance#t2a_vms

### DIFF
--- a/imagetest/test_suites/storageperf/setup.go
+++ b/imagetest/test_suites/storageperf/setup.go
@@ -19,27 +19,26 @@ var Name = "storageperf"
 var scripts embed.FS
 
 type storagePerfTest struct {
-	machineType string
-	arch string
-	diskType string
-	cpuMetric string
-	minCpuPlatform string
-	zone string
+	machineType    string
+	arch           string
+	diskType       string
+	cpuMetric      string
+	minCPUPlatform string
+	zone           string
 }
-
 
 const (
 	linuxInstallFioScriptURL   = "startupscripts/install_fio.sh"
 	windowsInstallFioScriptURL = "startupscripts/install_fio.ps1"
 )
 
-var	storagePerfTestConfig = []storagePerfTest{
+var storagePerfTestConfig = []storagePerfTest{
 	{
-		arch: "X86_64",
+		arch:        "X86_64",
 		machineType: "h3-standard-88",
-		zone: "us-central1-a",
-		diskType: imagetest.PdBalanced,
-		cpuMetric: "CPUS",
+		zone:        "us-central1-a",
+		diskType:    imagetest.PdBalanced,
+		cpuMetric:   "CPUS",
 	},
 	/* temporarily disable c3d hyperdisk until the api allows it again
 	{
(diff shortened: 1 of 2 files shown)


imagetest/test_suites/storageperf/setup.go
```diff
@@ -29,17 +29,24 @@ func TestSetup(t *imagetest.TestWorkflow) error {
...
-		// {"machineType": "c3d-standard-180", "zone": "us-east4-c", "diskType": imagetest.HyperdiskExtreme, "metric": "CPUS"}, // No public metric for this yet but the CPU count will work because they're so large
-		{"machineType": "n2-standard-80", "diskType": imagetest.HyperdiskExtreme, "metric": "N2_CPUS"},
-		{"machineType": "c3-standard-88", "diskType": imagetest.PdBalanced, "metric": "C3_CPUS"},
-		{"machineType": "c3d-standard-180", "zone": "us-east4-c", "diskType": imagetest.PdBalanced, "metric": "CPUS"},
-		{"machineType": "n2d-standard-64", "diskType": imagetest.PdBalanced, "metric": "N2D_CPUS"},
-		{"machineType": "n1-standard-64", "diskType": imagetest.PdBalanced, "minCpuPlatform": "Intel Skylake", "metric": "CPUS"}, // N1 CPUS aren't measured separately
-		// zone for h3 is a temporary measure while h3 is pre-GA
-		{"machineType": "h3-standard-88", "zone": "us-central1-a", "diskType": imagetest.PdBalanced, "metric": "CPUS"}, // No public metric
+	var paramMaps []map[string]string
+	if strings.Contains(t.Image, "arm64") {
```
dorileo, line 33: What if we do it a bit more robust and make `paramMap` a bit smarter and give it the knowledge of what's the target os type? We can either make `paramMaps` like `[]TestConfig` where `TestConfig` (maybe `TestConfig` is not the best naming though) would be:

```golang
type TestConfig struct {
   machineType string
   diskType string
   metric string
   targetOS string
   ...
}
```

And then when using `paramMaps` we'd filter it out based on the required criteria, being it target OS or anything else depending on the requirements we have in hand.

a-crate: I'd like to wait until after #800 is reviewed to do something like that, it would be much more robust (and easier) to use the compute API directly to do this. I can hold this PR until then.

a-crate:
Going to do this now rather than later after all due to some new context.

imagetest/test_suites/storageperf/setup.go
a-crate: Take a look, switched to a struct with defined fields for existing options and handled each of them in a fairly similar way to what's being done currently.

imagetest/test_suites/storageperf/setup.go
```diff
@@ -116,24 +169,18 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 	return nil
 }
 
-// Due to compatability issues or other one off errors, some combinations of machine types and images must be skipped. This function returns true for those combinations.
-func skipMachineTypeImage(machineType, image string) bool {
-	if strings.HasPrefix(machineType, "c3d") && (strings.Contains(image, "windows-2012") || strings.Contains(image, "windows-2016")) {
+func skipTest(tc storagePerfTest, image string) bool {
+	if tc.arch == "ARM64" && !strings.Contains(image, "arm64") {
```
dorileo, line 173: The goal is to disable arm64 tests at all?

a-crate: No, just to avoid running non-arm test configurations on arm images and vice versa.

PR #848, opened 2023-10-10
Shutdown Script Minor Change

koln67:
Minor change to the wait step for tests running the shutdown scripts.

This is in preparation for adding guest attributes: after a follow up change, the first boot of the shutdown script will wait for a guest attribute with a special name. This should make the shutdown script test robust even after using guest attributes instead of the serial log.

### DIFF
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -25,7 +25,7 @@ func main() {
 	log.Printf("FINISHED-BOOTING")
 	defer func() {
 		if willRebootDuringTest, err := utils.GetMetadataAttribute("willRebootDuringTest"); err != nil {
-			log.Printf("found reboot metadata key")
+			log.Printf("found reboot metadata key %s", willRebootDuringTest)
 		}
 		for f := 0; f < 5; f++ {
 			log.Printf("FINISHED-TEST")
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -24,9 +24,6 @@ func main() {
 	}
 	log.Printf("FINISHED-BOOTING")
 	defer func() {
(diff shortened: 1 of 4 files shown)


imagetest/cmd/wrapper/main.go
```diff
@@ -24,6 +24,9 @@ func main() {
 	}
 	log.Printf("FINISHED-BOOTING")
 	defer func() {
+		if willRebootDuringTest, err := utils.GetMetadataAttribute("willRebootDuringTest"); err != nil {
```
dorileo, line 27: You are not actually using willRebootDuringTest other than printing it, I would leave it out of this PR and only introduce it in the followup PR where we actually use it.

koln67: done

imagetest/fixtures.go
```diff
@@ -141,6 +141,56 @@ func (t *TestWorkflow) CreateTestVM(name string) (*TestVM, error) {
 	return &TestVM{name: vmname, testWorkflow: t, instance: i}, nil
 }
 
+// CreateTestVMWithReboot adds the necessary steps to create a VM with the
+// specified name to the workflow, while expecting that it will be rebooted.
+// This should only be used in cases where there is a special interaction
+// with the wrapper binary, such as in the shutdown script tests.
+func (t *TestWorkflow) CreateTestVMWithReboot(name string) (*TestVM, error) {
```
dorileo, line 148: I wonder if we can't reuse most of the current `CreateTestVM()` function and only add an argument/flag to make the differentiation.

dorileo:
Overall the PR is looking good; The strategy looks good too.

imagetest/fixtures.go
koln67: done

PR #860, opened 2023-10-12
Remove 2012 R2 from prod windows build pipeline

jjerger:
EOS was October 10; these will no longer be used


concourse/templates/daisy.libsonnet
```diff
@@ -67,11 +67,9 @@
     iso_path_2022:: error 'must set iso_path_2022 in daisy image task',
     iso_path_2019:: error 'must set iso_path_2019 in daisy image task',
     iso_path_2016:: error 'must set iso_path_2016 in daisy image task',
-    iso_path_2012r2:: error 'must set iso_path_2012r2 in daisy image task',
```
jjerger, line 70: I'm... Not confident this is a good idea in retrospect.

Perhaps leaving these in and still importing the secret (even if unused) is the better option here? Then we can add it back in if needed in the future.

Plus this is also going to require reworking the other pipelines that have 2012 for test purposes.

Actually, reading that back makes me realize this is still needed.

PR #874, opened 2023-10-17
Test vm params

koln67:
Convert CreateTestVMMultipleDisks to use only a test vm params struct

### DIFF
--- a/imagetest/fixtures.go
+++ b/imagetest/fixtures.go
@@ -58,6 +58,19 @@ type TestVM struct {
 	instance     *daisy.Instance
 }
 
+// TestVMParams stores config params for a test vm
+type TestVMParams struct {
+	Disks []*compute.Disk
+	// this bool pointer must be explicitly initialized to use
+	// the special guest attribute for tests which reboot
+	VmRebootsDuringTest *bool
+	ExtraScopes         []string
+	MachineType         string
+	MinCpuPlatform      string
+	Hostname            string
+	Zone                string
+}
+
 // AddUser add user public key to metadata ssh-keys.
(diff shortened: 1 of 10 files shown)


imagetest/fixtures.go
```diff
@@ -58,6 +58,19 @@ type TestVM struct {
 	instance     *daisy.Instance
 }
 
+// TestVMParams stores config params for a test vm
+type TestVMParams struct {
```
a-crate, line 62: This has one of the same limitations that that param map did: every option needs to be explicitly added and copied over to the daisy.Instance in appendCreateVMStep. Why not take a daisy.Instance directly and allow the caller to specify any options daisy is capable of understanding?

imagetest/fixtures.go
```diff
@@ -58,6 +58,19 @@ type TestVM struct {
...
+	Disks []*compute.Disk
+	// this bool pointer must be explicitly initialized to use
+	// the special guest attribute for tests which reboot
+	VMRebootsDuringTest *bool
```
a-crate, line 66: Maybe this should extend the existing `func (t *testVM) Reboot` instead of being a separate parameter, or be a separate `RebootViaManager` function that explains the difference. That would make everything else cleanly map to an option that can be set directly on a compute.Instance or a daisy.Instance

imagetest/fixtures.go
koln67: I see, working on a new PR which uses *daisy.Instance directly

imagetest/fixtures.go
koln67: The planned solution in a new pr is to set the *daisy.Instance metadata map directly

PR #880, opened 2023-10-20
Move some tests from imagevalidation to packagevalidation

koln67:


imagetest/test_suites/hostnamevalidation/image_validation_test.go
```diff
@@ -1,7 +1,7 @@
 //go:build cit
```
zmarano, line 1: change the filename as well?

koln67: done

PR #885, opened 2023-10-24
Deletewinvalidation

koln67:


imagetest/test_suites/disk/disk_readwrite_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+func TestDiskReadWrite(t *testing.T) {
+	utils.WindowsOnly(t)
```
a-crate, line 15: Why is this windows only? Cant' we just change the paths for linux and test on both?

imagetest/test_suites/disk/disk_resize_test.go
```diff
@@ -23,6 +23,7 @@ const (
 
 // TestDiskResize Validate the filesystem is resized on reboot after a disk resize.
 func TestDiskResize(t *testing.T) {
+	utils.LinuxOnly(t)
```
a-crate, line 26: I think this should be supported on windows too  https://cloud.google.com/compute/docs/disks/resize-persistent-disk

imagetest/test_suites/disk/disk_readwrite_test.go
a-crate: Expanding the test is much more working than just moving it over in this PR, but I think there should be a TODO comment for this at least.

imagetest/test_suites/disk/disk_resize_test.go
a-crate: Ditto on todo

koln67: it is probably supported, however the test currently assumes linux with the markerfile and a unix call in verifyDiskSize().

I think windows disk resizing would make the most sense in a follow up PR: I've added "TODO" in the comments

imagetest/test_suites/disk/disk_readwrite_test.go
koln67: I added the linux test for disk, mostly a copy of the part of the hotattach test which reads the file.

imagetest/test_suites/disk/disk_resize_test.go
a-crate: Not sure what's up with the test setup, but I get this failure running the test on windows:
```
<testcase classname="disk-windows-2016" name="TestDiskResize" time="0">
			<failure type="Failure">FileIOError: stat /disk64.exe: no such file or directory</failure>
</testcase>
```
I don't think it's skipping correctly

koln67: it was because of the build directive on the first two lines, and because the unix import fails the build on windows. Both should be fixed with the latest commit

PR #889, opened 2023-10-27
Remove SAP workload test (and required scripts from concourse)

a-crate:


a-crate:
/assign @zmarano 
Zach, was there any nuance I missed to taking the SAP workload tests out or can we just remove them like this?

zmarano:
> /assign @zmarano Zach, was there any nuance I missed to taking the SAP workload tests out or can we just remove them like this?

I think that's right!

PR #890, opened 2023-10-30
util: unify metadata code

dorileo:
Remove duplicate methods and stop assuming instance attributes for all calls, let the callers decide the metadata root level.

### DIFF
--- a/imagetest/utils/test_utils.go
+++ b/imagetest/utils/test_utils.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net"
-	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
@@ -18,6 +17,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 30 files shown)


imagetest/utils/test_utils.go
```diff
@@ -201,7 +125,9 @@ func DownloadPrivateKey(user string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	sourcesPath, err := GetMetadataAttribute("daisy-sources-path")
+
+	// TODO: pass down the main context here
+	sourcesPath, err := GetMetadata(context.Background(), "instance", "attributes", "daisy-sources-path")
```
drewhli, line 130: I think it's better to have the `GetMetadata` method itself call `context.Background()` instead of having to pass it in. It'll make the method more intuitive to call.

dorileo: The idea is to never have anyone calling `GetMetadata()` with a `context.Background()` instead using the main process's context, hence the comment right above (on line 129). Having `GetMetadata()` creating its own context will only lead to disasters due to no context cancelation strategy.

dorileo: I introduced a deadline based context cancelation and removed the TODOS and this context creation inside these util functions. Thanks.

PR #892, opened 2023-10-31
oslogin: only test sshd's PAM config if 2fa is enabled

dorileo:
The OS Login stack will only have the sshd's PAM config containing the pam_oslogin_login.so module when two factors authentication is enabled.


drewhli (approved):
/lgtm

PR #900, opened 2023-11-03
Update README.md with manager instructions

koln67:


imagetest/README.md
```diff
@@ -150,6 +150,11 @@ To build from a directory other than `imagetest`
 
     $ local_build.sh -o $outspath -i $path_to_imagetest
 
+To run the tests, cd into $outspath, set the shell variables and run
+
+    $ manager -zone $ZONE -project $PROJECT -images $images -filter $test_suite_name -local_path $some_path
```
a-crate, line 155: If they already did `cd $outspath` couldn't they just do `-local_path .`?

koln67: Yeah, local_path can be any path. Setting it to . probably makes it easier to understand

PR #910, opened 2023-11-09
Retry network performance results fetch three times

a-crate:
Retry GET for network results and POST for network results three times each with an increasing delay after each failed request.


a-crate:
This won't eliminate 404s entirely but it should cut down on them in the short term.

PR #912, opened 2023-11-09
Wait for fio to finish installing on linux

koln67:
Sometimes, there is a race condition where the linux startup script to install fio does not complete before the test runs.

This causes the fio command to fail because it was not found. To fix this, wait until fio is installed.

### DIFF
--- a/imagetest/test_suites/storageperf/storage_perf_utils.go
+++ b/imagetest/test_suites/storageperf/storage_perf_utils.go
@@ -3,7 +3,6 @@ package storageperf
 import (
 	"fmt"
 	"os/exec"
-	"time"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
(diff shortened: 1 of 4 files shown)


imagetest/test_suites/storageperf/storage_perf_utils.go
```diff
@@ -119,6 +121,16 @@ type FIOStatistics struct {
 	X              map[string]interface{} `json:"-"`
 }
 
+// In some cases, fio might take longer to install on linux than expected.
+// To ensure tests do not fail from not finding fio, wait for the install.
+func waitFIOInstalledLinux() {
```
drewhli, line 126: It's probably a good idea to have a maximum time for this to run for before we give up and fail the test. With this, if `fio` somehow fails to get installed, we wait for the entire 45min timeout. We could definitely shorten that down to 5 minutes or less.

drewhli (changes requested):

imagetest/test_suites/storageperf/storage_perf_utils.go
a-crate: It might also be more reliable (and faster, since it's not polling every 10 seconds) to switch to calling the `installFioLinux()` function so that we know exactly when it's done. That would bypass the conditions that allowed for the possibility of a race condition in the first place.

koln67: done

PR #916, opened 2023-11-10
Revamp guest-test-infra presubmit checks

a-crate:
We are missing a lot of gobuild/gocheck/gotest coverage because these containers are scoped to only subdirectories which belong to the same module as the main directory. This change will modify the citpresubmit container to either build CIT or test it based on the argument given to the container. This should be accompanied by the change in guest-test-infra presubmit configuration to check/test/build each module in guest-test-infra separately.

This also includes a fix for registry-image-forked's `gocheck` presubmit, which would otherwise fail because it's accessing linux only APIs in a file that's not scoped for linux builds only.

### DIFF
--- a/container_images/citpresubmit/main.sh
+++ b/container_images/citpresubmit/main.sh
@@ -15,26 +15,24 @@
 
 set -xe
 
-BUILD_DIR=$1
-[[ -n $BUILD_DIR ]] && cd $BUILD_DIR
-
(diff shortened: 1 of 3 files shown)


a-crate:
/hold
Should be merged at the same time as https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2116

PR #924, opened 2023-11-14
Print vm name for failing tests

koln67:
In the error log for test failures, print the machine type. For example, we would know if the failure was pd balanced for n1-standard-64 based on the error message vm name.


drewhli (approved):
/lgtm

PR #929, opened 2023-11-17
Add 2FA OSLogin Tests.

drewhli:


imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -42,11 +44,14 @@ const (
...
-	normal2FASshKey  = "normal-2fa-ssh-key"
-	admin2FASshKey   = "admin-2fa-ssh-key"
+	// SSH keys
+	normalUserSSHKey = "normal-user-ssh-key"
+	adminUserSSHKey  = "admin-user-ssh-key"
+	normal2FASSHKey  = "normal-2fa-ssh-key"
+	admin2FASSHKey   = "admin-2fa-ssh-key"
+
+	// Time to wait for agent check
+	waitAgent = time.Second * 15
```
dorileo, line 54: If we are setting a const for this can we set the other sleep times as well? at least the ones your are changing. Please add context to why the values seem to be the right one. I general a fall back retry strategy seem always to be a better option rather than these arbitrary sleep but I'm not blocking this PR because of that, let's see if we can address it in a specific project later.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -156,7 +161,7 @@ func TestAgent(t *testing.T) {
 		t.Fatalf("Error changing metadata: %v", err)
 	}
 	// Give the API time to update.
-	time.Sleep(time.Second * 5)
+	time.Sleep(time.Second * 4)
```
dorileo, line 164: I see you are shifting some of these sleep times, why?

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -222,8 +227,8 @@ func TestSSH(t *testing.T) {
 	}
 }
 
+// Checks if an admin user can use sudo after successfully SSH-ing to an instance.
 func TestAdminSSH(t *testing.T) {
-	time.Sleep(20 * time.Second)
```
dorileo, line 226: why is this sleep not relevant any more?

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -271,3 +276,192 @@ func TestAdminSSH(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to get hostname: %v", err)
+	}
+	hostname := fmt.Sprintf("%s:22", vmname)
+	signer, err := ssh.ParsePrivateKey([]byte(privateSSHKey))
+	if err != nil {
+		t.Fatalf("failed to parse private key: %v", err)
+	}
+
+	cb := func(name, instruction string, questions []string, echos []bool) (answers []string, err error) {
```
dorileo, line 312: I suppose we are handling the 2fa challenges here, right? can you add a comment block describing it? if you could add the context to what challenge the test is using and why would be great.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -271,3 +276,192 @@ func TestAdminSSH(t *testing.T) {
...
+			return
+		}
+		if firstQuestionRegex.MatchString(questions[0]) {
+			// 1 is the response for Authenticator OTP
+			answers[0] = "1"
+			return
+		}
+
+		// If not the first question, input code for two-factor.
+		ctx := context.Background()
```
dorileo, line 329: Why are we creating a new context here? if you need a new context please derive the main one, otherwise just use the main one.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -271,3 +276,192 @@ func TestAdminSSH(t *testing.T) {
...
+		t.Fatalf("failed to get hostname: %v", err)
+	}
+	hostname := fmt.Sprintf("%s:22", vmname)
+	signer, err := ssh.ParsePrivateKey([]byte(privateSSHKey))
+	if err != nil {
+		t.Fatalf("failed to parse private key: %v", err)
+	}
+
+	// Callback function for SSH keyboard-interactive challenges.
+	cb := func(name, instruction string, questions []string, echos []bool) (answers []string, err error) {
```
dorileo, line 402: Can we generalize this callback? as it's repeating itself here and in the other test function.

imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -271,3 +276,192 @@ func TestAdminSSH(t *testing.T) {
...
+			return
+		}
+		if firstQuestionRegex.MatchString(questions[0]) {
+			// 1 is the response for Authenticator OTP
+			answers[0] = "1"
+			return
+		}
+
+		// If not the first question, input code for two-factor.
+		ctx := context.Background()
```
dorileo, line 419: ditto.

concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -153,7 +153,7 @@ local base_buildpackagejob = {
       },
     },
```
dorileo, line 154: I'm skipping this changes as I know you are still working on those.

dorileo (changes requested):

imagetest/test_suites/oslogin/ssh_test.go
drewhli: The tests are run in sequence. So if the TestAgent test is already done by the time the first test `TestSSH` runs, then there's no more need to wait for TestAgent to finish if it's already done.

imagetest/test_suites/oslogin/ssh_test.go
drewhli: I've tried reducing the time to see if we can get away without it. There was no difference between 4s and 5s, so I went with the lower one to save a few seconds per test run.

imagetest/test_suites/oslogin/ssh_test.go
drewhli: Acknowledged.

I'll add context here for now, but I'll add it in comments as well in a followup commit.
The time for waitAgent is, obviously, for the TestAgent check. The TestAgent itself has a few `sleep` statements as it needs to wait for metadata changes to propagate and for the guest agent to respond to the metadata changes accordingly. So this time is determined by the total amount of sleep time (8 seconds at the time), the runtime of the agent check test (around 1-2.5s), with a little bit of extra padding to make extra sure things are ready.

imagetest/test_suites/oslogin/ssh_test.go
drewhli: I will document this in a followup commit.

concourse/pipelines/guest-package-build.jsonnet
drewhli: It turns out the diff view is messed up. Nothing to change. Only changes from me are lines 684 and 707 in this file.

dorileo: ok

PR #930, opened 2023-11-17
Skip tests on zypper install issue

koln67:
Most commonly, sometimes on startup the sles instances might have a zypper process running, leading to an exit code 7 when running "zypper install." Other errors such as 104 (repository not available) or 106 (zypper refreshing repositories failed) sometimes happen as well, indicating a temporary backend error.

This is not actionable by PD or teams working on disk performance and should not fail the test, so skipping the specific test case run is the proposed solution. I'm open to other suggestions for zypper backend errors. 

Also, use the full size of the mounted disk for benchmarking
https://cloud.google.com/compute/docs/disks/benchmarking-pd-performance#raw-disk

### DIFF
--- a/imagetest/test_suites/storageperf/iops_read_test.go
+++ b/imagetest/test_suites/storageperf/iops_read_test.go
@@ -72,7 +72,7 @@ func RunFIOReadLinux(t *testing.T, mode string) ([]byte, error) {
 	}
 
 	if !utils.CheckLinuxCmdExists(fioCmdNameLinux) {
-		if err = installFioLinux(); err != nil {
+		if err = installFioLinux(t); err != nil {
 			return []byte{}, fmt.Errorf("linux fio installation failed: err %v", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


a-crate:
This is basically a failure in setting up the test environment, which we handle everywhere else in CIT by failing the test case or retrying. This would be a departure from how we handle other setup failures.
I know this is a performance test so it's not being used to block releases to prod or staging but we also talked about skipping for setup failures during shapevalidation and decided not to. Test skips are treated as pass, not failure, so this kind of design could lead to an image unintentionally passing CIT in the right conditions. I think we should be wary of introducing things like that, even if they're not currently being used to block any releases.

koln67:
It looks like for most of these errors, the retry will fail with the same error because the VM has an incorrect configuration on test startup.

I changed it so there is a custom error message returned in these cases. When reading the test failure message, we will know that it is a known error in the test environment setup.

PR #932, opened 2023-11-22
Increase machine size for network test

koln67:
For EL7 network tests, some race conditions relating to networkmanager on el7 can lead to the machine exhausting cpus or memory on a n1-standard-1. 

Use n1-standard-32 instead for the instance which needs to reboot.

### DIFF
--- a/imagetest/test_suites/network/setup.go
+++ b/imagetest/test_suites/network/setup.go
@@ -63,7 +63,9 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 	// VM2 for multiNIC
 	networkRebootInst := &daisy.Instance{}
 	networkRebootInst.Metadata = map[string]string{imagetest.ShouldRebootDuringTest: "true"}
-	vm2, err := t.CreateTestVMMultipleDisks([]*compute.Disk{{Name: vm2Config.name}}, networkRebootInst)
+	// this vm could run out of memory and lead to race conditions on n1-standard-1, the default shape. Instead, allocate more cpus and memory.
+	networkRebootInst.MachineType = "n1-standard-8"
+	vm2, err := t.CreateTestVMMultipleDisks([]*compute.Disk{{Name: vm2Config.name, SizeGb: 30}}, networkRebootInst)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


a-crate:
What's the root cause of this? It shouldn't take 32 vCPUs and 120GB of memory to run this test suite, I think we should fix whatever is causing it to need this large amount of resources.

koln67:
It appears to be an issue with NetworkManager on EL7. If it runs out of space or vcpus (not clear which one it runs out of on n1-standard-1), the NetworkManager service on the vm can't connect to the state "CONNECTED_GLOBAL", and os-config agent can't connect to the metadata server. 

n1-standard-8 still sometimes fails this test, so the smallest machine we can try is n1-standard-16 (I haven't verified whether that is enough space to pass consistently).

The root cause is in NetworkManager El7 (version 1.18.8-2.el7_9) using a ton of space, which is unlikely to be patched for our use case.

ajorg:
Would that mean that a customer using EL7 on n1-standard-1 is likely to find that networking doesn't work? Or that the guest agent can't contact the metadata server?

koln67:
> Would that mean that a customer using EL7 on n1-standard-1 is likely to find that networking doesn't work? Or that the guest agent can't contact the metadata server?

When the vm restarts, it sometimes is not connected to the internet, leading to the os-config agent failing to reach the metadata server, regardless of retries.

The test still indicates that networking on EL7 n1-standard-1 works on startup. At worst, n1-standard-1 networking on EL7 might sometimes not work on reboot, though this might also be a test configuration issue and not a general issue.

a-crate:
> At worst, n1-standard-1 networking on EL7 might sometimes not work on reboot

The suite reboots intentionally to catch this, if that's the case then this is a real test failure and we should fix whatever is causing this.

ajorg:
> and we should fix whatever is causing this

If it can't be fixed, say because it's in the OS and EL7 is too old to be fixed, should it become a known failure that stays failing so it stays known? (Assuming we don't find a workaround.)

a-crate:
> > and we should fix whatever is causing this
> 
> If it can't be fixed, say because it's in the OS and EL7 is too old to be fixed, should it become a known failure that stays failing so it stays known? (Assuming we don't find a workaround.)

What we have done for other known failures that will never be fixed is include some logic that doesn't run the test on known failing images, and known failures that will be fixed at some indeterminate time in the future are `Skip()`ed with some message. Or at least that's what we've discussed in meetings, I don't think this practice is universally applied in the codebase.

It's a constraint of our CI that we can't release images with failing test suites so we don't leave known failures in.

PR #935, opened 2023-11-27
Reduce Parallel Thread Count for Network Performance Tests for SUSE Images.

drewhli:
When the number of parallel threads is high, the images seem to reject them for some reason. This is a patchwork fix so the tests don't constantly time out.


imagetest/test_suites/networkperf/startupscripts/netclient_startup.sh
```diff
@@ -46,6 +47,7 @@ elif [[ -f /usr/bin/zypper ]]; then
   echo "$(date +"%Y-%m-%d %T"): zypper found Installing iperf."
   sudo zypper --no-gpg-checks refresh
   sudo zypper --no-gpg-checks --non-interactive install https://iperf.fr/download/opensuse/iperf-2.0.5-14.1.2.x86_64.rpm netcat
+  parallelthreads=4
```
koln67, line 50: If this is a temporary patch fix, it would be useful for future reference to add a comment or TODO, which mentions what the optimal future fix would be.

drewhli: Done.

PR #940, opened 2023-11-29
Don't delete resources named do-not-delete

a-crate:
/cc @zmarano @jjerger @koln67 @drewhli


a-crate:
Also includes a partial revert of #933, the library still exists but the cleanerupper doesn't use it after this PR. It deletes things too early but the image runs once per hour and I need a little more time.

a-crate:
I was wrong, it is deleting things on time. Just got nervous after missing the name attribute. Cancel the revert.

PR #943, opened 2023-11-30
Switch order of read and write tests

koln67:
On the test grid, sometimes the c3-standard-88 hyperdisk extreme test has unusual behavior. Sometimes (but not all the time), the IOPS are high enough for random read to pass, but the IOPS are a little bit lower for random write, just enough to fail the test.

This change is testing the hypothesis that the Random Read test is changing the state in a way which slows down the future Random Write test, by running random write first instead of random read. If this is confirmed, a follow up change would be made to split the Read and Write tests into two VMs.

### DIFF
--- a/imagetest/test_suites/storageperf/setup.go
+++ b/imagetest/test_suites/storageperf/setup.go
@@ -172,7 +172,7 @@ func TestSetup(t *imagetest.TestWorkflow) error {
 		testVMs = append(testVMs, vm)
 	}
 	for _, vm := range testVMs {
-		vm.RunTests("TestRandomReadIOPS|TestSequentialReadIOPS|TestRandomWriteIOPS|TestSequentialWriteIOPS")
+		vm.RunTests("TestRandomWriteIOPS|TestSequentialWriteIOPS|TestRandomReadIOPS|TestSequentialReadIOPS")
 	}
 	return nil
 }


a-crate:
This won't actually swap the order (at least not reliably). `go test` does not define the testing order (outside of the `-shuffle` flag but that's obviously not helpful), meaning the go tool/compiler can do whatever it feels like. Example:

```
acrate@acrate-p1g4 ~/g/g/imagetest (workflowclean)> go test -v -run 'TestUseGVNIC|TestAddAliasIPRanges'
=== RUN   TestUseGVNIC
--- PASS: TestUseGVNIC (0.00s)
=== RUN   TestAddAliasIPRanges
--- PASS: TestAddAliasIPRanges (0.00s)
PASS
ok  	github.com/GoogleCloudPlatform/guest-test-infra/imagetest	0.030s
acrate@acrate-p1g4 ~/g/g/imagetest (workflowclean)> go test -v -run 'TestAddAliasIPRanges|TestUseGVNIC'
=== RUN   TestUseGVNIC
--- PASS: TestUseGVNIC (0.00s)
=== RUN   TestAddAliasIPRanges
--- PASS: TestAddAliasIPRanges (0.00s)
PASS
ok  	github.com/GoogleCloudPlatform/guest-test-infra/imagetest	0.030s
acrate@acrate-p1g4 ~/g/g/imagetest (workflowclean)>
```

As long as we are using go test binaries we have no control over test execution order inside the VM.

PR #947, opened 2023-12-01
Update TestGuestPackages

elicriffield:
Update TestGuestPackages to test for all the packages and check by Ach and check for packages that should not be installed

### DIFF
--- a/imagetest/test_suites/packagevalidation/package_test.go
+++ b/imagetest/test_suites/packagevalidation/package_test.go
@@ -4,10 +4,7 @@
 package packagevalidation
 
 import (
-	"bufio"
-	"bytes"
 	"os/exec"
-	"runtime"
 	"strings"
 	"testing"
 
@@ -18,21 +15,29 @@ import (
 func findUniq(arrayA, arrayB []string) []string {
 	unique := []string{}
 	for _, itemA := range arrayA {
-		if !contains(arrayB, itemA) {
+		found := false
+		for _, itemB := range arrayB {
+			if strings.Contains(itemB, itemA) {
+				found = true
+				break
+			}
+		}
+		if !found {
 			unique = append(unique, itemA)
 		}
 	}
 	return unique
 }
 
-// contains checks if a slice contains a particular string.
-func contains(slice []string, str string) bool {
-	for _, v := range slice {
-		if v == str {
-			return true
+// removeFromArray removes a specific string from an array of strings.
+func removeFromArray(arr []string, strToRemove string) []string {
+	var newArr []string
+	for _, item := range arr {
+		if item != strToRemove {
+			newArr = append(newArr, item)
 		}
 	}
-	return false
+	return newArr
 }
 
 func TestStandardPrograms(t *testing.T) {
@@ -77,10 +82,14 @@ func TestGuestPackages(t *testing.T) {
 
 	// What packages to check for and excude for diffrent images
 	requiredPkgs := []string{"google-guest-agent", "google-osconfig-agent"}
-	if strings.Contains(image, "rhel") {
-		requiredPkgs = append(requiredPkgs, "google-compute-engine", "gce-disk-expand", "google-cloud-sdk")
-	}
+	requiredPkgs = append(requiredPkgs, "google-compute-engine", "gce-disk-expand", "google-cloud-sdk")
+	requiredPkgs = append(requiredPkgs, "google-compute-engine-oslogin")
+
 	if strings.Contains(image, "sles") || strings.Contains(image, "suse") {
+		requiredPkgs = removeFromArray(requiredPkgs, "google-cloud-sdk")
+		requiredPkgs = removeFromArray(requiredPkgs, "google-compute-engine")
+		requiredPkgs = removeFromArray(requiredPkgs, "google-compute-engine-oslogin")
+		requiredPkgs = removeFromArray(requiredPkgs, "gce-disk-expand")
 		requiredPkgs = append(requiredPkgs, "google-guest-configs") // SLES name for 'google-compute-engine' package.
 		requiredPkgs = append(requiredPkgs, "google-guest-oslogin")
 	}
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -4,22 +4,45 @@
...
 	"testing"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
+// findUnique returns a slice of elements that are in arrayA but not in arrayB.
+func findUniq(arrayA, arrayB []string) []string {
+	unique := []string{}
+	for _, itemA := range arrayA {
+		if !contains(arrayB, itemA) {
```
a-crate, line 21: I think you can just do `strings.Contains` here instead of re-implementing `contains`, unless I'm missing something.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -38,41 +61,74 @@ func TestStandardPrograms(t *testing.T) {
 func TestGuestPackages(t *testing.T) {
 	utils.LinuxOnly(t)
```
a-crate, line 62: Doesn't need to happen in this PR but we should (at some point) expand this to include windows packages from `googet installed`.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -38,41 +61,74 @@ func TestStandardPrograms(t *testing.T) {
...
+		listPackagesCmd = []string{"dpkg-query", "-W", "--showformat", "${Package}\n"}
+	}
+
+	// What packages to insure are not installed
+	excludePkgs := []string{}
+
+	// What packages to check for and excude for diffrent images
+	requiredPkgs := []string{"google-guest-agent", "google-osconfig-agent"}
+	if strings.Contains(image, "rhel") {
+		requiredPkgs = append(requiredPkgs, "google-compute-engine", "gce-disk-expand", "google-cloud-sdk")
```
a-crate, line 81: Everything except sles should check for `google-cloud-sdk`, right now it's only rhel checking for it.
Same with `gce-disk-expand`, rhel and debian are checking but centos isn't.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -38,41 +61,74 @@ func TestStandardPrograms(t *testing.T) {
...
 	}
-	packages := []string{"google-guest-agent", "google-osconfig-agent"}
-	if strings.Contains(image, "sles") {
-		packages = append(packages, "google-guest-configs") // SLES name for 'google-compute-engine' package.
-		packages = append(packages, "google-guest-oslogin")
-	} else {
```
a-crate, line 52: The stuff in the else here is inconsistently applied in the new code, centos is missing the check for `google-compute-engine` and all of them are missing the check for `google-compute-engine-oslogin`

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -38,41 +61,74 @@ func TestStandardPrograms(t *testing.T) {
...
-		cmd.Stdout = stdout
-		if err := cmd.Start(); err != nil {
-			t.Errorf("error starting check command: %v", err)
-			continue
+		requiredPkgs = append(requiredPkgs, "epel-release")
+	}
+	if strings.Contains(image, "debian") {
+		requiredPkgs = append(requiredPkgs, "haveged", "net-tools")
+		requiredPkgs = append(requiredPkgs, "google-cloud-packages-archive-keyring", "isc-dhcp-client")
+		requiredPkgs = append(requiredPkgs, "google-compute-engine", "gce-disk-expand", "python3", "google-cloud-cli")
```
a-crate, line 93: I think checking for `python3` is a mistake, we don't really care about python we care that the gcloud cli is installed and working which is covered by everything else in this suite. In some hypothetical `python4` future, we don't really care which version gcloud uses as long as it works.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -38,41 +61,74 @@ func TestStandardPrograms(t *testing.T) {
...
+		excludePkgs = append(excludePkgs, "cloud-initramfs-growroot")
+		if runtime.GOARCH == "amd64" {
```
a-crate, line 95: Do we actually need the grub/shim stuff? IMO the specific packages are less import than testing if the image has a working, secureboot capable bootloader which is part of the imageboot suite.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -38,41 +61,74 @@ func TestStandardPrograms(t *testing.T) {
...
 		}
-		if err := cmd.Wait(); err != nil {
-			t.Errorf("error running check command: %v %s %s", err, stdout, stderr)
-			continue
+		if runtime.GOARCH == "arm64" {
+			requiredPkgs = append(requiredPkgs, "grub-efi-arm64")
 		}
+	}
+
+	cmd := exec.Command(listPackagesCmd[0], listPackagesCmd[1:]...)
```
a-crate, line 103: lines 103-121 can be a lot less verbose, something like this works fine and is much easier to follow imo:
```
cmd := exec.Command(listPackagesCmd[0], listPackagesCmd[1:]...)
out, err := cmd.Output()
if err != nil {
	t.Errorf("Failed to execute list packages command: %v", err)
	return
}
installedPkgs := strings.Split(string(out), "\n")
```

imagetest/test_suites/packagevalidation/package_test.go
elicriffield: Yeah that would work better

imagetest/test_suites/packagevalidation/package_test.go
elicriffield: Yeah that'll be coming in a future PR

imagetest/test_suites/packagevalidation/package_test.go
elicriffield: i know debian gets google-cloud-cli and not google-cloud-sdk, but centos should so i'll rework this part so it checks

imagetest/test_suites/packagevalidation/package_test.go
a-crate: It's a depency of the google-guest-agent package

imagetest/test_suites/packagevalidation/package_test.go
elicriffield: oh yeah that's way better

imagetest/test_suites/packagevalidation/package_test.go
elicriffield: thats cool with me, makes this a lot simpler

imagetest/test_suites/packagevalidation/package_test.go
elicriffield: true we explicitly install it, but if the gcloud cli is working its fine, no need to duplicate

elicriffield:
Updated based on review. 
When building the required package list it basically assumes a redhat variant as the default and modifies the list for everything else.

imagetest/test_suites/packagevalidation/package_test.go
a-crate: I'm sorry, I didn't realize I mistyped. I meant `slices.Contains`, it does exactly what you're doing manually.

elicriffield: Ah ok, rewrote to uses slices.Contains

elicriffield:
Updated to use slices.Contains

PR #957, opened 2023-12-05
Use guest-image-private resource_type

a-crate:
/assign @dorileo

### DIFF
--- a/concourse/tasks/render-templates.yaml
+++ b/concourse/tasks/render-templates.yaml
@@ -1,13 +1,7 @@
 platform: linux
 
-resource_types:
-- name: guest-image-private
-  type: registry-image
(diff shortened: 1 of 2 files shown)


concourse/tasks/render-templates.yaml
```diff
@@ -1,13 +1,7 @@
 platform: linux
 
-resource_types:
-- name: guest-image-private
-  type: registry-image
-  source:
-    repository: gcr.io/compute-image-tools/registry-image-forked
-
 image_resource:
-  type: registry-image-private
```
dorileo, line 10: Can we do the other way around and stick to `registry-image-private`? Meaning changing the resource type name.

PR #965, opened 2023-12-06
Test that nvme udev rules are triggering and creating device symlinks from the name

a-crate:


imagetest/test_suites/disk/block_device_naming_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	err = exec.Command("udevadm", "settle").Run()
+	if err != nil {
+		t.Fatal(err)
+	}
+	disks, err := os.ReadDir("/dev/disk/by-id")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var disklist string
```
dorileo, line 28: what if we make disklist an actual list i.e. `var diskNames []string` and then in the final output we do `strings.Join(diskNames, " ")`?

imagetest/test_suites/disk/setup.go
```diff
@@ -30,5 +47,21 @@ func TestSetup(t *imagetest.TestWorkflow) error {
...
+				continue
+			}
+			inst := &daisy.Instance{}
+			inst.MachineType = tc.machineType
+			inst.Name = "test-block-naming-" + tc.machineType
+			vm, err := t.CreateTestVMMultipleDisks([]*compute.Disk{{Name: inst.Name, Type: imagetest.PdBalanced}, {Name: "secondary", Type: imagetest.PdBalanced, SizeGb: 10}}, inst)
+			if err != nil {
+				return err
+			}
+			vm.RunTests("TestBlockDeviceNaming")
```
dorileo, line 63: Non related to this PR note: we could change this `RunTests()` to instead of receiving strings with the tests names receive function pointers so we can check if the test name is correct in build time rather than in run time.

a-crate: Ack. I agree with you that it would be good to move away from this. There are two barriers:
1) The test execution happens inside the VM after downloading the test binary, we can't pass a pointer acrossthis boundary. We need a (probably home grown) system for identifying and referring to test functions.
2) there's a lot of assumptions in CIT around the "build a `go test` binary and execute it inside the VM" architecture

I don't think there's any quick fixes for this, it would involve refactoring the test wrapper and modifying the result fetching/generation at a minimum.

dorileo: Agreed, thanks for clarifying.

koln67:
Looks like comments from @dorileo were resolved

PR #969, opened 2023-12-07
Check for google-cloud-cli on all (non suse) distros, falling back to google-cloud-sdk if neccessary

a-crate:
/cc @elicriffield @zmarano

### DIFF
--- a/imagetest/test_suites/packagevalidation/package_test.go
+++ b/imagetest/test_suites/packagevalidation/package_test.go
@@ -112,7 +112,7 @@ func TestGuestPackages(t *testing.T) {
 	missingPkgs := findUniq(requiredPkgs, installedPkgs)
 	for _, pkg := range missingPkgs {
 		// Accept google-cloud-sdk as a replacement for google-cloud-cli during migration
-		if pkg == "google-cloud-cli" && strings.Contains(installedPkgs, "google-cloud-sdk") {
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -93,14 +93,11 @@ func TestGuestPackages(t *testing.T) {
...
 	}
 	if strings.Contains(image, "debian") {
-		requiredPkgs = removeFromArray(requiredPkgs, "google-cloud-sdk") // debian has google-cloud-cli
-		requiredPkgs = append(requiredPkgs, "google-cloud-cli")
 		requiredPkgs = append(requiredPkgs, "haveged", "net-tools")
 		requiredPkgs = append(requiredPkgs, "google-cloud-packages-archive-keyring", "isc-dhcp-client")
 		excludePkgs = append(excludePkgs, "cloud-initramfs-growroot")
 	}
 	if strings.Contains(image, "ubuntu") {
-		requiredPkgs = removeFromArray(requiredPkgs, "google-cloud-sdk")
```
zmarano, line 103: On Ubuntu, there isn't a package installed. There is a snap for google-cloud-cli (or sdk) instead. Ideally we would look to see if that snap is preseeded however that can be later.

a-crate: Removed the check for ubuntu and added a TODO here.

PR #970, opened 2023-12-08
Update Networkperf Tests

drewhli:
- Make the windows client startup script test the connection to the server first before running iperf
- Re-organize the startup scripts to make more sense, and for easier code maintenance.
- For SUSE images, we download the pre-built binary from the iperf site for x86 images, and build our own for ARM64. This is because iperf only provides x86 binaries for iperf, which does not work on ARM64.
- Increase the timeout period to 5min from 1min, to account for cases where either the server or the client takes a much longer time to spin up than its counterpart.


imagetest/test_suites/networkperf/performance_test.go
```diff
@@ -51,14 +51,19 @@ func TestNetworkPerformance(t *testing.T) {
...
+	resultsArray := strings.Split(results, " ")
+	result_perf, err := strconv.ParseFloat(resultsArray[5], 64)
 	if err != nil {
 		t.Fatalf("Error: %v", err)
 	}
 
+	// Check the units.
+	units := resultsArray[6]
+	if !strings.HasPrefix(units, "G") { // If the units aren't in Gbits/s, we automatically fail.
+		t.Fatalf("Error: Did not meet performance expectation on machine type %s with network %s. Expected: %v Gbits/s, Actual: %v %s", machineTypeName, network, expected, result_perf, units)
```
a-crate, line 64: The error message should reflect the actual reason we failed, ie that the units of measurement are not what we expected

imagetest/test_suites/networkperf/setup.go
```diff
@@ -95,12 +95,14 @@ var scripts embed.FS
 var targets embed.FS
 
 const (
```
a-crate, line 97: I realize this is both pedantic and pre-exisitng but technically these are URIs not URLs since they're on the local filesystem.

I only bring this up because the `startup-script-url` is the thing that we to execute the test wrapper and it's slightly confusing to reference the script on the filesystem as `linuxInstallStartupScriptURL` but then we actually execute the `linuxInstallStartupScriptURL` as the `startup-script` and execute the wrapper as the `startup-script-url`. This confused me at first when I read the section lower down on reading `linuxInstallStartupScriptURL` and I thought you were switching the mechanism from `startup-script` to `startup-script-url`. I think referencing these as URIs would clear up some potential confusion.

drewhli: Done.

imagetest/test_suites/networkperf/performance_test.go
drewhli: Done.

a-crate (approved):

PR #971, opened 2023-12-09
Windows account password test

koln67:
Test the guest agent functionality that a new windows user can be created: more related username tests based on the existing internal tests will be added in the future.

### DIFF
--- a/imagetest/test_suites/guestagent/windows_account_test.go
+++ b/imagetest/test_suites/guestagent/windows_account_test.go
@@ -1,0 +1,51 @@
+//go:build cit
+// +build cit
+
+package packagevalidation
+
+import (
+	"fmt"
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+	"strings"
+	"testing"
+)
+
+// Verifies that a powershell command ran with no errors and exited with an exit code of 0.
+// If a username or password was invalid, this should result in a testing error.
+// Returns the standard output in case it needs to be used later.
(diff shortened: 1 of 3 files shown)


imagetest/test_suites/guestagent/windows_account_test.go
```diff
@@ -0,0 +1,51 @@
+//go:build cit
+// +build cit
+
+package packagevalidation
```
a-crate, line 4: this should be the guestagent package

imagetest/test_suites/guestagent/windows_account_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return stdout
+}
+
+func TestCreateNewWindowsUser(t *testing.T) {
+	utils.WindowsOnly(t)
+	username := "windowsuser5"
+	password := "gyug3q445m0!"
+
+	createUserCmd := fmt.Sprintf("net user %s %s /add", username, password)
```
a-crate, line 34: Is this what we need to use? This is basically an OS level feature with no guest environment interaction, I was under the impression we need to test the automated windows password reset feature here:

https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation

https://github.com/GoogleCloudPlatform/compute-image-windows/blob/HEAD/examples/windows_auth_go_sample.go

jjerger: Correct - we are aiming to test the guest agent functionality of generating a password and setting it on the guest.

PR #975, opened 2023-12-12
Update README

drewhli:
- Change the imagetest README to be hopefully more readable.
- Change `TestCVMEnabled` under the CVM section to reflect the new `TestSEVEnabled` names.
- Add a section for the OSLogin test suite.
- Add a README to the OSLogin test suite for setting up projects/users for OSLogin testing.

A screenshot of the new README look can be seen below. Do not mind the thing next to the disk test suite header, that will not be there on the actual view.
Screenshot 2023-12-12 at 12 37 53


imagetest/test_suites/README.md
```diff
@@ -11,64 +11,52 @@ new GCE image.
...
 Tests are broken down by suite below:
 
-## Test suites ##
+## Test Suites
 
-### Test suite: cvm ###
+### Test suite: cvm
 
-#### TestCVMEnabled ####
+#### TestSEVEnabled
```
a-crate, line 18: Can you change this to TestSEVEnabled/TestSEVSNPEnabled/TestTDXEnabled? I also think we should be more explicit about what the test is actually doing, which is making sure the instance can boot with the respective confidential instance type set in the config and that it can load it's guest kernel module.

imagetest/test_suites/oslogin/README.md
```diff
@@ -0,0 +1,36 @@
+# Test Setup
```
a-crate, line 1: Is this everything needed to set up oslogin testing? Is a separate test org required and does it need any specific policies?

drewhli: I believe so, yes. A separate test org is not technically required. It does not need any specific policies, either, outside of allowing the person doing the setup to create secrets and provide IAM permissions to the test users. A test org would, however, have the advantages of 1) a clean environment, and 2) being able to easily the test users required. I can include these extra details in the setup page.

imagetest/test_suites/README.md
drewhli: Done.

imagetest/test_suites/oslogin/README.md
drewhli: Done.

a-crate (approved):

PR #978, opened 2023-12-13
Update README.md

koln67:
imagevalidation was split into different test suites, update the docs to reflect the change

### DIFF
--- a/imagetest/test_suites/README.md
+++ b/imagetest/test_suites/README.md
@@ -61,7 +61,7 @@ output of `/bin/hostname -f`. See `man 1 hostname` for more details.
 #### TestCustomHostname
 Test that custom domain names are correctly set.
 
-- <b>Background</b>: The domain name for a VM matches the name of the VM by default. However, if you
+- <b>Background</b>: The domain name for a VM matches the configured internal GCE DNS setting (https://cloud.google.com/compute/docs/internal-dns). By default, this is the VM name. However, if you
 specify a custom domain name at instance creation time, this will be used instead.
 
 - <b>Test logic</b>: Launch a VM with a custom domain name. Validate the domain name as with TestFQDN.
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/README.md
```diff
@@ -32,7 +32,54 @@ new space.
...
+
+#### TestFQDN
+Test that the fully-qualified domain name is correctly set.
+
+- <b>Background</b>: The FQDN is a complicated concept in Linux operating systems, and setting it in
+an incorrect way can lead to unexpected behavior in some software.
+
+- <b>Test logic</b>: Retrieve the intended FQDN from metadata and compare the full value to the
+output of `/bin/hostname -A`. See `man 1 hostname` for more details.
+
```
a-crate, line 60: we actually compare to `hostname -f`

imagetest/test_suites/README.md
```diff
@@ -32,7 +32,54 @@ new space.
...
+#### TestCustomHostname
+Test that custom hostnames are correctly set.
+
+- <b>Background</b>: The hostname for a VM matches the name of the VM by default. However, if you
+specify a custom hostname at instance creation time, this will be used instead.
```
a-crate, line 65: The test is called customhostname but what we are actually testing should be referred to as a custom domain name, not a custom hostname.

imagetest/test_suites/README.md
koln67: I updated the description

imagetest/test_suites/README.md
```diff
@@ -32,7 +32,54 @@ new space.
...
+- <b>Background</b>: The FQDN is a complicated concept in Linux operating systems, and setting it in
+an incorrect way can lead to unexpected behavior in some software.
+
+- <b>Test logic</b>: Retrieve the intended FQDN from metadata and compare the full value to the
+output of `/bin/hostname -f`. See `man 1 hostname` for more details.
+
+#### TestCustomHostname
+Test that custom domain names are correctly set.
+
+- <b>Background</b>: The domain name for a VM matches the name of the VM by default. However, if you
```
a-crate, line 64: Sorry one more thing. I don't think I can suggest edits in github but "The domain name for a VM matches the name of the VM by default." should actually be something like "The domain name for a VM matches the configured internal GCE DNS setting" with maybe a link to [this](https://cloud.google.com/compute/docs/internal-dns) page.
otherwise lgtm

PR #985, opened 2023-12-14
Add windows password test

koln67:
Tests windows automated password generation, similar to the example: https://cloud.google.com/compute/docs/instances/windows/automate-pw-generation

In this test, it first creates the user, then runs the example code, and finally validates that the new password for the user is correct.

### DIFF
--- a/imagetest/test_suites/hotattach/hot_attach_test.go
+++ b/imagetest/test_suites/hotattach/hot_attach_test.go
@@ -18,31 +18,6 @@ import (
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
-func getProjectAndZone() (string, string, error) {
-	var projectZoneUrl string
-	if runtime.GOOS == "windows" {
-		procStatus, err := utils.RunPowershellCmd("Invoke-RestMethod -Headers @{'Metadata-Flavor' = 'Google'} -Uri \"http://metadata.google.internal/computeMetadata/v1/instance/zone\"")
-		if err != nil {
-			return "", "", fmt.Errorf("failed to get project or zone on windows: %v", err)
-		}
-		projectZoneUrl = strings.TrimSpace(procStatus.Stdout)
-	} else {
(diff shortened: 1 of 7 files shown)


imagetest/utils/test_utils.go
```diff
@@ -199,6 +199,33 @@ func GetProjectZone(ctx context.Context) (string, string, error) {
...
+func GetInstanceZoneProject() (string, string, string, error) {
+	var fqdnString string
+	if runtime.GOOS == "windows" {
+		procStatus, err := RunPowershellCmd("Invoke-RestMethod -Headers @{'Metadata-Flavor' = 'Google'} -Uri \"http://metadata.google.internal/computeMetadata/v1/instance/hostname\"")
+		if err != nil {
+			return "", "", "", fmt.Errorf("failed to get project, zone, or instance on windows: %v", err)
+		}
+		fqdnString = strings.TrimSpace(procStatus.Stdout)
+	} else {
+		fqdnBytes, err := exec.Command("curl", "http://metadata.google.internal/computeMetadata/v1/instance/hostname", "-H", "Metadata-Flavor: Google").Output()
```
a-crate, line 212: This mechanism seems kind of fragile, since it relies on DNS naming conventions. It will break if project level DNS is not set to zonal or if the instance edits DNS settings or uses a custom hostname, which is something we do in some tests. We also have `instance/zone`, `project/project-id`, and `instance/name` keys to tell use these things authoritatively.

imagetest/test_suites/hotattach/hot_attach_test.go
```diff
@@ -215,28 +215,23 @@ func TestFileHotAttach(t *testing.T) {
...
 		}
 	}
-	instName, err := exec.Command("hostname").Output()
+	instName, instanceZone, projectNumber, err := utils.GetInstanceZoneProject()
 	if err != nil {
-		t.Fatalf("failed to get hostname: %v %v", instName, err)
+		t.Fatalf("failed to get instance, zone, or project details: error %v", err)
 	}
 
-	projectNumber, instanceZone, err := getProjectAndZone()
```
a-crate, line 223: I don't think there's any remaining callers of this function after deleting this, just delete the function too

koln67:
/hold
will unhold once i'm back

a-crate (approved):

PR #990, opened 2023-12-18
Test that startup and shutdown scripts are not executed on agent reinstallation

a-crate:
/cc @zmarano @koln67

### DIFF
--- a/imagetest/test_suites/metadata/shutdown_script_test.go
+++ b/imagetest/test_suites/metadata/shutdown_script_test.go
@@ -4,6 +4,10 @@
 package metadata
 
 import (
+	"path"
+	"os/exec"
+	"io"
+	"time"
 	"fmt"
 	"io/ioutil"
 	"strings"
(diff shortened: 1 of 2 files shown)


imagetest/test_suites/metadata/shutdown_script_test.go
```diff
@@ -71,14 +75,59 @@ func testShutdownScriptTimeWindows() error {
...
+// by checking the output content of the Shutdown script. It also checks that
+// shutdown scripts don't run on reinstall or upgrade of the guest-agent.
 func TestShutdownScripts(t *testing.T) {
 	result, err := utils.GetMetadata(utils.Context(t), "instance", "guest-attributes", "testing", "result")
 	if err != nil {
 		t.Fatalf("failed to read shutdown script result key: %v", err)
 	}
 	if result != expectedShutdownContent {
-		t.Fatalf(`shutdown script output expected "%s", got "%s".`, expectedShutdownContent, result)
+		t.Errorf(`shutdown script output expected "%s", got "%s".`, expectedShutdownContent, result)
```
koln67, line 86: which checks are expected to use t.Errorf instead of t.Fatalf? Is it only when the output is there but different from what is expected?

a-crate: Yea, I used t.Fatal for failure to setup the conditions for the test and t.Error for when the test was setup correctly but it didn't pass

PR #993, opened 2023-12-21
Update guestagent suite

a-crate:
Test that pre and post snapshot scripts are executed when a snapshot request with guestflush is sent The windows equivalent of this is on OS feature doesn't have any guest environment components

Fix a bug with the telemetry test where it was triggering incorrectly and restart the agent after changing telemetry settings until we reevaluate agent telemetry behavior

/cc @dorileo @koln67 @drewhli


imagetest/test_suites/guestagent/snapshot_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Fatal(err)
+	}
+
+	// Enable snapshot scripts in the agent and restart it.
+	// Wait 5 seconds for connection to snapshot service
+	agentcfg, err := os.ReadFile("/etc/default/instance_configs.cfg")
+	if err != nil {
+		t.Fatal(err)
+	}
+	agentcfg = append(agentcfg, []byte("[Snapshots]\nenabled = true\ntimeout_in_seconds = 300\n")...)
```
dorileo, line 62: why timeout of 300s?

imagetest/test_suites/guestagent/setup.go
```diff
@@ -12,24 +12,47 @@
...
+	telemetryenabledinst.Name = "telemetry-enabled"
+	telemetryenabledvm, err := t.CreateTestVMMultipleDisks([]*compute.Disk{{Name: telemetryenabledinst.Name, Type: imagetest.PdBalanced}}, telemetryenabledinst)
 	if err != nil {
 		return err
 	}
-	telemetryenabled.RunTests("TestTelemetryEnabled") // Enabled by default
+	telemetryenabledvm.AddMetadata("disable-guest-telemetry", "false")
+	telemetryenabledvm.RunTests("TestTelemetryEnabled")
+
+	if !utils.HasFeature(t.Image, "WINDOWS") {
```
dorileo, line 47: We are not running the snapshot test for windows, why?

imagetest/test_suites/guestagent/telemetry_test.go
```diff
@@ -0,0 +1,132 @@
...
+	out, err := cmd.Output()
+	if err != nil {
+		t.Fatalf("could not get agent output: %v", err)
+	}
+	return string(out)
+}
+
+func TestTelemetryEnabled(t *testing.T) {
+	output := getAgentOutput(t)
+	if !strings.Contains(output, "Successfully scheduled job telemetryJobID") {
```
dorileo, line 56: side note: in the future we can leverage the command monitor to tell the status of subsystems.

imagetest/test_suites/guestagent/telemetry_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	daisyCompute "github.com/GoogleCloudPlatform/compute-daisy/compute"
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+func restartAgent(t *testing.T) {
+	var cmd *exec.Cmd
+	if utils.IsWindows() {
+		cmd = exec.Command("powershell.exe", "-NonInteractive", "Restart-Service", "GCEGuestAgent")
```
dorileo, line 30: please use `exec.CommandContext()` everywhere.

imagetest/test_suites/guestagent/telemetry_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	daisyCompute "github.com/GoogleCloudPlatform/compute-daisy/compute"
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
+)
+
+func restartAgent(t *testing.T) {
+	var cmd *exec.Cmd
```
dorileo, line 28: this is a helper, please use `t.Helper()`.

imagetest/test_suites/guestagent/telemetry_test.go
```diff
@@ -0,0 +1,132 @@
...
+	} else {
+		cmd = exec.Command("systemctl", "restart", "google-guest-agent")
+	}
+	err := cmd.Run()
+	if err != nil {
+		t.Fatalf("could not restart agent: %v", err)
+	}
+}
+
+func getAgentOutput(t *testing.T) string {
```
dorileo, line 40: ditto for `t.Helper()`

imagetest/test_suites/guestagent/telemetry_test.go
```diff
@@ -0,0 +1,132 @@
...
+	var cmd *exec.Cmd
+	if utils.IsWindows() {
+		cmd = exec.Command("powershell.exe", "-NonInteractive", "Get-WinEvent", "-Providername", "GCEGuestAgent")
```
dorileo, line 43: ditto for `exec.CommandContext()`

imagetest/test_suites/guestagent/snapshot_test.go
a-crate: I set it to the maximum allowable to avoid flakes due to snapshot scripts taking longer than expected (not that I'm expecting this).

Worth nothing that 
1) test timeouts for CIT executions are orders of magnitude longer than this. 15 minutes would be a short timeout for CIT
2) we don't pass the test suite timeout down to the wrapper inside, so inside the VM we have no idea what the timeout is. The `utils.Context` wrapper actually always returns context.Background because of CIT semantics

imagetest/test_suites/guestagent/setup.go
a-crate: VSS snapshots are the equivalent on windows but there is not guest environment component, it's entirely an OS feature

imagetest/test_suites/guestagent/snapshot_test.go
dorileo: sounds good to me.

imagetest/test_suites/guestagent/setup.go
dorileo: Yep, true that. Sounds good. Thank you.

imagetest/test_suites/guestagent/telemetry_test.go
a-crate: That will be good. Much more reliable than this, I fear it will be flaky. It works for now at least.

PR #998, opened 2024-01-03
Add a fallback command to retry reinstalling guest agent during startup/shutdown script testing

a-crate:
dnf and yum don't have a command that forces a reinstall or an upgrade
Also swapped the argument order for zypper, newer version require the --force flag after the install command
/cc @koln67 @drewhli

### DIFF
--- a/imagetest/test_suites/metadata/startup_script_test.go
+++ b/imagetest/test_suites/metadata/startup_script_test.go
@@ -103,7 +103,7 @@ func TestStartupScripts(t *testing.T) {
 			t.Fatalf("could not reinstall guest-agent: %s", err)
 		}
 	} else {
-		var cmd,fallback *exec.Cmd
+		var cmd, fallback *exec.Cmd
 		switch {
 		case utils.CheckLinuxCmdExists("apt"):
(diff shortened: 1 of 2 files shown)


imagetest/test_suites/metadata/startup_script_test.go
```diff
@@ -103,21 +103,29 @@ func TestStartupScripts(t *testing.T) {
...
 		}
 	} else {
-		var cmd *exec.Cmd
+		var cmd,fallback *exec.Cmd
 		switch {
 		case utils.CheckLinuxCmdExists("apt"):
 			cmd = exec.Command("apt", "reinstall", "-y", "google-guest-agent")
 		case utils.CheckLinuxCmdExists("dnf"):
 			cmd = exec.Command("dnf", "-y", "reinstall", "google-guest-agent")
+			cmd = exec.Command("dnf", "-y", "upgrade", "google-guest-agent")
```
koln67, line 112: minor typo: the second line probably should be fallback = exec.Command("dnf", "-y", "upgrade", "google-guest-agent")

a-crate: Woops, yeah.

PR #1002, opened 2024-01-08
Hyperdisk test case commands

koln67:
Implement hyperdisk benchmarking instructions, as the doc was recently updated at https://cloud.google.com/compute/docs/disks/benchmark-hyperdisk-performance

### DIFF
--- a/imagetest/test_suites/storageperf/iops_write_test.go
+++ b/imagetest/test_suites/storageperf/iops_write_test.go
@@ -95,7 +95,7 @@ func RunFIOWriteLinux(t *testing.T, mode string) ([]byte, error) {
 			return []byte{}, fmt.Errorf("fio installation on linux failed: err %v", err)
 		}
 		if usingHyperdisk {
-			err = fillDisk(symlinkRealPath, ctx)
+			err = fillDisk(ctx, symlinkRealPath)
 			if err != nil {
 				return []byte{}, fmt.Errorf("fill disk preliminary step failed: err %v", err)
 			}
--- a/imagetest/test_suites/storageperf/iops_write_test.go
+++ b/imagetest/test_suites/storageperf/iops_write_test.go
@@ -20,7 +20,7 @@ func TestRandomWriteIOPS(t *testing.T) {
 	var err error
 	if runtime.GOOS == "windows" {
 		if randWriteIOPSJson, err = runFIOWindows(randWrite); err != nil {
-			t.Fatalf("windows fio rand write failed with error: %v", err)
+			t.Fatalf("windows fio rand write failed with error: %v. If testing locally, check the guidance at storageperf/startupscripts/install_fio.ps1", err)
 		}
 	} else {
 		if randWriteIOPSJson, err = runFIOLinux(t, randWrite); err != nil {
--- a/imagetest/test_suites/storageperf/iops_write_test.go
+++ b/imagetest/test_suites/storageperf/iops_write_test.go
@@ -95,7 +95,7 @@ func RunFIOWriteLinux(t *testing.T, mode string) ([]byte, error) {
 			return []byte{}, fmt.Errorf("fio installation on linux failed: err %v", err)
 		}
 		if usingHyperdisk {
-			err = fillDisk(symlinkRealPath, ctx, t)
+			err = fillDisk(symlinkRealPath, ctx)
 			if err != nil {
 				return []byte{}, fmt.Errorf("fill disk preliminary step failed: err %v", err)
 			}
--- a/imagetest/test_suites/storageperf/iops_write_test.go
+++ b/imagetest/test_suites/storageperf/iops_write_test.go
@@ -19,11 +19,11 @@ func TestRandomWriteIOPS(t *testing.T) {
 	var randWriteIOPSJson []byte
 	var err error
 	if runtime.GOOS == "windows" {
-		if randWriteIOPSJson, err = RunFIOWindows(randWrite); err != nil {
+		if randWriteIOPSJson, err = runFIOWindows(randWrite); err != nil {
 			t.Fatalf("windows fio rand write failed with error: %v", err)
 		}
 	} else {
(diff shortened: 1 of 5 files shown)


imagetest/test_suites/storageperf/storage_perf_utils.go
```diff
@@ -207,3 +214,113 @@ func checkZypperTransientError(err error, stdout, stderr string) error {
...
+	}
+	if diskType == imagetest.HyperdiskExtreme || diskType == imagetest.HyperdiskThroughput || diskType == imagetest.HyperdiskBalanced {
+		return true
+	}
+
+	return false
+}
+
+// function to get num numa nodes
+func getNumNumaNodes() (int, error) {
```
a-crate, line 232: This won't work on windows

imagetest/test_suites/storageperf/storage_perf_utils.go
```diff
@@ -207,3 +214,113 @@ func checkZypperTransientError(err error, stdout, stderr string) error {
...
+}
+
+// function to get cpu mapping as strings if there is only one numa node
+// returned format is queue_1_cpus, queue_2_cpus, error
+func getCPUNvmeMapping(symlinkRealPath string) (string, string, error) {
+	cpuListCmd := exec.Command("cat", "/sys/class/block/"+symlinkRealPath+"/mq/*/cpu_list")
+	cpuListBytes, err := cpuListCmd.CombinedOutput()
+	if err != nil {
+		return "", "", err
+	}
```
a-crate, line 265: Ditto windows support

imagetest/test_suites/storageperf/storage_perf_utils.go
```diff
@@ -207,3 +214,113 @@ func checkZypperTransientError(err error, stdout, stderr string) error {
...
+	queue1Cpus := strings.TrimSpace(cpuListOutLines[0])
+	queue2Cpus := strings.TrimSpace(cpuListOutLines[1])
+	return queue1Cpus, queue2Cpus, nil
+}
+
+// fill the disk before testing to reach the maximum read iops and bandwidth
+// TODO: implement this for windows by passing in the \\\\.\\PhysicalDrive1 parameter
+func fillDisk(symlinkRealPath string) error {
+	if runtime.GOOS == "windows" {
+		fmt.Println("fill disk preliminary step not yet implemented for windows")
```
a-crate, line 280: Can you call `t.Logf` with this message so that we can surface the fact that it's missing when the performance is degraded?

imagetest/test_suites/storageperf/storage_perf_utils.go
```diff
@@ -207,3 +214,113 @@ func checkZypperTransientError(err error, stdout, stderr string) error {
...
+	} else {
+		// hard coding the filesize to 500G as that conforms to the docs while giving
+		// sufficiently high performance
+		fillDiskCmdOptions := fillDiskCommonOptions + " --filesize=500G --filename=" + symlinkRealPath
```
a-crate, line 284: Why don't we fill the entire FS? Because it takes too long, or something else?

imagetest/test_suites/storageperf/iops_write_test.go
```diff
@@ -16,8 +16,10 @@ import (
 )
 
 const (
```
a-crate, line 18: A lot of these parameters are repeated in each definition, is there any way to separate this into standardFioOptions, randomFioOptions, sequentialFioOptions, etc and just append each list of options when we need it in getLinuxWriteOptions?

a-crate: Ideally, do the for read options if possible

imagetest/test_suites/storageperf/storage_perf_utils.go
koln67: This is a large PR, so I'll plan on a follow up PR for windows hyperdisk

imagetest/test_suites/storageperf/iops_write_test.go
koln67: I refactored some helper methods so that it uses common options and methods for most of the test

imagetest/test_suites/storageperf/storage_perf_utils.go
koln67: right now, the reason is because it takes too long.

After looking at what the PD team does, it might be possible to have a larger batch size and fill a larger file size, though I'll explore that more once I get pd-balanced to reach the performance targets.

imagetest/test_suites/storageperf/storage_perf_utils.go
koln67: I added a log for this case as well as the generic pd case where the fill disk step has not yet been implemented.

a-crate:
This breaks the test on windows images:
```
	<testsuite name="storageperf-windows-2016" tests="24" failures="24" errors="0" disabled="0" skipped="0" time="0">
		<testcase classname="storageperf-windows-2016" name="TestRandomReadIOPS" time="0.32">
			<failure type="">    iops_read_test.go:23: windows fio rand read failed with error: fio.exe returned with error: exit status 1  Start-Process : This command cannot be run due to the error: The system cannot &#xA;        find the file specified.&#xA;        At line:1 char:1&#xA;        + Start-Process C:\fio.exe -ArgumentList &#34;--name=write_iops_test --file ...&#xA;        + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~&#xA;            + CategoryInfo          : InvalidOperation: (:) [Start-Process], InvalidOp &#xA;           erationException&#xA;            + FullyQualifiedErrorId : InvalidOperationException,Microsoft.PowerShell.C &#xA;           ommands.StartProcessCommand&#xA;         </failure>
		</testcase>
[...]
```

a-crate:
Huh, I guess I've never run storageperf manually before, I thought it was working outside of the testgrid for some reason. My mistake
/lgtm

PR #1004, opened 2024-01-08
Disable centos-7-predictable-nic-builds

a-crate:
We don't need new images to continue testing the existing ones. The evidence looks pretty conclusive in favor of this being the culprit already.

/hold


a-crate:
DeleteAfter is set to 6 months, so we will still have the existing images for plenty of time while we test. Periodics will keep running.
/unhold
/cc @zmarano @elicriffield

PR #1008, opened 2024-01-09
package_test: define a generic package rules structure

dorileo:
This change introduces a generic package rules structure, it supports checking for package names alternatives (different distros may have different namings) and defines rules for skipping a given image (based on a expression matching) and also rules for exact image matching (working in the opposite direction of the skipping rules, it will only apply if the image name expression matches).

### DIFF
--- a/imagetest/test_suites/packagevalidation/package_test.go
+++ b/imagetest/test_suites/packagevalidation/package_test.go
@@ -7,34 +7,37 @@ import (
 	"fmt"
 	"os/exec"
 	"regexp"
-	"slices"
 	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
-// findUnique returns a slice of elements that are in arrayA but not in arrayB.
-func findUniq(arrayA, arrayB []string) []string {
-	unique := []string{}
-	for _, itemA := range arrayA {
-		if !slices.Contains(arrayB, itemA) {
-			unique = append(unique, itemA)
-		}
-	}
+// osPackage defines the rules for expected installed packages.
+type osPackage struct {
+	// name is the name of the package, a package could have alternative names
+	// depending on the distro, see alternatives field.
+	name string
 
-	return unique
-}
+	// shouldNotBeInstalled defines if we are checking if the package should or
+	// should not be installed.
+	shouldNotBeInstalled bool
 
-// removeFromArray removes a specific string from an array of strings.
-func removeFromArray(arr []string, strToRemove string) []string {
-	var newArr []string
-	for _, item := range arr {
-		if item != strToRemove {
-			newArr = append(newArr, item)
-		}
-	}
-	return newArr
+	// alternatives are alternative names, a package can be named differently
+	// depending on the distribution.
+	alternatives []string
+
+	// imagesSkip are the image name matching expression for images we don't want
+	// to check this package rule.
+	// The expression matching is applied with strings.Contains() so if the image
+	// name contains the substring it will match.
+	imagesSkip []string
+
+	// images is the oposite of imagesSkip and defines the image name matching
+	// expression of the images this rule must apply.
+	// The expression matching is applied with strings.Contains() so if the image
+	// name contains the substring it will match.
+	images []string
 }
 
 func TestStandardPrograms(t *testing.T) {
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -151,15 +172,45 @@ func TestGuestPackages(t *testing.T) {
...
+
+		imageMatched := len(curr.images) == 0
+		for _, matchExpression := range curr.images {
+			if strings.Contains(image, matchExpression) {
+				imageMatched = true
+			}
+		}
+
+		if skipPackage || !imageMatched {
+			break
```
a-crate, line 197: `s/continue/break/`, no? This way we continue to the next package rather than skipping the rest of them.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -110,39 +109,61 @@ func TestGuestPackages(t *testing.T) {
...
+		},
+		&osPackage{
+			name:   "google-cloud-packages-archive-keyring",
+			images: []string{"debian"},
+		},
+		&osPackage{
+			name:   "isc-dhcp-client",
+			images: []string{"debian"},
+		},
+		&osPackage{
```
a-crate, line 163: This package rule requires that `cloud-initramfs-growroot` is installed on debian if I'm reading the logic correctly, we want it to not be installed.

Maybe a package rule item can represent desire installation state as a boolean.

dorileo: Huuum, great, I missed that detail, I'll adjust.

imagetest/test_suites/packagevalidation/package_test.go
dorileo: exactly, good catch. thank you.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -151,15 +172,45 @@ func TestGuestPackages(t *testing.T) {
 		return
 	}
 
-	missingPkgs := findUniq(requiredPkgs, installedPkgs)
-	for _, pkg := range missingPkgs {
-		t.Errorf("Required package not installed: %s\n", pkg)
+	installedMap := make(map[string]bool)
+	for _, curr := range strings.Split(string(out), "\n") {
```
a-crate, line 176: I think `out` is undefined here, this should loop over installedPkgs, or modify `listPkgs` functions to return maps instead.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -110,39 +109,61 @@ func TestGuestPackages(t *testing.T) {
...
+			name:         "google-cloud-cli",
+			alternatives: []string{"google-cloud-sdk"},
+		},
+		&osPackage{
+			name:       "google-compute-engine-oslogin",
+			imagesSkip: []string{"sles", "suse"},
+		},
+		&osPackage{
+			name:   "epel-release",
+			images: []string{"centos-7"},
```
a-crate, line 145: There is a bug in the existing logic, we should check this on `rhel-7` too.

imagetest/test_suites/packagevalidation/package_test.go
dorileo: using installedPackages should be enough (last minute rebase, lost track of this change). Thanks for pointing that out.

dorileo: done. Thank you.

imagetest/test_suites/packagevalidation/package_test.go
dorileo: done. thank you.

imagetest/test_suites/packagevalidation/package_test.go
dorileo: done. thank you.

imagetest/test_suites/packagevalidation/package_test.go
dorileo: done. thank you.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -110,56 +113,113 @@ func TestGuestPackages(t *testing.T) {
...
+		packageNames = append(packageNames, curr.alternatives...)
+
+		for _, currPackage := range packageNames {
+			if _, found := installedMap[currPackage]; found {
+				packageInstalled = true
+				break
+			}
+		}
+
+		if curr.shouldNotBeInstalled && packageInstalled {
```
a-crate, line 217: There's a lot of logic here for checking the same two variables, can we just do something like
```
if !curr.shouldNotBeinstalled != packageInstalled {
   t.Errorf("package %s has wrong installation state, got %t want %t [..]
```

dorileo: Yep, we can do that.

dorileo: done. thanks.

a-crate (approved):

PR #1012, opened 2024-01-10
Don't create windowscontainers or sql VMs if the image being tested isn't windows

a-crate:
/cc @drewhli @koln67

### DIFF
--- a/imagetest/test_suites/sql/setup.go
+++ b/imagetest/test_suites/sql/setup.go
@@ -4,6 +4,7 @@ import (
 	"embed"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
 // Name is the name of the test package. It must match the directory name.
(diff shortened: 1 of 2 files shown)


imagetest/test_suites/sql/setup.go
```diff
@@ -28,67 +29,68 @@ const (
 
 // TestSetup sets up the test workflow.
 func TestSetup(t *imagetest.TestWorkflow) error {
```
drewhli, line 31: We can skip the test instead if the image isn't windows.

imagetest/test_suites/windowscontainers/setup.go
```diff
@@ -2,16 +2,19 @@ package windowscontainers
...
 import (
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest"
+	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
 // Name is the name of the test package. It must match the directory name.
 var Name = "windowscontainers"
 
 // TestSetup sets up the test workflow.
 func TestSetup(t *imagetest.TestWorkflow) error {
```
drewhli, line 12: DItto above.

imagetest/test_suites/sql/setup.go
a-crate: For other tests where there's nothing applicable to the image we've been avoiding creating the image at all rather than skipping, to keep results publish to the test grid limited to stuff that's relevant to test.

drewhli: That makes sense. Thanks for clarifying.

drewhli (approved):

PR #1018, opened 2024-01-17
Pdbalancefill

koln67:
https://cloud.google.com/compute/docs/disks/performance#zonal

PD balanced disks can only provide 6 iops per GB.

This change calculates the approximate disk size needed to reach the hard coded performance limit and allocates a disk of that size.

### DIFF
--- a/imagetest/test_suites/packagevalidation/package_test.go
+++ b/imagetest/test_suites/packagevalidation/package_test.go
@@ -114,58 +114,58 @@ func TestGuestPackages(t *testing.T) {
 	}
 
 	pkgs := []*osPackage{
-		&osPackage{
+		{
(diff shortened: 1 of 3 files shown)


koln67:
/hold

Wait until after disk size quota has been addressed

koln67:
/unhold
Quota was increased from 200 TB to 250 TB

a-crate (approved):

a-crate:
I think `compute-image-test-pool-001` is still at 100TB, can you request that one be bumped to 250 as well so we can run it manually?

koln67:
> I think `compute-image-test-pool-001` is still at 100TB, can you request that one be bumped to 250 as well so we can run it manually?

done

PR #1021, opened 2024-01-17
In windows security tests, add an adminstrator to the test vm and check the ssh config in reference to that user

a-crate:
/cc @drewhli @elicriffield

### DIFF
--- a/imagetest/test_suites/security/image_security_test.go
+++ b/imagetest/test_suites/security/image_security_test.go
@@ -135,7 +135,7 @@ func TestPasswordSecurity(t *testing.T) {
 		t.Fatalf("couldn't get image from metadata")
 	}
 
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/security/image_security_test.go
```diff
@@ -195,13 +195,23 @@ func verifyPassword(ctx context.Context) error {
 	return nil
 }
 
-func verifySSHConfig(image string) error {
+func verifySSHConfig(t *testing.T, image string) error {
```
drewhli, line 198: Should we be using `t.Helper()` here?

a-crate: Probably. Done.

drewhli (approved):

PR #1028, opened 2024-01-22
Ubuntu 16 and 18 hyperdisk fix

koln67:
On ubuntu 16 and 18, the --numa_cpu_nodes option in the fio command does not work with the default fio installation. 

After removing that option and adding the --name option to ensure the fio command compiles, the hyperdisk tests seem to still reach the performance targets on ubuntu 16 and 18.


a-crate:
These packages are available in the repos for ubuntu 16 and 18, can we just install them instead so we can take advantage of the full performance?

PR #1029, opened 2024-01-22
Add LSSD support and cases to storageperf

a-crate:
LSSDs present on the machine are collected into a raid array as directed in our public documentation

I abstracted installFioLinux to install any package and used that to install mdadm for the array setup, then added some logic to install libnuma-dev on ubuntu 1604 and 1804 since this came up just now and I already wrote that abstraction

Added performance targets for LSSDs

Added some logic to determine "disk class" eg. PD, hyperdisk, or LSSD and reworked logic to return fio options based off that

getMountDiskSize returns 0 for LSSD so that it doesn't wait for quota it doesn't need

Test setup logic just checks of diskType is lssd and skips adding a secondary disk if so.

/cc @koln67 @zmarano

### DIFF
--- a/imagetest/test_suites/storageperf/iops_read_test.go
+++ b/imagetest/test_suites/storageperf/iops_read_test.go
@@ -19,7 +19,7 @@ func TestRandomReadIOPS(t *testing.T) {
 	var randReadIOPSJson []byte
 	var err error
 	if runtime.GOOS == "windows" {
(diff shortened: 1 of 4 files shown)


a-crate:
Also using `-lssd` machine types breaks the assumption that we can get the CPU count with the `[0-9]+$` regex, so I switched to fetching the machine type and checking the CPU count through the compute API.

PR #1032, opened 2024-01-23
Switch disk types to hyperdisk extreme for machine types that don't support PD

a-crate:
/cc @drewhli @zmarano 

This is extremely ugly (O(N^3) for the triply nested loop) but there is no direct link between a VM and a disk and finding the link between disk and attachedDisk is O(N^2) because of the way daisy stores this data. To make this prettier and more efficient we could either break the daisy API by switching CreateDisks to a map (will cause problems for anyone using the latest version of the CLI tool to import a pre-written json workflow) or we could modify CIT or daisy to better parse and handle failures from the compute API.

Either way these are long(er) term solutions and we should (imo) do this in the interim to make things work for these machine types right now.


a-crate:
Switched this to only modify disks that were using the default disk type.

drewhli (approved):

PR #1033, opened 2024-01-23
Add C4 performance and shapevalidation

a-crate:
Also added an optional filter flag to pare down perf and shapevalidation cases to only what is needed, should make support matrixing faster

/cc @zmarano @koln67


imagetest/test_suites/networkperf/setup.go
```diff
@@ -15,7 +16,10 @@ import (
 // Name is the name of the test package. It must match the directory name.
 var Name = "networkperf"
 
+var testFilter = flag.String("networkperf_test_filter", ".*", "regexp filter for storageperf test cases, only cases with a matching name will be run")
```
koln67, line 19: small typo: should probably say "filter for networkperf test cases"

a-crate: Woops, yeah.

a-crate:
C4 shapevalidation case names were colliding. Should be ok now.

a-crate:
I'll re-request review from you when I have the presubmits for this working..

a-crate:
Ok should be ready now

PR #1043, opened 2024-01-26
Pipe GCEGuestAgent logs on windows to Format-List to force line wrapping

a-crate:


imagetest/test_suites/guestagent/telemetry_test.go
```diff
@@ -40,17 +40,18 @@ func restartAgent(t *testing.T) {
 
 func getAgentOutput(t *testing.T) string {
 	t.Helper()
-	var cmd *exec.Cmd
 	if utils.IsWindows() {
-		cmd = exec.CommandContext(utils.Context(t), "powershell.exe", "-NonInteractive", "Get-WinEvent", "-Providername", "GCEGuestAgent")
-	} else {
-		cmd = exec.CommandContext(utils.Context(t), "journalctl", "-o", "cat", "-eu", "google-guest-agent")
+		o, err := utils.RunPowershellCmd(`Get-WinEvent -Providername GCEGuestAgent | Format-List -Property Message`)
```
bkatyl, line 44: I actually think `out` was better than `o`. Just because something can be a single character that doesn't always make it better from a readability prospective especially when someone is quickly skimming the code.

go/go-style/decisions#variable-names - "Single-word names like count or options are a good starting point."

a-crate: Done

bkatyl (approved):

PR #1052, opened 2024-02-05
Refactoring disk tests to work on COS

sejalsharma-google:
In order to run CIT against COS, we are proposing the following refactoring for the disk tests.

### DIFF
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -87,8 +87,7 @@ func main() {
 		log.Fatalf("failed to get metadata _test_package_name: %v", err)
 	}
 
-	// NOTE(sejalsharma): modified the following line "", "image_test".
-	workDir, err := ioutil.TempDir("/etc", "iimage_test")
+	workDir, err := ioutil.TempDir("/etc", "image_test")
 	if err != nil {
 		log.Fatalf("failed to create work dir: %v", err)
 	}
--- a/imagetest/cmd/wrapper/main.go
+++ b/imagetest/cmd/wrapper/main.go
@@ -87,7 +87,8 @@ func main() {
 		log.Fatalf("failed to get metadata _test_package_name: %v", err)
 	}
 
-	workDir, err := ioutil.TempDir("", "image_test")
+	// NOTE(sejalsharma): modified the following line "", "image_test".
+	workDir, err := ioutil.TempDir("/etc", "iimage_test")
 	if err != nil {
 		log.Fatalf("failed to create work dir: %v", err)
 	}
--- a/imagetest/test_suites/disk/disk_resize_test.go
+++ b/imagetest/test_suites/disk/disk_resize_test.go
@@ -16,7 +16,6 @@ import (
 )
 
 const (
-	// NOTE(sejalsharma): modified the following line "/boot-marker".
 	markerFile      = "/var/boot-marker"
 	gb              = 1024.0 * 1024.0 * 1024.0
 	defaultDiskSize = 20
@@ -53,7 +52,6 @@ func TestDiskResize(t *testing.T) {
 }
 
 func getDiskSize() (int64, error) {
-	// NOTE(sejalsharma): / to /var.
 	fstatOut, err := exec.Command("df", "-B1", "--output=size", "/var").CombinedOutput()
 	if err != nil {
 		return 0, fmt.Errorf("df command failed with error %v", err)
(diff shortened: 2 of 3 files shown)


imagetest/test_suites/disk/disk_resize_test.go
```diff
@@ -52,7 +53,8 @@ func TestDiskResize(t *testing.T) {
 }
 
 func getDiskSize() (int64, error) {
-	fstatOut, err := exec.Command("df", "-B1", "--output=size", "/").CombinedOutput()
+	// NOTE(sejalsharma): / to /var.
+	fstatOut, err := exec.Command("df", "-B1", "--output=size", "/var").CombinedOutput()
```
a-crate, line 57: I think this one should still be `/`, since we're measuring disk size instead of writing anything here

imagetest/cmd/wrapper/main.go
```diff
@@ -87,7 +87,8 @@ func main() {
 		log.Fatalf("failed to get metadata _test_package_name: %v", err)
 	}
 
-	workDir, err := ioutil.TempDir("", "image_test")
+	// NOTE(sejalsharma): modified the following line "", "image_test".
+	workDir, err := ioutil.TempDir("/etc", "iimage_test")
```
a-crate, line 91: Can you switch this to `image_test` and drop the extra i?

a-crate (commented):
The `// NOTE..` comments aren't needed we can just see the git history which will reference this PR if context is needed.

a-crate:
Once you get the wrapper and the test binary running with this PR you will also run into problems when we try to install things during tests. Off the top of my head we do it [here](https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/imagetest/test_suites/storageperf/startupscripts/install_fio.sh), [here](https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/imagetest/test_suites/metadata/metadata_utils.go#L12), and [here](https://github.com/GoogleCloudPlatform/guest-test-infra/blob/master/imagetest/test_suites/networkperf/startupscripts/linux_common.sh). Not sure what a solution for those will look like.

imagetest/test_suites/disk/disk_resize_test.go
sejalsharma-google: The reason the disk size is measuring from "/var" is because we changed the boot marker file installation: "markerFile = "/var/boot-marker". This is due to the read-only root file system of COS, requiring file installations to be in /var. So the test would only pass if we measured in /var too.

sejalsharma-google:
Thanks for the review! As for the other tests that require installations, I will try to find workarounds when I get to them, thank you for linking them. And as for the disk resize measuring (/var instead of /)- is it possible to make these changes only for COS images (as to not break for other operating systems)? Let me know what you think would be a good solution to this

imagetest/test_suites/disk/disk_resize_test.go
a-crate: The boot marker file doesn't affect reading the root disk size, it should still read root. This (`/var`) would still work but it would be needlessly confusing as `df` will look up the filesystem `/var` is on (`/` unless something weird happens) and just read the file size of `/` anyway.

a-crate:
If we need an exception for specific images in test code, we usually fetch the image name from the MDS and do something like `if strings.HasPrefix(image, "cos-") { ...` but in the case of reading `/` vs `/var` COS should be reading `/` too.

imagetest/test_suites/disk/disk_resize_test.go
sejalsharma-google: When I switch the df command back to "/", the test fails. I tried running the df command on a COS VM and it is measuring /var as having a different size than /, I've attached a snippet below. I believe this might be due to [the way COS manages partitions](https://cloud.google.com/container-optimized-os/docs/concepts/disks-and-filesystem).
Screenshot 2024-02-06 at 10 22 22 AM

a-crate: Oh, I understand now, thanks. Measuring `/var` only for COS and root for everyone else would be ideal then.

sejalsharma-google: Thanks! I can work on adding this as COS specific only (strings.HasPrefix(image, "cos-")) and update the code.

sejalsharma-google: Updated the code to be COS specific

a-crate (approved):

PR #1057, opened 2024-02-07
Exclude 4th gen machine type testing from release pipelines

a-crate:
I also dropped the serial_groups tags (this is a no-op, these don't work cross-pipeline anyway)

/cc @zmarano @elicriffield


elicriffield (approved):
LGTM

PR #1058, opened 2024-02-07
Add better support for COS, SLES, and Ubuntu in the network suite.

a-crate:
Bug filed internally for this.

### DIFF
--- a/imagetest/test_suites/network/alias_ip_test.go
+++ b/imagetest/test_suites/network/alias_ip_test.go
@@ -20,24 +20,12 @@ const (
 )
 
 func TestAliases(t *testing.T) {
-	ctx := utils.Context(t)
-	if image, err := utils.GetMetadata(ctx, "instance", "image"); err != nil {
-		t.Fatalf("could not determine image: %v", err)
-	} else if strings.Contains(image, "sles-15") || strings.Contains(image, "opensuse-leap") || strings.Contains(image, "ubuntu-1604") {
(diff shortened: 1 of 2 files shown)


a-crate:
As long as I'm looking at this I also added exceptions for ubuntu-1604 (there is an issue filed for this as well)

And switch `/boot-marker` to `/var/boot-marker` so it will work on cos.

imagetest/test_suites/network/alias_ip_test.go
```diff
@@ -16,16 +16,28 @@ import (
...
 const (
-	markerFile = "/boot-marker"
+	markerFile = "/var/boot-marker"
 )
 
 func TestAliases(t *testing.T) {
+	ctx := utils.Context(t)
+	if image, err := utils.GetMetadata(ctx, "instance", "image"); err != nil {
+		t.Fatalf("could not determine image: %v", err)
+	} else if strings.Contains(image, "sles-15") || strings.Contains(image, "opensuse-leap") || strings.Contains(image, "ubuntu-1604") {
```
drewhli, line 26: I wonder if it's better to just avoid running these tests in the first place (leaving them out of the list of tests in the `setup` step) instead of skipping in the actual test.

a-crate: Yeah that's probably better. Done.

drewhli (approved):

PR #1062, opened 2024-02-12
Fix wrapper workdir creation on windows

a-crate:


imagetest/cmd/wrapper/main.go
```diff
@@ -87,7 +88,11 @@ func main() {
 		log.Fatalf("failed to get metadata _test_package_name: %v", err)
 	}
 
-	workDir, err := ioutil.TempDir("/etc", "image_test")
+	workDirPath := "/etc/"
+	if runtime.GOOS == "windows" {
+		workDirPath = "C:\\"
+	}
+	workDir, err := ioutil.TempDir(workDirPath, "image_test")
```
drewhli, line 95: `ioutil` is deprecated, so I'd change this to use `os.TempDir()` with `path.Join()` instead.

drewhli (changes requested):

imagetest/cmd/wrapper/main.go
a-crate: I used `os.MkdirTemp` instead since it's a 1:1 replacement suggested by the deprecation noticed, `os.TempDir` behaves slightly differently

drewhli (approved):

PR #1079, opened 2024-02-21
Add a few more package version upload tasks

koln67:
Add some more instances of the uploadpackageversiontask.

In some cases such as windows x86_64, I'm unsure if the a previous "uploadpackagetask" would map to multiple uploadversiontasks, as x86_64 could refer to multiple different ostypes for windows. 

Similarly for guest-diskexpand on debian, the previous task only had one OS of debian, however this could possibly map to multiple ostypes for debian 10 (xenial), debian 11 (bookworm), and debian 12 (bullseye).


koln67:
/hold
wait until we have os types to match the former "generic windows" and "generic debian" targets

koln67:
/unhold
For now, only add tasks with os types which are available until the generic debian and generic windows os types are deployed

bkatyl (approved):

PR #1098, opened 2024-03-11
oslogin: retry for a minute when creating ssh client

dorileo:
Due to the async nature of the test implementation the client side should perform a few attempts to create the ssh client before failing, in the mean time the server could be still in initialization process and the client is failing before we have any change of success - creating test flakiness.

### DIFF
--- a/imagetest/test_suites/oslogin/ssh_test.go
+++ b/imagetest/test_suites/oslogin/ssh_test.go
@@ -180,6 +180,28 @@ func TestAgent(t *testing.T) {
 	}
 }
 
+func sshClient(t *testing.T, hostname string, config *ssh.ClientConfig) *ssh.Client {
+	t.Helper()
+	var client *ssh.Client
+	var err error
+
+	// Retry dialing target system for a minute.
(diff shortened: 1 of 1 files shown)


imagetest/test_suites/oslogin/ssh_test.go
```diff
@@ -180,6 +180,28 @@ func TestAgent(t *testing.T) {
...
+	var client *ssh.Client
+	var err error
+
+	// Retry dialing target system for a minute.
+	for i := 0; i < 5; i++ {
+		client, err = ssh.Dial("tcp", hostname, sshConfig)
+		if err != nil {
+			time.Sleep(10 * time.Second)
+		} else {
+			defer client.Close()
```
a-crate, line 194: Shouldn't this be happening outside of `sshClient`? Otherwise the client is closed with the return on line 202 and unusable on 493

dorileo: yep, just changed it.

a-crate (approved):

a-crate (approved):

PR #1108, opened 2024-03-15
metadata: fix the dnf and yum repo restriction

dorileo:
The repository names for EL distros are pretty stable for guest environment packages, this change fixes it to the known and stable name.

### DIFF
--- a/imagetest/test_suites/metadata/metadata_utils.go
+++ b/imagetest/test_suites/metadata/metadata_utils.go
@@ -4,45 +4,20 @@ import (
 	"fmt"
 	"io"
 	"os/exec"
-	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/guest-test-infra/imagetest/utils"
 )
 
-// yumDnfRunPackageRepoQuery is a common utility between yum and dnf to run the query command
-// provided in the cmd argument and parses its output. The output format is the same both for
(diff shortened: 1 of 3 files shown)


imagetest/test_suites/metadata/metadata_utils.go
```diff
@@ -4,44 +4,11 @@ import (
...
-func yumGetPackageRepo(pkg string) (string, error) {
-	return yumDnfRunPackageRepoQuery(exec.Command("yum", "info", pkg))
-}
-
-// dnfGetPackageRepo queries the dnf package database for the repository pkg was installed from.
-func dnfGetPackageRepo(pkg string) (string, error) {
-	return yumDnfRunPackageRepoQuery(exec.Command("dnf", "info", "-C", "--installed", pkg))
-}
-
 func reinstallPackage(pkg string) error {
```
a-crate, line 12: This isn't really a generic function anymore (which is fine) but it doesn't need the argument since it's only really capable of reinstall a few packges and we only use it once. Can we drop `pkg string`, rename it `reinstallAgent` and hardcode `google-compute-engine-windows`/`google-guest-agent` package names?

dorileo: done. thank you.

a-crate (approved):

PR #1122, opened 2024-04-02
Update package_test.go to include COS package validation

sejalsharma-google:


imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -126,7 +141,7 @@ func TestGuestPackages(t *testing.T) {
 		},
 		{
 			name:       "google-compute-engine",
-			imagesSkip: []string{"sles", "suse"},
+			imagesSkip: []string{"sles", "suse", "cos"},
 		},
 		{
 			name:   "google-guest-configs",
```
a-crate, line 147: This package is the same as `google-compute-engine` just with a different name for sles and suse, it looks like you ship it the same way. Should probably be included here.

imagetest/test_suites/packagevalidation/package_test.go
```diff
@@ -136,17 +151,21 @@ func TestGuestPackages(t *testing.T) {
 			name:   "google-guest-oslogin",
 			images: []string{"sles", "suse"},
 		},
+		{
+                        name: "oslogin",
+                        images: []string{"cos"},
+                },
 		{
 			name:       "gce-disk-expand",
```
a-crate, line 159: I think you ship cloud-disk-resize instead if you want to check for it

a-crate (commented):
This is going to fail the `gofmt` test but LGTM after that
/ok-to-test

a-crate (approved):

a-crate (approved):

PR #1132, opened 2024-04-12
Disable dhcp test on debian 12

dorileo:


a-crate (approved):

a-crate:
debian uses systemd-networkd, in which case this test is supposed to be checking only that network interfaces were configured with DHCP. Instead it's failing because there's no dhclient process.

PR #1154, opened 2024-05-02
Add COS configs to the concourse pipeline

sejalsharma-google:
This PR pairs with this PR in compute-image-tools: https://github.com/GoogleCloudPlatform/compute-image-tools/pull/2309.

This PR does the following:
- Creates a new buildpackageimagetaskcos. This is similar to buildpackageimagetask but has unique parameter sets and invokes install_package_cos.wf.json.
- Adds a call to the buildpackageimagetaskcos for COS 113. In the future we will expand this to more milestones, but we are starting with one for simplicity. This passes in the parameters for the task (like source name, dest name, package version, and cos branch).
- Adds cos as a distro to test against guest-agent.

Notes:
- We might need to grant permissions to service accounts to make this work so permission-related issues are expected.
- Need to reconsider passing in the full path for source_image in the buildpackageimagetaskcos call.
- CCing @zmarano, @dorileo to the PR.

### DIFF
--- a/concourse/pipelines/guest-package-build.jsonnet
+++ b/concourse/pipelines/guest-package-build.jsonnet
@@ -370,6 +370,43 @@ local buildpackageimagetask = {
   },
 };
 
+local buildpackageimagetaskcos = {
+  local tl = self,
+
+  image_name:: error 'must set image_name in buildpackageimagetaskcos',
+  source_image:: error 'must set source_image in buildpackageimagetaskcos',
+  dest_image:: error 'must set dest_image in buildpackageimagetaskcos',
+  package_version:: error 'must set dest_image in buildpackageimagetaskcos',
+  cos_branch:: error 'must set dest_image in buildpackageimagetaskcos',
+  machine_type:: 'e2-medium',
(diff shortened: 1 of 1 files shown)


concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -474,6 +511,13 @@ local build_guest_agent = buildpackagejob {
             machine_type: 't2a-standard-2',
             worker_image: 'projects/compute-image-tools/global/images/family/debian-11-worker-arm64',
           },
+          buildpackageimagetaskcos {
+            image_name: 'cos-113',
```
a-crate, line 515: Should be we following COS-stable instead, or are we expecting to add more families with new releases of COS versions?

sejalsharma-google: I think for testing purposes it might make sense to start off with cos-113-lts, WDYT? If we can get this to work in the pipeline then ideally we would like to test on all active COS versions (ie 101, 105, 109, 117).

a-crate: sgtm

PR #1155, opened 2024-05-03
Fix CIT on SQL Server

a-crate:
/cc @elicriffield @zmarano


a-crate:
Adding on a fix for loadbalancer in sql server too

PR #1160, opened 2024-05-13
Pass commit-sha into COS preloading scripts (not package-version)

sejalsharma-google:
At this point in the concourse workflow, a tag is not yet generated. For that reason, we should be passing the commit sha into the preloading scripts.

CCing @dorileo, @a-crate, @zmarano.


concourse/pipelines/guest-package-build.jsonnet
```diff
@@ -515,7 +515,7 @@ local build_guest_agent = buildpackagejob {
             image_name: 'cos-113',
             source_image: 'projects/cos-cloud/global/images/family/cos-113-lts',
             dest_image: 'cos-113-((.:build-id))',
-            package_version: '((.:package-version))',
+            commit_sha: '((.:commit-sha))',
```
a-crate, line 518: I think you still need to add a `commit_sha` argument to the `buildpackageimagetaskcos` function and use it in the daisy call too

a-crate (approved):

PR #1165, opened 2024-05-14
Make it possible to release windows images with CIT failures

a-crate:
/cc


bkatyl (approved):

concourse/pipelines/windows-image-build-mbr.jsonnet
```diff
@@ -218,7 +218,7 @@ local imgpublishjob = {
       },
     },
   ] +
-  if job.runtests then
+  if job.env == 'testing' && job.runtests then
```
a-crate, line 221: For posterity, and potential future readers of this PR: this is set to job.env == testing because publish to testing is the final step here, it would be env == prod if there was a publish to prod step

PR #1166, opened 2024-05-14
Compute sha 256 hash

koln67:
Compute the sha 256 hash in the concourse image build pipeline and pass it in as part of the JSON.

### DIFF
--- a/concourse/pipelines/linux-image-build.jsonnet
+++ b/concourse/pipelines/linux-image-build.jsonnet
@@ -262,6 +262,15 @@ local imgpublishjob = {
             load_var: 'publish-version',
             file: 'publish-version/version',
           },
+	  {
+	    task: 'generate-hash',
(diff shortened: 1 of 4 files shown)


koln67:
/hold

do not merge until there is enough time in the day to manually test the concourse task

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -262,6 +262,15 @@ local imgpublishjob = {
             load_var: 'publish-version',
             file: 'publish-version/version',
           },
+	  {
+	    task: 'generate-hash',
+	    file: 'guest-test-infra/concourse/tasks/generate-hash.yaml',
```
ChaitanyaKulkarni28, line 267: we need to compute hash for windows images as well

koln67: done

PR #1167, opened 2024-05-16
Update CIT to run only during testing publish for Windows

jjerger:
Probably to be moved to prod when we can fully automate the publish process. For now, we don't want to run the tests on publish-to-prod as thats the final human step to send it to ARLE, and we should be blocking on publish-to-testing if CIT fails.


a-crate:
This will start running CIT on publish-to-testing for windows client images, which is not working right now

a-crate (approved):

PR #1168, opened 2024-05-16
Update prepublish task prod --> testing

jjerger:


a-crate:
Shouldn't we just drop this? It will run CIT twice like this

jjerger:
You're right, it's already run in the other set. Removed it.

a-crate (approved):

PR #1169, opened 2024-05-20
Win11 24h2 preview and sql server 2022 on win2025 preview

bkatyl:
Win11 24h2 preview and sql server 2022 on win2025 preview


jjerger (commented):
/lgtm

PR #1182, opened 2024-05-28
use package instead of repo_name for tagging.

dorileo:


dorileo:
yep, they are the same elsewhere.

PR #1192, opened 2024-06-06
Use start time in Ms as the filename

koln67:
img.tar.gz is being overriden when computing the hash


a-crate:
Is this going to produce confusing error messages? Does this name show up anywhere? Maybe it should be `((.:source-version))-((.:start-timestamp-ms)).tar.gz` if so
/lgtm
/hold for author

a-crate (approved):

PR #1200, opened 2024-06-07
Only print the hash and not the file name

koln67:
sha256sum prints the hash, then a space, then the file name. We only want the hash so drop the file name.


jjerger (commented):
/lgtm

PR #1211, opened 2024-06-28
Add step to load shasum in windows publish

koln67:
Load the shasum variable

### DIFF
--- a/concourse/pipelines/windows-image-build.jsonnet
+++ b/concourse/pipelines/windows-image-build.jsonnet
@@ -843,6 +843,12 @@ local ImgGroup(name, images, environments) = {
              [
                common.GcsImgResource(image, 'windows-install-media')
                for image in windows_install_media_images
+             ] +
+             // windows install media images use imgpublishjob. To ensure the pipeline compiles, this 
+             // shasumresource is needed, even though the media images do not have a sbom or shasum.
+             [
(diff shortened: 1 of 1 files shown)


koln67:
/hold

only merge if the manual concourse testing shows that this steps is necessary for the windows publish to succeed

concourse/pipelines/windows-image-build.jsonnet
```diff
@@ -819,6 +843,10 @@ local ImgGroup(name, images, environments) = {
              [
                common.GcsImgResource(image, 'windows-install-media')
                for image in windows_install_media_images
+             ] +
+             [
+               common.GcsShasumResource(image, 'windows-install-media')
```
jjerger, line 848: Did you intend to make this the shasum?

koln67: I added it so it passes the presubmit, however it doesn't have any function.

The windows media images don't use the sha-sum or sbom, however it uses the "imgpublishjob" which is common for all the windows builds.

koln67: Fixed: the current way skips the shasum steps for windows media images by passing in "generate_shasum = false"

jjerger (commented):
/lgtm

PR #1237, opened 2024-09-10
Include test suites instead of excluding

koln67:
Use "filter" instead of "exclude": this change uses "filter" for all the test suites which were not previously covered by "exclude."


koln67:
/unhold

Don't run winrm or windowscontainers on linux, and intentionally skip vmspec on windows after the windows-2016 test failed for vmspec.

a-crate (approved):

PR #1248, opened 2024-10-11
update go containers to also install protobuffer tools.

dorileo:


a-crate:
These containers need some refactoring. We should create a `guest-go-toolchain` image to have a single definition of the toolchain, and derive `go{build,test.check}` from the base image. That would also make it easier for us to update everything from `bullseye`->`bookworm` which is another thing that needs to happen.

/lgtm

PR #1254, opened 2024-10-17
package build: pass in extra git ref

dorileo:
Pass in the extra git ref used to build legacy guest agent and plugin manager side-by-side.

### DIFF
--- a/packagebuild/workflows/build_package.wf.json
+++ b/packagebuild/workflows/build_package.wf.json
@@ -26,6 +26,9 @@
     "git_ref": {
       "Required": true
     },
+    "extra_git_ref": {
+      "Required": false
+    },
     "version": {
       "Required": true
     },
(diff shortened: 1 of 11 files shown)


packagebuild/workflows/build_package.wf.json
```diff
@@ -60,6 +63,7 @@
             "repo-owner": "${repo_owner}",
             "repo-name": "${repo_name}",
             "git-ref": "${git_ref}",
+            "extra-git-ref": "${extra_git_ref}",
```
drewhli, line 66: We should modify the `daisy_startupscript-XXX.sh` scripts to also use these `refs`.

dorileo: Yes, it will come on its own PR, I believe that we can split.

drewhli (approved):

PR #1260, opened 2024-10-18
Build and test derivative images for guest-configs

a-crate:
/cc


a-crate:
Found some images that were missing from build_guest_agent, added those as well

PR #1262, opened 2024-10-18
Modify debian startupscripts to utilize extra_repo if it exists

drewhli:
/hold

### DIFF
--- a/packagebuild/daisy_startupscript_deb.sh
+++ b/packagebuild/daisy_startupscript_deb.sh
@@ -19,6 +19,8 @@ SRC_PATH=$(curl -f -H Metadata-Flavor:Google ${URL}/daisy-sources-path)
 REPO_OWNER=$(curl -f -H Metadata-Flavor:Google ${URL}/repo-owner)
 REPO_NAME=$(curl -f -H Metadata-Flavor:Google ${URL}/repo-name)
 GIT_REF=$(curl -f -H Metadata-Flavor:Google ${URL}/git-ref)
+EXTRA_REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/extra-repo)
+EXTRA_GIT_REF=$(curl -f -H Metadata-Flavor:Google ${URL}/extra-git-ref)
 BUILD_DIR=$(curl -f -H Metadata-Flavor:Google ${URL}/build-dir)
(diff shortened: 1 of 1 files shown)


packagebuild/daisy_startupscript_deb.sh
```diff
@@ -42,6 +44,14 @@ try_command apt-get install -y --no-install-{suggests,recommends} git-core \
 
 git_checkout "$REPO_OWNER" "$REPO_NAME" "$GIT_REF"
 
+if [[ ! -z "$EXTRA_REPO" ]]; then
```
ChaitanyaKulkarni28, line 47: Do we need to install protoc?

drewhli: Adding as dependency for `guest-agent` control file.

drewhli: On second though, installing it as part of this script. `debian-11` doesn't have `protoc-gen-go-grpc` in its `apt` repositories, so this is required for that anyways.

ChaitanyaKulkarni28: actually, we just introduced common [install_protoc](https://github.com/GoogleCloudPlatform/guest-test-infra/blob/eb8ab86e71a42dd023c296cc10f725eafd477f64/packagebuild/common.sh#L32) for that. You can use it here so it will be consistent across packaging

packagebuild/daisy_startupscript_deb.sh
```diff
@@ -38,10 +40,19 @@ sed -i 's/^.*debian buster-backports main.*$//g' /etc/apt/sources.list
 
 try_command apt-get -y update
 try_command apt-get install -y --no-install-{suggests,recommends} git-core \
-  debhelper devscripts build-essential equivs libdistro-info-perl
+  debhelper devscripts build-essential equivs libdistro-info-perl protobuf-compiler
+install_protoc
```
ChaitanyaKulkarni28, line 44: Is `unzip`/`make` already installed? You might want to add that if not already installed, I had to install it for googet packaging

drewhli: `unzip` isn't, but `make` is as it's a part of `build-essential`. Added `unzip` to the list of installs.

PR #1263, opened 2024-10-18
rpm build: build package with extra source

dorileo:
Introduce the required bits to honor extra source code when building rpm packages - this supports packaging that require more than a single source code such as guest-agent that is building plugin-manager side-by-side.

### DIFF
--- a/packagebuild/daisy_startupscript_rpm.sh
+++ b/packagebuild/daisy_startupscript_rpm.sh
@@ -27,6 +27,8 @@ SRC_PATH=$(get_md daisy-sources-path)
 REPO_OWNER=$(get_md repo-owner)
 REPO_NAME=$(get_md repo-name)
 GIT_REF=$(get_md git-ref)
+EXTRA_REPO=$(get_md extra-repo)
+EXTRA_GIT_REF=$(get_md extra-git-ref)
 BUILD_DIR=$(get_md build-dir)
 VERSION=$(get_md version)
 VERSION=${VERSION:-"dummy"}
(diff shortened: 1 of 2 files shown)


packagebuild/daisy_startupscript_rpm.sh
```diff
@@ -94,6 +96,20 @@ for dir in ${RPMDIR}/{SOURCES,SPECS}; do
   [[ -d "$dir" ]] || mkdir -p "$dir"
 done
 
+EXTRA_RPMBUILD_FLAGS=""
+if [[ -n "EXTRA_REPO" ]]; then
+  EXTRA_RPMBULD_FLAGS="--define 'has_extra_source 1'"
+
+  pushd $BUILD_DIR
```
ChaitanyaKulkarni28, line 103: I don't think we set BUILD_DIR

dorileo: indeed. even if we were it would be wrong, I sent a new version doing it right. Thank you.

packagebuild/daisy_startupscript_rpm.sh
```diff
@@ -75,6 +77,7 @@ fi
 try_command yum install -y $GIT rpmdevtools yum-utils python3-devel
 
 git_checkout "$REPO_OWNER" "$REPO_NAME" "$GIT_REF"
+ROOT_WORK_DIR=$(pwd)
```
ChaitanyaKulkarni28, line 80: missed earlier, don't we need to install protoc?

dorileo: yep, I missed that too. I'll get it done.

dorileo: done. thank you.

PR #1284, opened 2024-10-31
COS 101 is EOL. Move other COS package tests to Debian 12 worker.

zmarano:


concourse/pipelines/guest-package-build.jsonnet
a-crate, line 1: Can we add 117 as a replacement?

zmarano: Oh- yes- I didn't even see that was there. Done.

a-crate (approved):

PR #1285, opened 2024-10-31
Add arm64 variants for CentOS Stream 9, RHEL 8, and AlmaLinux 9

zmarano:
- Note RHEL 8 won't yet work.


a-crate (approved):
Sorry, missed this one somehow.

PR #1287, opened 2024-11-06
Test passing in cit suite filter as a list with override

jjerger:
Testing a different approach to allow per-OS filtering, but keep a default if it is not defined.

### DIFF
--- a/concourse/pipelines/linux-image-build.jsonnet
+++ b/concourse/pipelines/linux-image-build.jsonnet
@@ -239,6 +239,8 @@ local imgpublishjob = {
   trigger:: if tl.env == 'testing' then true
   else false,
 
+  citfilter:: '^(%s)$' % self.citsuites,
+  citsuites:: 'cvm|livemigrate|suspendresume|loadbalancer|guestagent|hostnamevalidation|imageboot|licensevalidation|network|security|hotattach|lssd|disk|packagevalidation|ssh|metadata|vmspec',
   runtests:: if tl.env == 'testing' then true
   else false,
 
@@ -357,6 +359,7 @@ local imgpublishjob = {
             {
               task: 'image-test-' + tl.image,
               config: imagetesttask {
+                filter: tl.citfilter
                 images: 'projects/bct-prod-images/global/images/%s-((.:publish-version))' % tl.image_prefix,
               },
               attempts: 3,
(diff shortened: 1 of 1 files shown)


concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -239,6 +239,8 @@ local imgpublishjob = {
   trigger:: if tl.env == 'testing' then true
   else false,
 
+  citfilter:: '^(%s)$' % self.citsuites,
```
a-crate, line 242: Why define `citsuites` just to interpolate it into `citfilter`? Just allow the caller to specify the full regexp.

concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -239,6 +239,8 @@ local imgpublishjob = {
...
+  citsuites:: 'cvm|livemigrate|suspendresume|loadbalancer|guestagent|hostnamevalidation|imageboot|licensevalidation|network|security|hotattach|lssd|disk|packagevalidation|ssh|metadata|vmspec',
```
a-crate, line 243: The current default is kinda defined in a wrapper around common.imagetesttask on line 18. There should be a single definition of the defaults to avoid confusion - either drop the wrapper or define the defaults in another variable and reference it in both places.

concourse/pipelines/linux-image-build.jsonnet
a-crate: Oh, I see, to allow people to add suites. If that's the approach, `citsuites` should probably be a list that gets concatenated with `|` characters when it's put into the surrounding `^()$` so that the usage is clearer.

jjerger: I think expecting the user to pass in the full regexp might be better here so we're not messing with transforming input.

concourse/pipelines/linux-image-build.jsonnet
jjerger: I dropped it in the line 18 in favor of next to the publish job definitions since that's where it's being used. If that placement doesn't make sense logically we can stick with having it defined globally vs just for the imgpublishjob where it's being used.

a-crate (approved):

PR #1301, opened 2024-11-12
update builds to occur starting between 10:30 and 11:00 AM

lpleahy:


concourse/pipelines/linux-image-build.jsonnet
```diff
@@ -465,7 +465,7 @@ local imggroup = {
                {
                  name: 'daily-time',
                  type: 'time',
-                 source: { interval: '24h' },
+                 source: { interval: '24h', start: '10:30 AM', stop: '11:00AM', location: 'America/Los_Angeles' },
```
bkatyl, line 468: I'd suggest 8am for the linux images so they're likely done when the windows images start.

Also I'd suggest adding: initial_version: true 
or extended the time window to midnight.

Based on https://github.com/concourse/time-resource/?tab=readme-ov-file#source-configuration

initial_version: Optional. When using start and stop as a trigger for a job, you will be unable to run the job manually until it goes into the configured time range for the first time (manual runs will work once the time resource has produced it's first version).

An alternative idea would be to stop at midnight, so we could re-run it within the time window.

bkatyl (approved):

PR #1326, opened 2025-01-07
Update go-containerregistry version

jjerger:
Newer version to support AR authentication; old version was not authenticating.

### DIFF
--- a/container_images/registry-image-forked/types.go
+++ b/container_images/registry-image-forked/types.go
@@ -161,7 +161,7 @@ func (source Source) AuthOptions(repo name.Repository, scopeActions []string) ([
 	} else if source.Google {
 		logrus.Warnf("Forked registry image: will use Google default credentials")
 		var err error
-		if auth, err = google.NewEnvAuthenticator(); err != nil {
+		if auth, err = google.NewEnvAuthenticator(context.Background()); err != nil {
 			logrus.Errorf("failed to determine Google default credentials: %v.", err)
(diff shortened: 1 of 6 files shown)


container_images/registry-image-forked/types.go
```diff
@@ -161,7 +162,7 @@ func (source Source) AuthOptions(repo name.Repository, scopeActions []string) ([
 	} else if source.Google {
 		logrus.Warnf("Forked registry image: will use Google default credentials")
 		var err error
-		if auth, err = google.NewEnvAuthenticator(); err != nil {
+		if auth, err = google.NewEnvAuthenticator(context.Background()); err != nil {
```
ChaitanyaKulkarni28, line 165: nit: we should use same context here and line 205 instead of creating a new one

jjerger: Made them use the same background context

container_images/registry-image-forked/types.go
```diff
@@ -152,6 +153,7 @@ func (source Source) Mirror() (Source, bool, error) {
 
 // AuthOptions is
 func (source Source) AuthOptions(repo name.Repository, scopeActions []string) ([]remote.Option, error) {
+	var ctx context.Background()
```
ChaitanyaKulkarni28, line 156: nit: `ctx := context.Background()` instead of `var ctx context.Background()`

jjerger: copied the line below without thinking about it. Fixed

PR #1327, opened 2025-01-07
Update go-containerregistry sum

jjerger:
Reflects new go-containerregistry version


ChaitanyaKulkarni28:
Ah, this was the go mod tidy thing it was earlier complaining about. 

/lgtm

PR #1343, opened 2025-02-20
Use the a3 ultra machine shape by default

koln67:
Previously the machine shape was hard coded into the test suite. We want to make it generic, so the first step is to pass in the machine shape argument.


ajorg (approved):
/lgtm

PR #1370, opened 2025-04-21
Add Debian13 to dev pipeline

jjerger:
We want to start automated build for testing


jjerger:
This can be committed now, although the build may not work correctly until some of the package repos are updated, which is a WIP

PR #1439, opened 2025-07-21
Update pipelines to use new guest package builder project.

drewhli:
Most of the diff is from some auto-formatting of the oslogin jobs, it seems.


ChaitanyaKulkarni28:
Don't we need to update images to test list as well? Its hardcoding `gcp-guest` everywhere

concourse/pipelines/guest-package-build.jsonnet
ChaitanyaKulkarni28, line 1: Do we need to change project on line 25 for publishresulttask as well?

PR #1462, opened 2025-08-08
Force Deb13 to use phase2 guest agent package

jjerger:
Pinning it this way temporarily so that we can build with phase 2 guest agent. This will be removed once we can move to phase 3.


dorileo:
I don't think this is the way to go.

PR #1521, opened 2025-09-24
Add trigger field to package build job to customize whether it triggers off of commits

drewhli:
Also disable triggers for guest agent stable and dev pipelines.

/hold


drewhli:
Not exactly sure why the presubmit is showing the triggers still as true. Running `jsonnet` locally, it shows that the triggers are updated to `false` as expected...

ChaitanyaKulkarni28 (approved):

PR #1678, opened 2026-04-28
Update deprecate-images.sh to use gcloud instead of gcrane. 

shenpai35:
Also updated container-build.jsonnet.

### DIFF
--- a/concourse/scripts/deprecate-images.sh
+++ b/concourse/scripts/deprecate-images.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
+
 ##
 # deprecate-images.sh
-# Tags older images in GCR/AR as deprecated, keeping only the latest.
+# Tags older images in GCR/AR as deprecated using gcloud.
 ##
 
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


concourse/scripts/deprecate-images.sh
```diff
@@ -39,7 +40,8 @@ for sha_with_prefix in $OLD_SHAS; do
     count=$((count + 1))
   else
     echo "Tagging $IMAGE_URL@$sha_with_prefix as $TAG"
-    gcrane tag "$IMAGE_URL@$sha_with_prefix" "$TAG"
+    # gcloud requires the full image:tag or image@digest as the source/destination
+    gcloud container images add-tag "${IMAGE_URL}@${sha_with_prefix}" "${IMAGE_URL}:${TAG}" --quiet
```
SaswatPadhi, line 44: [nit] can we remove `--quiet`?

Verbosity in workflow logs might actually be useful.

SaswatPadhi (approved):
LGTM; thanks!

PR #1680, opened 2026-04-28
add --quiet flag back to disable interactive prompt.

shenpai35:


SaswatPadhi (approved):
Thanks, didn't realize `--quiet` suppresses interaction, not verbosity. This is so weird.

For example, https://linux.die.net/man/8/apt-get and many others.

PR #1683, opened 2026-05-01
Pin deprecation script location to guest-test-infra instead of as a variable input.

shenpai35:
Since we build some containers in other repos such as compute-daisy, we don't want to duplicate the script there as well.

### DIFF
--- a/concourse/pipelines/container-build.jsonnet
+++ b/concourse/pipelines/container-build.jsonnet
@@ -104,7 +104,7 @@ local DeprecateOldImagesTask(image_url) = {
         tag: 'latest' 
       },
     },
-    inputs: [{ name: guest-test-infra }],
+    inputs: [{ name: 'guest-test-infra' }],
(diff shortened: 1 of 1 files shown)


concourse/pipelines/container-build.jsonnet
```diff
@@ -104,11 +104,11 @@ local DeprecateOldImagesTask(image_url, input_name) = {
         tag: 'latest' 
       },
     },
-    inputs: [{ name: input_name }],
+    inputs: [{ name: 'guest-test-infra' }],
```
SaswatPadhi: Hmm, do we still need to define this `inputs` key, even if we are not using it any more?

shenpai35: Yes I think we still need this to have jobs built from this repo and compute-daisy repo to access the script.

SaswatPadhi (approved):

PR #1685, opened 2026-05-01
Add extra_resources: ['guest-test-infra'] to container images built

shenpai35:
outside of this repo.


SaswatPadhi (commented):
I defer to Brian for a thorough review, because at this point I am unsure about `input` vs `inputs` vs `extra_resources`.

We are adding `extra_resources` to many jobs here, ~~but they are already green without it.~~

For the specific daisy test-runner failing job, I think we just need to add another input (or an extra resource?):

https://github.com/GoogleCloudPlatform/guest-test-infra/blob/65a45d769e01f5aea9937aae68d57b640e93b3c0/concourse/pipelines/container-build.jsonnet#L287

SaswatPadhi:
Based on offline discussions, it sounds like `extra_resources` makes the "resource" available and `input`/`inputs` mounts the resource once available.

We are only adding `extra_resources` for the jobs that need it (don't already have guest-test-infra resource available).

Thanks for the clarification @shenpai35.

bkatyl (approved):

PR #1695, opened 2026-05-12
Refactor  Concourse pipeline test steps to use across jobs for improved configuration management

ChaitanyaKulkarni28:
Refactor guest agent test loops to native Concourse across step blocks
- reduces the generated static pipeline payload 
- should eliminate endless vertical scrolling on the job build page by collapsing hundreds of step
- should show interactive matrix navigation grids directly inside each task block  to easily identify passing and failing permutations


ChaitanyaKulkarni28:
the across step must be explicitly opted-in to using the `--enable-across-step` flag, closing it for now as across is not enabled

PR #1781, opened 2026-08-31
Remove use_dynamic_template now that we are fully utilizing the new format

TylerJDao:
/cc @bkatyl 

use_dynamic_template is no longer needed since we have fullly switched to using the consolidated workflow files.

/hold


TylerJDao:
Closing this PR, the use_dynamic_template is needed for non-RHEL Linux jobs in the pipeline.

